PR #10, opened 2016-01-28
Run testrunner tests during test phase of core

lossyrob:
This PR makes it so that `mvn test` will run the "testrunner" test cases, which use the `testxml` test cases to run a custom-built test runner to test the core library.

This required the copying out of the test runner code, and necessary pieces, to the core test code. This was a bit of code to copy, with some modifications, because the `testrunner` and the `testbuilder` share quite a bit of code. We might want to think of a better way to do this; a better split implementation could be that we pull the test running into it's own module, which the jts core tests depend on, and contains code that the testbuilder can also rely on. We can do this as part of this PR, or we can merge this in and do that as part of some later work.

Also update the README.


dr-jts:
Would it make sense to pull the mvn commands out into a separate file, something like DEVELOPING.md ?

dr-jts:
My thought is that the TestRunner should go into a separate module (prob. "testrunner").  Either JTS core tests depend on this, or possibly the XML tests move into that module as well.  TestBuilder can remain in app and depend on the testrunner module.

lossyrob:
@dr-jts re: the README, if you'd like we could pull that out...I'm used to seeing development-specific stuff in READM.md's, but if you want them separate then that's totally cool by me and I'll make the change.

re: the testrunner stuff...sounds good. I can try to make this change at some point soon and update the PR.

dr-jts:
Yeah, I think it would be nice to have them separate.  I expect the README to grow with info about modules, and maybe links to docs etc.  Also helps to make commits clear about what they contain.

Be great if you can update the testrunner location too.

PR #22, opened 2016-03-10
Change EPL/BSD license headers to EPL/EDL

lossyrob:
This PR changes the previous copyright headers, which were dual EPL/BSD, to EDL.


jodygarnett:
Looks good to me, resulting in a more consistent codebase

bjornharrtell:
Since EDL is incompatible with GPL, might this change pose a problem for projects like GeoServer?@jodygarnett? From https://eclipse.org/legal/eplfaq.php#GPLCOMPATIBLE:

"Based upon the position of the Free Software Foundation, you may not combine EPL and GPL code in any scenario where linking exists between code made available under those licenses"

mprins:
Note that the license in the pom.xml still states BSD license

dr-jts:
@bjornharrtell That link refers to EPL.  EDL is apparently essentially BSD, and thus should not be a problem?  http://www.gnu.org/licenses/license-list.en.html#ModifiedBSD

bjornharrtell:
Yes sorry I was getting confused. You are correct and it's clearly stated at https://eclipse.org/org/documents/edl-v10.php.

PR #24, opened 2016-03-25
Dependency management cleanup and improvement

fgdrf:
This pull request cleans up checked-in jars that are never used because of maven dependency management:
- junit-3.7.jar
- jdom.jar (1.0 beta8-dev) in root-pom Version 1.1.3 is declared
- xerces.jar (1.2.0 ??)

In addition the version of dependencies are moved to dependencyManagement-section. everywhere a dependency is used, the version is used from there.

Last but not least a minor improvment to use project.version property instead of project.parent.version, version is inherited from parent by default.

Signed-off-by: Frank Gasdorf fgdrf@users.sourceforge.net


lossyrob:
+1, @dr-jts good to merge?

PR #43, opened 2016-11-07
Creates fat-jar for TestBuilder and sets up TestRunner in Maven build. 

jnh5y:


dr-jts:
Are you able to update this to remove the unneeded resources reference?

Also a shorter name for the jar would be nice, if possible?  Or is it standard?

jnh5y:
@dr-jts I opened a new PR for just the test runner + Maven/JUnit work.  Another PR merged the resources fix and the jar-with-dependencies bit.  (And I'm closing this PR.  New PR is https://github.com/locationtech/jts/pull/60.)

In terms of names, we can sort out a shorter one.  I've entered https://github.com/locationtech/jts/issues/61 to track that.

PR #75, opened 2017-02-17
Add K Nearest Neighbor search into STR-Tree

jiayuasu:
Add K Nearest Neighbor search into STR-Tree and corresponding unite test case

Signed-off-by: Jia Yu <jiayu2@asu.edu>

### DIFF
--- a/modules/app/src/main/java/org/locationtech/jtstest/function/SpatialIndexFunctions.java
+++ b/modules/app/src/main/java/org/locationtech/jtstest/function/SpatialIndexFunctions.java
@@ -117,12 +117,19 @@ public class SpatialIndexFunctions
     return (Geometry) result;
   }
 
-  public static Geometry[] strTreeNNk(Geometry geoms, Geometry geom)
+  public static Geometry strTreeNNk(Geometry geoms, Geometry geom)
   {
     STRtree index = buildSTRtree(geoms);
-    int k = 10;
-    Object[] result = index.nearestNeighbour(geom.getEnvelopeInternal(), geom, new GeometryItemDistance(), k);
-    return (Geometry[]) result;
+    int k = 2;
+    Object[] knnObjects = index.nearestNeighbour(geom.getEnvelopeInternal(), geom, new GeometryItemDistance(), k);
+    Geometry[] knnGeoms = new Geometry[k];
+    // Use a loop to cast objects to geometries.
+    for(int i=0;i<k;i++)
+    {
+    	knnGeoms[i] = (Geometry) knnObjects[i];
+    }
+    GeometryCollection geometryCollection = geoms.getFactory().createGeometryCollection(knnGeoms);
+    return geometryCollection;
   }
   
   public static Geometry quadTreeQuery(Geometry geoms, Geometry queryEnv)
--- a/modules/app/src/main/java/org/locationtech/jtstest/function/SpatialIndexFunctions.java
+++ b/modules/app/src/main/java/org/locationtech/jtstest/function/SpatialIndexFunctions.java
@@ -14,6 +14,7 @@ package org.locationtech.jtstest.function;
 
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
 
@@ -117,18 +118,12 @@ public class SpatialIndexFunctions
     return (Geometry) result;
   }
 
-  public static Geometry strTreeNNk(Geometry geoms, Geometry geom)
+  public static Geometry strTreeNNk(Geometry geoms, Geometry geom, int k)
   {
     STRtree index = buildSTRtree(geoms);
-    int k = 2;
     Object[] knnObjects = index.nearestNeighbour(geom.getEnvelopeInternal(), geom, new GeometryItemDistance(), k);
-    Geometry[] knnGeoms = new Geometry[k];
-    // Use a loop to cast objects to geometries.
-    for(int i=0;i<k;i++)
-    {
-    	knnGeoms[i] = (Geometry) knnObjects[i];
-    }
-    GeometryCollection geometryCollection = geoms.getFactory().createGeometryCollection(knnGeoms);
+    List knnGeoms = new ArrayList(Arrays.asList(knnObjects));
+    Geometry geometryCollection = geoms.getFactory().buildGeometry(knnGeoms);
     return geometryCollection;
   }
   
--- a/modules/app/src/main/java/org/locationtech/jtstest/function/SpatialIndexFunctions.java
+++ b/modules/app/src/main/java/org/locationtech/jtstest/function/SpatialIndexFunctions.java
@@ -117,6 +117,14 @@ public class SpatialIndexFunctions
     return (Geometry) result;
   }
 
+  public static Geometry[] strTreeNNk(Geometry geoms, Geometry geom)
+  {
+    STRtree index = buildSTRtree(geoms);
+    int k = 10;
+    Object[] result = index.nearestNeighbour(geom.getEnvelopeInternal(), geom, new GeometryItemDistance(), k);
+    return (Geometry[]) result;
+  }
+  
   public static Geometry quadTreeQuery(Geometry geoms, Geometry queryEnv)
   {
     Quadtree index = buildQuadtree(geoms);
--- a/modules/core/src/main/java/org/locationtech/jts/index/strtree/BoundablePairDistanceComparator.java
+++ b/modules/core/src/main/java/org/locationtech/jts/index/strtree/BoundablePairDistanceComparator.java
@@ -1,5 +1,6 @@
+
 /*
- * Copyright (c) 2016 Jia Yu.
+ * Copyright (c) 2017 Jia Yu.
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
--- a/modules/core/src/main/java/org/locationtech/jts/index/strtree/BoundablePairDistanceComparator.java
+++ b/modules/core/src/main/java/org/locationtech/jts/index/strtree/BoundablePairDistanceComparator.java
@@ -1,3 +1,14 @@
+/*
+ * Copyright (c) 2016 Jia Yu.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v. 1.0 which accompanies this distribution.
+ * The Eclipse Public License is available at http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ */
 package org.locationtech.jts.index.strtree;
 
 import java.io.Serializable;
--- a/modules/core/src/main/java/org/locationtech/jts/index/strtree/BoundablePairDistanceComparator.java
+++ b/modules/core/src/main/java/org/locationtech/jts/index/strtree/BoundablePairDistanceComparator.java
@@ -1,0 +1,53 @@
+package org.locationtech.jts.index.strtree;
+
+import java.io.Serializable;
+import java.util.Comparator;
+
+
+/**
+ * The Class BoundablePairDistanceComparator. It implements Java comparator and is used 
+ * as a parameter to sort the BoundablePair list.
+ */
+public class BoundablePairDistanceComparator implements Comparator<BoundablePair>, Serializable{
+	
+	/** The normal order. */
+	boolean normalOrder;
+
+	/**
+	 * Instantiates a new boundable pair distance comparator.
+	 *
+	 * @param normalOrder The true means puts the least record at the head of this queue.
+	 * This is the natural order. PriorityQueue peek() will get the least element. Vice versa.
+	 */
+	public BoundablePairDistanceComparator(boolean normalOrder)
+	{
+		this.normalOrder = normalOrder;
+	}
+	
+	/* (non-Javadoc)
+	 * @see java.util.Comparator#compare(java.lang.Object, java.lang.Object)
+	 */
+	public int compare(BoundablePair p1, BoundablePair p2) {
+		double distance1 = p1.getDistance();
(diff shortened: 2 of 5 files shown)


modules/core/src/main/java/org/locationtech/jts/index/strtree/BoundablePairDistanceComparator.java
```diff
@@ -0,0 +1,53 @@
+package org.locationtech.jts.index.strtree;
```
jnh5y: New files require a license header.

The form of the header is...

/*
 * Copyright (c) 2016 Vivid Solutions.
 *
 * All rights reserved. This program and the accompanying materials
 * are made available under the terms of the Eclipse Public License v1.0
 * and Eclipse Distribution License v. 1.0 which accompanies this distribution.
 * The Eclipse Public License is available at http://www.eclipse.org/legal/epl-v10.html
 * and the Eclipse Distribution License is available at
 *
 * http://www.eclipse.org/org/documents/edl-v10.php.
 */

You can list yourself as having copyright.

modules/core/src/test/java/org/locationtech/jts/index/GeometryDistanceComparator.java
```diff
@@ -0,0 +1,58 @@
+package org.locationtech.jts.index;
```
jnh5y: Ditto about the license headers for this file.

modules/core/src/main/java/org/locationtech/jts/index/strtree/STRtree.java
```diff
@@ -263,7 +263,30 @@ protected Comparator getComparator() {
...
   }
-  
+  /**
+   * Finds k items in this tree which are the top k nearest neighbors to the given {@link Object}, 
+   * using {@link ItemDistance} as the distance metric.
+   * A Branch-and-Bound tree traversal algorithm is used
+   * to provide an efficient search. This method implements the KNN algorithm described in the following paper:
+   * Roussopoulos, Nick, Stephen Kelley, and Frédéric Vincent. "Nearest neighbor queries." ACM sigmod record. Vol. 24. No. 2. ACM, 1995.
+   * <p>
+   * The query <tt>object</tt> does <b>not</b> have to be
```
dr-jts: Should this say item rather than object?

modules/core/src/main/java/org/locationtech/jts/index/strtree/STRtree.java
```diff
@@ -263,7 +263,30 @@ protected Comparator getComparator() {
...
+   * have to be compatible with the <tt>itemDist</tt> 
+   * distance metric. 
+   * 
+   * @param env the envelope of the query item
+   * @param item the item to find the nearest neighbour of
+   * @param itemDist a distance metric applicable to the items in this tree and the query item
+   * @param k the K nearest items in kNearestNeighbour
+   * @return the K nearest items in this tree
+   */
+  public Object[] kNearestNeighbour(Envelope env, Object item, ItemDistance itemDist,int k)
```
dr-jts: It might be better to call this nearestNeighbour (using overloading), or else nearestNeighbourK (so that the methods are adjacent in the Javadoc)

dr-jts: @jiayuasu What do you think about this suggestion?

modules/core/src/main/java/org/locationtech/jts/index/strtree/STRtree.java
jiayuasu: Well, the query object actually is a Java Object class although it is used as "Item" in ItemBoundable. I am following the same name convention of the original NearestNeighbor() function.

modules/core/src/main/java/org/locationtech/jts/index/strtree/STRtree.java
jiayuasu: @dr-jts I agree with you. I think using overloading is better and easy for understanding. I will make it same with "nearestNeighbour" with an additional parameter "int k".

modules/core/src/main/java/org/locationtech/jts/index/strtree/STRtree.java
dr-jts: Have a look at http://stackoverflow.com/questions/1667212/how-to-add-reference-to-a-method-parameter-in-javadoc

It should probably say {@code item}, or else maybe {@ link Object} (although I think the former is more specific)

jiayuasu: @dr-jts Sure. I understand your point. I will use {@code item} style. I will push them now.

modules/core/src/main/java/org/locationtech/jts/index/strtree/STRtree.java
```diff
@@ -373,5 +402,91 @@ public Object nearestNeighbour(Envelope env, Object item, ItemDistance itemDist)
...
+         * Otherwise, expand one side of the pair,
+         * (the choice of which side to expand is heuristically determined) 
+         * and insert the new expanded pairs into the queue
+         */
+        bndPair.expandToQueue(priQ, distanceLowerBound);
+      }
+    }
+    // done - return items with min distance
 
+    Object[] result = new Object[kNearestNeighbors.size()];
```
dr-jts: Can this use the Collection.toArray() method?

modules/core/src/main/java/org/locationtech/jts/index/strtree/STRtree.java
```diff
@@ -373,5 +402,91 @@ public Object nearestNeighbour(Envelope env, Object item, ItemDistance itemDist)
           ((ItemBoundable) minPair.getBoundable(1)).getItem()
       };
   }
+  private Object[] nearestNeighbour(BoundablePair initBndPair, double maxDistance, int k)
```
dr-jts: Is it possible that the original NN(1) method can just delegate to this one with k = 1 ?   That would be fewer code pathways to maintain.  Unless it is slower?

modules/core/src/main/java/org/locationtech/jts/index/strtree/STRtree.java
jiayuasu: @dr-jts No, it cannot. The K Nearest Neighbor Priority Queue needs to contain top K ItemBoundable (item+its bounding envelope). We have two options (1) Use the existing way (2) Maintain an additional list of top k nearest neighbors' bounding envelopes all the time.  The (1) is OK because normally the K is much smaller than the original dataset. The (2) is time-consuming because we need to update the additional list all the time along with KNN Priority Queue.

modules/core/src/main/java/org/locationtech/jts/index/strtree/STRtree.java
jiayuasu: @dr-jts Yes, you are right. If k = 1, the KNN method has the same result and same performance with the original NN(1) method because the maintained KNN Queue just becomes 1NN. However, their APIs are a little bit different. The return of the original NN(1) method is a different array in the following format:[Nearest item, Query Point]. The new NN(K) method returns an array [Nearest item1, item2, item3,...., item k] without including Query Point.

modules/core/src/main/java/org/locationtech/jts/index/strtree/STRtree.java
dr-jts: Oh, right, I missed that.  In that case might be nice to pull this code out into a private static method.

modules/core/src/main/java/org/locationtech/jts/index/strtree/STRtree.java
dr-jts: Right, good point!  

You know, in that case these method probably really should have different names (since the return data is different).  So how about calling the new method nearestNeighbourK ?

jiayuasu: @dr-jts Oh, sorry. I missed this: the original NN(1) has an additional API wrapper to return the item only.

```
  public Object nearestNeighbour(Envelope env, Object item, ItemDistance itemDist)
  {
    Boundable bnd = new ItemBoundable(env, item);
    BoundablePair bp = new BoundablePair(this.getRoot(), bnd, itemDist);
    return nearestNeighbour(bp)[0];
  }
```

So, the NN(1) and the NN(K) have the same return format from the user's perspective. Do you think I still need to use "nearestNeighbourK"?

modules/core/src/main/java/org/locationtech/jts/index/strtree/STRtree.java
jiayuasu: Sure. I will make it into a separate private static method.

modules/core/src/main/java/org/locationtech/jts/index/strtree/STRtree.java
dr-jts: I guess not. Although @bjornharrtell is going to complain I think!  (because JSTS has problems with overloading...)

jiayuasu: Haha, all right. Let me push the private static method for now.

bjornharrtell: Might work. I resolve the overload by runtime inspection of the number value using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger. (yikes...)

dr-jts: Great.  Let's stick with the overloaded method name then.

jiayuasu: OK. So what should I do for this PR now?

dr-jts: I think the code is good.  Give us a bit of time to finalize the acceptance.

dr-jts: If you like you could add a function to the TestBuilder to expose the kNN functionality.  I added a NN function just now (see last commit to master)

jiayuasu: @dr-jts I exposed the kNN functionality following your NN function. Hope it is OK. 😁

modules/app/src/main/java/org/locationtech/jtstest/function/SpatialIndexFunctions.java
```diff
@@ -117,6 +117,14 @@ public static Geometry strTreeNN(Geometry geoms, Geometry geom)
     return (Geometry) result;
   }
 
+  public static Geometry[] strTreeNNk(Geometry geoms, Geometry geom)
```
dr-jts: You can add int k as a parameter to this method, and it will be exposed in the TestBuilder UI

modules/app/src/main/java/org/locationtech/jtstest/function/SpatialIndexFunctions.java
```diff
@@ -117,6 +117,14 @@ public static Geometry strTreeNN(Geometry geoms, Geometry geom)
...
+  {
+    STRtree index = buildSTRtree(geoms);
+    int k = 10;
+    Object[] result = index.nearestNeighbour(geom.getEnvelopeInternal(), geom, new GeometryItemDistance(), k);
+    return (Geometry[]) result;
```
dr-jts: Not sure TestBuilder handles Geometry[] results.  I'll have to check this, or perhaps you can run it and see.

It should be enhanced to automatically convert them to GeometryCollections, but that's not in place yet.

jiayuasu: @dr-jts I converted them to GeometryCollection manually because the TestBuilder doesn't take Geometry[].

modules/app/src/main/java/org/locationtech/jtstest/function/SpatialIndexFunctions.java
jiayuasu: @dr-jts I just try to add "int k" as a parameter. But it seems TestBuilder UI doesn't take more than 2 parameters. It only has two input areas. Not sure about this now.

modules/app/src/main/java/org/locationtech/jtstest/function/SpatialIndexFunctions.java
```diff
@@ -117,12 +117,19 @@ public static Geometry strTreeNN(Geometry geoms, Geometry geom)
...
-  public static Geometry[] strTreeNNk(Geometry geoms, Geometry geom)
+  public static Geometry strTreeNNk(Geometry geoms, Geometry geom)
   {
     STRtree index = buildSTRtree(geoms);
-    int k = 10;
-    Object[] result = index.nearestNeighbour(geom.getEnvelopeInternal(), geom, new GeometryItemDistance(), k);
-    return (Geometry[]) result;
+    int k = 2;
+    Object[] knnObjects = index.nearestNeighbour(geom.getEnvelopeInternal(), geom, new GeometryItemDistance(), k);
+    Geometry[] knnGeoms = new Geometry[k];
```
dr-jts: Follow the pattern here:  https://github.com/locationtech/jts/blob/master/modules/app/src/main/java/org/locationtech/jtstest/function/SpatialIndexFunctions.java#L96

modules/app/src/main/java/org/locationtech/jtstest/function/SpatialIndexFunctions.java
dr-jts: You can add more parameters.  See this example:  

https://github.com/locationtech/jts/blob/master/modules/app/src/main/java/org/locationtech/jtstest/function/CreateShapeFunctions.java#L73

modules/app/src/main/java/org/locationtech/jtstest/function/SpatialIndexFunctions.java
jiayuasu: @dr-jts OK. It fixed now, 😁

Issue #69, opened 2016-12-13
Feature suggestion: RadialDistanceSiplifier


HarelM:
Hi, I posted the same enhancement request here:
https://github.com/NetTopologySuite/NetTopologySuite/issues/142
But I was told I should post it here or in the mailing list.
In short, a very simple algorithm to simplify a line according to distance between points.
.Net code can be found there, I'm not a very experienced Java developer so I don't really think I should do a pull request to this repository, having said that, It would be nice to get this :-)

dr-jts:
A nice idea.  It would take a bit of work to make this a production-quality
operation in JTS, so unless funding is available to do this it will have to
wait for time to be available .

dr-jts:
You could also open an issue on the JTS project for this, to help in
tracking this.

HarelM:
I'm sorry, I'm new to this fine project, can you elaborate a bit?
In term of funding, I guess I can't help a lot, I'm writing my own open-source-probably-not-gonna-earn-me-a-dime project...
Also I'm not sure how changes here are converted to the .Net project which is what I actually need :-/...

dr-jts:
Make a GitHub issue under the JTS project in the same way you did for NTS.
 (No need to copy the code, though!)

dr-jts:
Once the code makes it way into JTS, it's up to the NTS project to port it.

HarelM:
Isn't this the GitHub project for JTS? Isn't this an issue without the code? I'm not sure I'm following... :-/

dr-jts:
Whoops, yes, sorry, this is the right place for this.

PR #83, opened 2017-02-28
Added RadialDistanceByAngleSimplifier (#69) to jts-lab module

FObermaier:
Added RadialDistanceByAngleSimplifierTest

### DIFF
--- a/modules/lab/src/main/java/org/locationtech/jtslab/simplify/RadialDistanceByAngleSimplifier.java
+++ b/modules/lab/src/main/java/org/locationtech/jtslab/simplify/RadialDistanceByAngleSimplifier.java
@@ -1,0 +1,249 @@
+/*
+ * Copyright (c) 2016 Vivid Solutions.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v. 1.0 which accompanies this distribution.
+ * The Eclipse Public License is available at http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ */
+package org.locationtech.jtslab.simplify;
+
+import org.locationtech.jts.algorithm.Angle;
+import org.locationtech.jts.geom.Coordinate;
+import org.locationtech.jts.geom.CoordinateSequence;
+import org.locationtech.jts.geom.CoordinateSequenceFactory;
+import org.locationtech.jts.geom.CoordinateSequences;
+import org.locationtech.jts.geom.Geometry;
+import org.locationtech.jts.geom.LineString;
+import org.locationtech.jts.geom.Lineal;
+
+
+    /**
+     * This class simplifies a line string by removing points that create a sharp turn within a radius from previous point.
+     * This class follows the JTS simplifier pattern
+     *
+     * @author 
+     *   Harel M (https://github.com/harelm)
+     * @since 1.15  
+     */
+    public class RadialDistanceByAngleSimplifier
+    {
+        private Geometry _geometry;
+        private double _distanceTolerance;
+        private double _angleTolerance;
+
+        /**
+         * Sets the radial distance tolerance. The lower it is the less simplified the line will be.
+         * The default is {@code 0}.
+         */
+        public void setDistanceTolerance(double value) { 
+          _distanceTolerance = value; 
+        }
+        
+        /**
+         * The angle tolerance - the lower the number will be the less simplified the line will be.
+         * Unit is in radiant.
+         */
+        public void setAngleTolerance(double value) { 
+          _angleTolerance = Angle.normalizePositive(value); 
+        }
+
+        /**
+         * This simplifies a line by getting the radial distance tolerance and angle distance tolerance
+         * 
+         * @param geometry The geometry to simplify
+         * @param distanceTolerance The radial distance tolerance.
+         * @param angleTolerace The angle tolerance in radiant.
+         * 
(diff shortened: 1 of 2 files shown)


dr-jts:
Not sure about the provenance and sign-off on this code.  @jnh5y any thoughts?  The code apparently authored by harelm, but commited by FObermaier - is that acceptable?

modules/lab/src/test/java/org/locationtech/jtslab/simplify/RadialDistanceByAngleSimplifierTest.java
```diff
@@ -0,0 +1,123 @@
+package org.locationtech.jtslab.simplify;
+
+import org.locationtech.jts.geom.Coordinate;
```
jnh5y: This file needs a license header.

modules/lab/src/main/java/org/locationtech/jtslab/simplify/RadialDistanceByAngleSimplifier.java
```diff
@@ -0,0 +1,249 @@
...
+import org.locationtech.jts.geom.LineString;
+import org.locationtech.jts.geom.Lineal;
+
+
+    /**
+     * This class simplifies a line string by removing points that create a sharp turn within a radius from previous point.
+     * This class follows the JTS simplifier pattern
+     *
+     * @author 
+     *   Harel M (https://github.com/harelm)
```
jnh5y: @FObermaier can you say more about the provenance here?  If you can link to the project, that might help.  Generally, if we get a quick 'ok' from the original author for creating the derivative work, that'd be best.

modules/lab/src/main/java/org/locationtech/jtslab/simplify/RadialDistanceByAngleSimplifier.java
```diff
@@ -0,0 +1,249 @@
+/*
+ * Copyright (c) 2016 Vivid Solutions.
```
mprins: This copyright is incorrect

HarelM:
Quick "OK" 😊
Go ahead, feel free to use it. Happy I could contribute to this awesome project!

modules/lab/src/main/java/org/locationtech/jtslab/simplify/RadialDistanceByAngleSimplifier.java
HarelM: Original code can be seen here:
https://github.com/IsraelHikingMap/Site/blob/master/IsraelHiking.API/Services/RadialDistanceByAngleSimplifier.cs
You have my blessing.

HarelM:
Also worth mentioning somewhere that this simplifier works better after Douglas-Packard simplification from my experience, not sure where you'd put such comment.

dr-jts:
@HarelM That would be appropriate for the class Javadoc.  Is it possible to quantify/qualify how it is better?  E..g better performance, or better quality of output?

HarelM:
Well basically both, but mainly better quality output.
Since this is a ```O(n)``` simplifier the less points it has the faster it will perform and D-P reduces points.
But this simplifier keeps sharp angles in case the segment after the turn is long enough. the reason for this is to remove GPS points that are measured in the same location. If D-P is ran before it, sharp turns will be kept.
When in action the following will be simplified (see the original blue line and the simplified red line):
simplified

For some reason git hub won't let me upload an image that shows a case where it was no simplified since the segment was long enough.

jnh5y:
@FObermaier, in order to accept this PR, you'd need to fill out the Eclipse Contributor Agreement for  felix.obermaier@netcologne.de.  The paperwork is here: https://www.eclipse.org/legal/ECA.php.

FObermaier:
@jnh5y I just did that.

jnh5y:
@FObermaier thanks!  @dr-jts back to you for the regularly scheduled PR review:).

modules/lab/src/main/java/org/locationtech/jtslab/simplify/RadialDistanceByAngleSimplifier.java
```diff
@@ -0,0 +1,269 @@
...
+    /**
+     * This simplifies a line by getting the radial distance tolerance and angle distance tolerance
+     * 
+     * @param geometry The geometry to simplify
+     * @param distanceTolerance The radial distance tolerance.
+     * @param angleTolerace The angle tolerance in radiant.
+     * 
+     * @return A simplified {@link Geometry}
+     */
+    public static Geometry Simplify(Geometry geometry, double distanceTolerance, double angleTolerace)
```
dr-jts: This method name should start with lowercase.

dr-jts:
Is there any support for changing the name of this algorithm to something slightly more descriptive?  Say `ShortSharpAngleSimplifier` ?

HarelM:
I have no objection to the new name, it sounds more descriptive and less confusing indeed.
My original name was based on the fact that I used the concept of radial simplifier and that it also simplfies by angle, thus this not so good name. Note that the angle is a parameter so it can simplify non sharp angles too, but that's just a technicality...

FObermaier:
I don't have any objections either. Do you want me to refactor it?

dr-jts:
@FObermaier   Would be great if you can do the rename.

FObermaier:
> @dr-jts dr-jts assigned FObermaier on 9 May

@dr-jts , I'm not sure, is there anything in particular you want me to do with this?

HarelM:
In case anyone is interested, I wrote a short wiki page about the algorithm to find unmapped OSM routes. it uses this simplifier.
https://github.com/IsraelHikingMap/Site/wiki/Candidates-algorithm:-find-unmapped-routes

PR #84, opened 2017-02-28
Add link to Javadocs in README

fosskers:
I didn't notice an easy link to it elsewhere (i.e. the User Guide) other than from its LT page.


dr-jts:
Good point.  This might be better under the Documentation heading.

fosskers:
Woops, you're right. Fixed now.

PR #90, opened 2017-03-18
Fix ScaledNoder behaviour

jnh5y:
It was not consistent between scale=1 (no rounding) and other scales (rounding).
Now, it skip scaling process for 0.0 scale, but not for 1.0 scale.

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/noding/ScaledNoder.java
+++ b/modules/core/src/main/java/org/locationtech/jts/noding/ScaledNoder.java
@@ -49,10 +49,10 @@ public class ScaledNoder
     this.noder = noder;
     this.scaleFactor = scaleFactor;
     // no need to scale if input precision is already integral
-    isScaled = ! isIntegerPrecision();
+    isScaled = ! isInfinitePrecision();
   }
 
-  public boolean isIntegerPrecision() { return scaleFactor == 1.0; }
+  public boolean isInfinitePrecision() { return scaleFactor == 0.0; }
 
   public Collection getNodedSubstrings()
   {


dr-jts:
No unit tests?  Not sure what original problem was?  Is this really a non-breaking change?

mukoki:
Hi Martin,

Here is a small beanshell script written with OpenJUMP to demonstrate the problem 👍 

```
import com.vividsolutions.jts.io.WKTReader;
import com.vividsolutions.jts.noding.*;
import com.vividsolutions.jts.noding.snapround.*;
import com.vividsolutions.jts.algorithm.RobustLineIntersector;

// create two intersecting linestrings with coordinates having a precision of 0.0001
l1 = new WKTReader().read("LINESTRING(0 0, 20.11111 30.11111)");
l2 = new WKTReader().read("LINESTRING(20.11111 0, 10.11111 30.11111)");
nss1 = new NodedSegmentString(l1.getCoordinates(),"");
nss2 = new NodedSegmentString(l2.getCoordinates(),"");

noder = new MCIndexSnapRounder(new PrecisionModel(1));
// Create a ScaledNoder with a scale factor of 100 
snoder = new ScaledNoder(noder, 100);

Collection coll = new ArrayList();
coll.add(nss1);
coll.add(nss2);

snoder.computeNodes(coll);
print(snoder.getNodedSubstrings());
// -> [LINESTRING (0.0 0.0, 13.43 20.11), LINESTRING (13.43 20.11, 20.11 30.11), LINESTRING (20.11 0.0, 13.43 20.11), LINESTRING (13.43 20.11, 10.11 30.11)]
// Intersection coordinate AND input coordinates are rounded to 0.01
// Note that here, the rounding performed by MCIndexSnapRounder is relative to already scaled coordinates, but this is not the point.

// now do the same with new ScaledNoder(noder, 1);
// -> [LINESTRING (0.0 0.0, 13.0 20.0), LINESTRING (13.0 20.0, 20.11111 30.11111), LINESTRING (20.11111 0.0, 13.0 20.0), LINESTRING (13.0 20.0, 10.11111 30.11111)]
// Intersection coordinate is rounded to 1 unit (13 20) BUT input coordinates are not rounded as in the previous case
// That's because the rounding is not performed when scaling is 1. I think scaling of input coordinates should behave the same way whatever the scale factor is.
```
Michaël

mukoki:
Hi Martin, 
As stated in the above example, with a ScaledNoder with a scale different from 1, all input coordinates are rounded to the precision defined by the noder (not only intersection points)
With a ScaledNoder  with scale = 1 is rounding the intersection point, but other input coordinates are left with their original precision.
My point is that the behaviour should not depend on the scale factor.  Of course, it will change the result of some operations (computation made with a ScaleNoder of factor 1).
If you confirm to me that a ScaledNoder with scale 1 should round all input coordinates to integers I can try to write a breaking unit test showing the problem.

modules/core/src/main/java/org/locationtech/jts/noding/ScaledNoder.java
```diff
@@ -49,10 +49,10 @@ public ScaledNoder(Noder noder, double scaleFactor, double offsetX, double offse
     this.noder = noder;
     this.scaleFactor = scaleFactor;
     // no need to scale if input precision is already integral
-    isScaled = ! isIntegerPrecision();
+    isScaled = ! isInfinitePrecision();
   }
 
-  public boolean isIntegerPrecision() { return scaleFactor == 1.0; }
+  public boolean isInfinitePrecision() { return scaleFactor == 0.0; }
```
jodygarnett: This fix changes public API, indeed changes the meaning.
Please advise?

mukoki:
You are right. It is probably not very safe to remove a public method.
I don't really change the meaning because I changed the method name, but we can also keep isIntegerPrecision() method to be on the safe side (I think it will not be used in the whole JTS API though).
The motivation for this change is that I don't want a special behaviour for ScaledNoder when scaleFactor == 1.0.
Hope it makes sense. I tried to demonstrate why I think the current behaviour is not correct in the case of scaledFactor == 1.0. If you or Martin agree about my explanations, I can try to add a unit test.

grootstebozewolf:
Thanks for leaving this draft open — scoped from the NetTopologySuite.Proofs snap / precision lane against current master + NTS 2.6.

### What we checked
- Master `ScaledNoder`: `isScaled = !(scaleFactor == 1.0)` (NTS twin identical).
- This PR: skip only when `scaleFactor == 0.0`, so **scale=1 would pre-round**.
- mukoki’s 2017 beanshell (MCIndexSnapRounder + scale 1 vs 100) — historical mixed precision (integer **nodes**, full-precision **ends** at scale=1).
- Modern production path: `BufferOp` → `ScaledNoder(SnapRoundingNoder(PM(1)), fixedPM.getScale())`.

Write-up + MRE: [jts-90-scalednoder-lane](https://github.com/grootstebozewolf/NetTopologySuite.Proofs/blob/main/docs/jts-90-scalednoder-lane-2026-08.md)  
(`dotnet run --project tests/Discussion839Mre -c Release -- --jts90`)

### Results
| Check | Result |
|---|---|
| Mukoki mixed-precision under **SnapRoundingNoder** | **Masked** — SRN snaps vertices; scale=1 output is fully on the integer grid |
| `Math.round(x * 0)` if master treats scale=0 as scaled | **Live footgun** — finite inputs collapse to the origin |
| A priori scale=1 vs skip+SRN on mukoki lines | Bit-equal on this toy; not a general topology proof |
| Named corpus snap claim gap? | **No** — wrapper policy / scale=0 safety, not `round(x·s)/s` algebra |

### Concern with merging as written
`SnapRoundingNoder` documents that **a priori vertex rounding can distort** snap-rounding topology; BufferOp only uses `ScaledNoder` as an optional speed path and notes topology risk. Forcing scale at **scale=1** turns the intentional pass-through into a priori integer round, which fights that design (and renames a public method without a dual).

### Suggested narrow alternative
1. Keep `isIntegerPrecision()` (`scale == 1`).
2. Treat **`scale == 0` as not scaled** (or throw) so FLOATING-scale callers never hit `round(x*0)`.
3. Do **not** invert scale=1 unless there is a remaining MCIndexSnapRounder-only call site that still needs it — with tests.
4. Add unit tests for scale 0 / 1 / 100 grid membership under `SnapRoundingNoder`.

Happy to re-run the MRE if you land a revised patch.

*(Affiliation: [NetTopologySuite.Proofs](https://github.com/grootstebozewolf/NetTopologySuite.Proofs) — formal snap/precision companion, not a JTS committer review.)*

PR #92, opened 2017-03-18
Add better reading support of WKB / WKT

jnh5y:
Following the OGC 06-103r4 spec (useful for geopackage reading + unit tests).


dr-jts:
@jnh5y  Looks good to merge.

jratike80:
@jnh5y, is there still something to check before merging? GeoPackages are coming more popular and for example brand new GeoServer 2.12 has support for GeoPackages in core but it fails with XYZ geometries because JTS does not know what to do with them.

dr-jts:
I will merge this.  Just one comment - it would be good to have the Javadoc updated so it specifies what syntax can be read.  That's the only way for users to know what can be handled.

dr-jts:
More complete unit tests for WKT would be good too.  Yeah, I know it's a drag...

mukoki:
Hi Martin, I added some javadoc and some more tests for WKT Reader.
Let me know if you want more precisions.
Michaël

jnh5y:
Committed this as 1d7970532e9f3cc9c15ec4705134c9ba1eee031e

Issue #104, opened 2017-06-30
Wrong assumption in CascadedPolygonUnion


strk:
Debugging an issue with CascadedPolygonUnion I found an unsafe assumption in `unionUsingEnvelopeIntersection` method of `CascadedPolygonUnion` class.

The method extracts, from 2 multipolygons, the ones that intersect an envelope and the ones that do not (disjointPolys). Then unions the intersecting ones togheter and finally combine the unioned intersecting polygons *and* the disjoint polygons, assuming that they would not be intersecting.

This is a wrong assumption, in that the union of the intersecting polygons may slightly move the envelope due to use of `SnapIfNeededOverlayOp` in `Geometry.union()`.

strk:
Actually, it's not just about the Envelope growing, which may still be handled relatively easily, but generally about segments moving.

Consider a multipolygon with a very small space between two adjacent components.
One of the components is tall enough to be in the common envelope, the other is outside.
When unioning the tall one with other elements, it ends up overlapping the previously nearby-but-disjoint polygon, due to SnapOp heuristic.

strk:
@dr-jts I've an XML test exposing this problem but did not find how to run it in JTS.
Asked on the mailing list but I'm afraid that list is not really used as I'm the only poster ?
https://dev.locationtech.org/mhonarc/lists/jts-dev/

Specific message: https://dev.locationtech.org/mhonarc/lists/jts-dev/msg00167.html

Shall I still use the sourceforge mailing list instead ?

bjornharrtell:
It might be low-volume but you are not the only poster. :) https://dev.locationtech.org/mhonarc/lists/jts-dev/threads.html

strk:
Thanks! Where had you found that  'threads' link ? :)

strk:
XML tester for the issue is here: https://trac.osgeo.org/geos/raw-attachment/ticket/837/bug837.xml

I cannot find a way to run that single XML test with JTS so dunno what happens there.

strk:
For the record, this bug is confirmed in GEOS (but GEOS uses a different heuristic):
https://trac.osgeo.org/geos/ticket/837

bjornharrtell:
Found testrunner.sh is still here at https://github.com/locationtech/jts/blob/master/bin/testrunner.sh but it won't work at least because of namespace change and modularization of the source code.

strk:
Ah, thanks. How hard would it be to fix it ?

strk:
[...]

PR #106, opened 2017-07-14
Fix testrunner.sh to work with current code layout

strk:
See https://dev.locationtech.org/mhonarc/lists/jts-dev/msg00167.html
and https://github.com/locationtech/jts/issues/104#issuecomment-314463753

\cc @bjornharrtell

### DIFF
--- a/bin/testrunner.sh
+++ b/bin/testrunner.sh
@@ -1,14 +1,14 @@
 #!/bin/sh
 
 if test "x$JTS_LIB_DIR" = "x"; then
-	JTS_LIB_DIR=`dirname $0`/../lib/
+	JTS_LIB_DIR=`dirname $0`/../modules/
 fi
 
 #---------------------------------#
 # dynamically build the classpath #
 #---------------------------------#
 THE_CLASSPATH=
-for i in `ls ${JTS_LIB_DIR}/*.jar`
+for i in `find ${JTS_LIB_DIR} -name '*.jar'`
 do
   THE_CLASSPATH=${THE_CLASSPATH}:${i}
 done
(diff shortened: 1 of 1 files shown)


bjornharrtell:
Nice! I tried and failed (with minimal effort..)

strk:
It is very usefu for mel to be able to run the same XML test with JTS (testrunner) and with GEOS (XMLTester) for debugging discrepancies. Actually I might be the original author of that script (not sure as history was trimmed in the move from SVN to GIT - after having been trimmed in the past during the move from CVS to SVN  ...)

bjornharrtell:
FYI the trimming was intentional to be able to re-license the code. Not saying that's entirely optimal but at least understandable.

strk:
interesting, like cutting a dog's leash to claim ownership of it ?

bjornharrtell:
AFAIK previous contributors have given permission for the licence change for the current code, but it was more or less impossible to verify for each historic version resulting in the decision to wipe history. But I'm not an authoritative source on the subject this is just my outside observations.

strk:
guess I could move testrunner.sh to GEOS codebase and drop from here, as I didn't like the license change either ..

strk:
There, moved to GEOS:
https://trac.osgeo.org/geos/changeset/b7ec17b493016dc0cf628682445e1041a2d9e096/git

The script can be removed from this repository.

strk:
For historical reference: https://sourceforge.net/p/jts-topo-suite/code/787/log/?path=/trunk/jts/bin/testbuilder.sh

dr-jts:
Agreed that including the original JTS scripts would be useful.

Probably they should reference the artifacts created in the module  `target `directories (although I'm no Maven expert, so I will confirm this).

I'll make an issue for this.

Issue #107, opened 2017-07-28
Invalid union result from valid polygon inputs


strk:
I found a case in which the union of two valid small polygons result in an invalid output.
NOTE I had to tweak the JTS testrunner slightly to get the HEXWKB printed in order to check for validity, I might send a PR to help with this in the future.

The XML test file can be found on https://trac.osgeo.org/geos/ticket/838 (WARNING: the test *expects* the invalid result, so you'll need more care to check the bug)

mukoki:
Hi strk,

In ticket  https://trac.osgeo.org/geos/ticket/838, you mention that the test has failed since
https://github.com/locationtech/jts/commit/8e6abedeca3f3c1905c2f58c0a2dccc77f1f2f12

Did you notice that your test worked before this commit ? It seems to me that the union of your polygons already produces an invalid polygon with JTS 1.14 (just tested from PostGIS/OpenJUMP).
More over, the commit should just be about z/m flag interpretation in jts

Your polygons have points located at just 1 ulp from each other, which is the main cause of failure in this kind of operation
x = 945169.6                   y= 1937162.2
and 
x = 945169.6000000001  y= 1937162.2

Michaël M

strk:
I just reported the commit of JTS I tested against, didn't try any 
previous version. I suspect this also failed in earlier versions.

Unfortunately the GIT repository of JTS does not have all the older
history, which makes things harder to bisect.

Yes, polygons are very close to each other, but are valid on
themselves, and the OverlayOp has a noding validator which is
evidently missing the collapse

dr-jts:
It's definitely interesting that the NodingValidator did not catch the failure.  I'll try and look into why that is.

This is the kind of problem that Snap Rounding will solve, if and when it gets implemented.

dr-jts:
I *think* what is happening is that the Overlay snapping heuristic is failing to snap correctly.  The Noding Validator checks whether noding is correct, but not that the result topology is correct.  I guess in this case the noding is correct, but the final topology is assembled incorrectly.

dr-jts:
BTW, you might want to determine the correct expected result and put that in the test case.  That way the TestRunner can detect the failure and report appropriately.  (And when the output is made input-format-sensitive, a full-precision WKB actual result will be reported.)

strk:
Just a note: on the GEOS side there's no snapping heuristic taking
place (ie: SnapOverlayOp is not used) - OverlayOp itself outputs
the invalid geometry.

[...]

PR #108, opened 2017-07-28
Allow requesting WKB output from jts tester

strk:
Lets you export a JTSTEST_WKB_OUTPUT variable to obtain HEXWKB
print of results.

Helps debugging robustness issues, whereas the obtained result could
otherwise become invalid due to loss of precision upon converting to
WKT.

The change was needed to confirm invalid output obtained from the
test for bug reported as #107

### DIFF
--- a/modules/tests/src/main/java/org/locationtech/jtstest/testrunner/GeometryResult.java
+++ b/modules/tests/src/main/java/org/locationtech/jtstest/testrunner/GeometryResult.java
@@ -14,6 +14,7 @@
 package org.locationtech.jtstest.testrunner;
 import org.locationtech.jts.geom.Geometry;
 import org.locationtech.jts.io.WKTWriter;
+import org.locationtech.jts.io.WKBWriter;
 
 
 /**
@@ -21,9 +22,14 @@ import org.locationtech.jts.io.WKTWriter;
  */
 public class GeometryResult implements Result {
   private Geometry geometry;
+  private boolean outputwkb = System.getenv("JTSTEST_WKB_OUTPUT") != null;
 
   public GeometryResult(Geometry geometry) {
     this.geometry = geometry;
+  }
+
+  public void setWKBOutput(boolean newval) {
+    this.outputwkb = newval;
   }
 
   public Geometry getGeometry() {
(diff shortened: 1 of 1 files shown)


strk:
Travis fails for some obscure reason related to myself having an ECAs (what?) and lacking a Signed-of-by...

strk:
Is there anything I can do to make this patch accepted ? It's discouraging to get no feedback in one week here...

dr-jts:
See the Contributor Guide for some info about how to make an acceptable PR.

dr-jts:
As far as the code goes, it would be better to expose this as a command-line arg to the TestRunner.   Reading the value of an environment variable deep within the codebase is not great practice, IMO.

dr-jts:
The functionality of outputting WKB is useful, though.

It's sometimes nice to find a solution which doesn't involve arcane flags.  How about this:  if the test XML has expected output specified in WKB, then out the actual value in WKB as well?

strk:
My goal here is to make it compatible with the GEOS XMLTester,
which has a flag. I've added a JTSXMLTester.sh script that handles
the XMLTester compatible flag  (--wkb-output) turning it into an env
variable.

I understand you don't like the env variable, but it's the least
invasive solution. Exposing a proper flag would take augmenting the
interface of the test result handler

jnh5y:
@strk have you had a chance to fill out the Eclipse Contributor Agreement?  https://github.com/locationtech/jts/blob/master/CONTRIBUTING.md#become-a-contributor

Let us know when you've had a chance to do that.  The Eclipse webhook for GitHub PRs is a bit touchy.  It is also looking for all the commits on the PR to be signed off (git commit -s).

dr-jts:
@strk any reason why you closed this?  It seems like a reasonable facility.  

Although the new JtsOp CLI might be an equivalent or more powerful way to accomplish this?

strk:
I was discouraged by the contributor agreement

jnh5y:
@strk what about the contributor agreement was discouraging?

From my understanding, the Eclipse Contributor Agreement is trying to cover two bases:  first, that you know that you are contributing to an open source project and second, that you are acknowledging that your contribution is yours and that you can contribute it freely.  

Given that you routinely contribute to open source, I doubt those concerns are at issue.  If Eclipse's policy is discouraging you, I'd like to discuss it.  I'm a LocationTech committer rep to various parts of Eclipse, and I'm more than happy to raise your concerns.

PR #110, opened 2017-08-02
Quadtree emptyness

mukoki:
Quadtree maybe empty even if root is not null. Test if root is empty instead.
NodeBase#isEmpty can return as soon as an element is found in the tree.


mukoki:
Can someone explain to me why the pull request did not pass Eclipse validation ?
Details say that I have invalid Signed-off-by footers ?

dr-jts:
No idea why this fails ip-validation.  How about I submit the PR on your behalf?

mukoki:
Martin,
Yes, if you can do that to apply the fix, go ahead.

jnh5y:
@mukoki I think you were just missing to commit with '-s', i.e., "git commit -s".  I've committed for you.

PR #118, opened 2017-08-19
Improve MultiPolygon centroid robustness

dbaston:
@jnh5y as we discussed the other day: this PR improves the robustness of centroid computations for MultiPolygons with disparate components.  It is a very late resolution to the topic discussed in this thread: https://sourceforge.net/p/jts-topo-suite/mailman/message/31595183/


jnh5y:
@dbaston great!  There are a few things to iron out around contributing to an Eclipse project.  From the https://github.com/locationtech/jts/blob/master/CONTRIBUTING.md, you'll need to sign the ECA (https://www.eclipse.org/legal/ECA.php) and 'sign' each git commit with 'git commit -s'.

modules/core/src/main/java/org/locationtech/jts/algorithm/Centroid.java
```diff
@@ -153,7 +153,6 @@ else if (ptCount > 0){
 
   private void setBasePoint(Coordinate basePt)
   {
-    if (this.areaBasePt == null)
```
jnh5y: *interesting*  This fixes the issue?

dbaston: For the test case I included, yes.

jnh5y:
@dr-jts, any input on this?  Looks good to me.  Dan explained it in person and it sounds reasonable.

dr-jts:
The fix looks good.  I am going to make some comments on the PR.

@dbaston are you going to fill out an ECA?  Seems worthwhile, since you've been so productive with JTS enhancements.

modules/core/src/test/java/org/locationtech/jts/algorithm/CentroidTest.java
```diff
@@ -0,0 +1,39 @@
...
+            Geometry component = g.getGeometryN(i);
+            double areaFraction = component.getArea() / totalArea;
+
+            Coordinate componentCentroid = component.getCentroid().getCoordinate();
+
+            cx += areaFraction * componentCentroid.x;
+            cy += areaFraction * componentCentroid.y;
+        }
+
+        assertEquals(new Coordinate(cx, cy), g.getCentroid().getCoordinate());
```
dr-jts: This seems highly susceptible to round-off issues.  Does it really work in this case?  

Would be better to use a precision tolerance in the comparison, I think.  

Also would be good to factor the area-weighted algorithm out into a separate method, both to highlight that it is an alternative algorithm to the current polygon centroid computation, and to allow reusing it for other test cases.  (Given this nice addition of a Centroid unit test, it would be nice to add more complete coverage to the tests it contains.  Also, that would provide better coverage of the area-weighted algorithm - which of course also needs to be "proven out" by some testing!)

modules/tests/src/test/resources/testxml/general/TestCentroid.xml
```diff
@@ -242,6 +242,30 @@
...
+     -92.65560500000008 36.58708800000005
+    )),
+    ((
+     -92.65512450000065 36.586800000000466,
+      -92.65512449999994 36.58680000000004,
+     -92.65512449998666 36.5867999999905,
+      -92.65512450000065 36.586800000000466
+    )))
+  </a>
+  <test><op name="getCentroid" arg1="A" >POINT (-92.6553838608954 36.58695407733924)</op></test>
```
dr-jts: A similar comment here.  This seems brittle WRT numeric precision.  Not your problem, but it would be nice if the XML test syntax allowed defining a precision tolerance for expected results.  (For centroid computation the tolerance can be quite fine, since the algorithm should produce very close results).

modules/core/src/test/java/org/locationtech/jts/algorithm/CentroidTest.java
dbaston: I can factor the weighted-average centroid out for use in other tests, but the intent wasn't to propose it as a separate algorithm for general use.  The intent with this test was only to show that the change to the base algorithm (moving the triangle base point for each polygon shell) produces results that are equivalent to the area-weighted average described in the linked email chain.

The results satisfy `Coordinate::equals` in this case but may suffer from round-off discrepancies with other inputs.  I can change the assertion to check that the distance between the two centroids is < 1e-14 or something, if that is preferable.

dr-jts: I just mean factoring it out as a method in the CentroidTest class.  That shows that it is useful for building tests, but doesn't promote using it for actual computation.

Adding a tolerance would be great.  Note that Coordinate provides an equals2D-with-tolerance method.

modules/tests/src/test/resources/testxml/general/TestCentroid.xml
dr-jts: Since the XML Test format doesn't provide a way of expressing a tolerance value, this is fine as it stands.  If it breaks in the future, that will be motivation to add a tolerance!

dr-jts:
@dbaston would the same robustness issue occur if the polygons in the example were holes in a larger polygon?  In this case I suspect the area base point should be reset for each ring?

dbaston:
Yes, it intuitively seems beneficial to reset the base point for each ring.  I don't have a test case showing failure without doing this, but I can make the change anyway if you like.

modules/core/src/test/java/org/locationtech/jts/algorithm/CentroidTest.java
dbaston: Sounds good.  I've made these changes.

modules/core/src/test/java/org/locationtech/jts/algorithm/CentroidTest.java
```diff
@@ -0,0 +1,50 @@
+package org.locationtech.jts.algorithm;
+
+import junit.framework.TestCase;
+import org.locationtech.jts.geom.Coordinate;
+import org.locationtech.jts.geom.Geometry;
+import org.locationtech.jts.io.WKTReader;
+
+public class CentroidTest extends TestCase {
+
+    private static final double tolerance = 1e-10;
```
dr-jts: A minor style point - I like to make static final fieldnames uppercase, so it's obvious that they are constants.

dr-jts:
@dbaston re setting base point for holes: on further reflection I suspect that the computation would be dominated by the larger enclosing shell, so that there's probably no need to worry about the affect of small holes.  Without an example demonstrating the issue there's no need to make a fix for this.

modules/core/src/main/java/org/locationtech/jts/algorithm/Centroid.java
```diff
@@ -153,7 +153,6 @@ else if (ptCount > 0){
 
   private void setBasePoint(Coordinate basePt)
```
dr-jts: How about we rename this to setAreaBasePoint?  Makes it a teeny bit more descriptive, and indicates that this functionality is only necessary for areas.

PR #121, opened 2017-08-22
Rename TestBuilder fatjar and update bat script to run it.

dr-jts:
Signed-off-by: Martin Davis <mbdavis@VividSolutions.com>


dr-jts:
Will resubmit with valid signoff.

PR #134, opened 2017-08-28
Added favicon.ico

lossyrob:
This will ideally serve out a favicon to the gh-pages site, as well as the javadocs.

Another solution is to place 

```
    <link rel="icon" type="image/png" href="jts_logo.png">
```

In the HTML pages; however because the javadoc HTML is autogenerated this only works for some of the docs pages.

If we merge this and this doesn't work with github pages I can take the change out.


dr-jts:
Don't think I see this working?

PR #139, opened 2017-08-30
Shorten name for TestBuilder JAR

### LINKED ISSUE #61: Shorten the jar-with-dependencies name.
I believe jar-with-dependencies is a default profile/type in the Maven assembly plugin.  For this issue, we'll shorten the name as possible/appropriate.

dbaston:
This should resolve #61

### DIFF
--- a/bin/testbuilder.bat
+++ b/bin/testbuilder.bat
@@ -10,4 +10,4 @@ set JAVA_LOOKANDFEEL=
 REM Default app options can be added here
 set APP_OPTS=%*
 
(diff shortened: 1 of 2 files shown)


dr-jts:
Nice.  It would be good to add to the commit the fix to the testbuilder.bat file as well.

https://github.com/locationtech/jts/blob/master/bin/testbuilder.bat

dbaston:
Good catch; just force-pushed an update.

dr-jts:
Can do the same for the JTSTestRunner too.

PR #152, opened 2017-09-04
Convert into static inner class

bjornharrtell:
The use of non-static inner classes in JTS is not part of public API, sparse and appears avoidable. They are hard to transpile so I propose refactor them away with this PR.


dr-jts:
Looks good to me.  Definitely nicer to make the dependencies explicit for inner classes.

PR #153, opened 2017-09-04
Refactor away use of functional style

bjornharrtell:
Another refactor proposal to ease transpilation. Also more in line with rest of the code.

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/noding/ScaledNoder.java
+++ b/modules/core/src/main/java/org/locationtech/jts/noding/ScaledNoder.java
@@ -12,11 +12,13 @@
 
 package org.locationtech.jts.noding;
 
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Iterator;
+import java.util.List;
 
 import org.locationtech.jts.geom.Coordinate;
 import org.locationtech.jts.geom.CoordinateArrays;
-import org.locationtech.jts.util.CollectionUtil;
 
 /**
  * Wraps a {@link Noder} and transforms its input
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/locationtech/jts/noding/ScaledNoder.java
```diff
@@ -71,15 +73,12 @@ public void computeNodes(Collection inputSegStrings)
...
-//  	System.out.println("Scaled: scaleFactor = " + scaleFactor);
-    return CollectionUtil.transform(segStrings,
-                                    new CollectionUtil.Function() {
-      public Object execute(Object obj) {
-        SegmentString ss = (SegmentString) obj;
-        return new NodedSegmentString(scale(ss.getCoordinates()), ss.getData());
-      }
-                                    }
-      );
+    List nodedSegmentStrings = new ArrayList();
```
schlosna: nit: might want to presize: `new ArrayList(segStrings.size());`

modules/core/src/main/java/org/locationtech/jts/noding/ScaledNoder.java
```diff
@@ -71,15 +73,12 @@ public void computeNodes(Collection inputSegStrings)
...
+    for (Iterator i = segStrings.iterator(); i.hasNext(); ) {
```
dr-jts: Can the transpiler handle Java 5 for/in syntax?  I think that will avoid the cast, and is a little simpler.

bjornharrtell: Not currently (AFAIK there is no for/in use currently in JTS) but I would think it should not be that hard to implement. 

But, since `segStrings` is a generic `Collection` I don't see how I can avoid casting anyway?

PR #154, opened 2017-09-04
Make ordinate indexes static final

bjornharrtell:


dr-jts:
Java interface fields are static by definition.  We can make this change, but it might be nice to handle this automatically in the transpiler.

bjornharrtell:
Ah didn't know this, should be possible to handle in transpilation fine then. Closing until further notice.

PR #155, opened 2017-09-04
Avoid using reflection

bjornharrtell:


modules/core/src/main/java/org/locationtech/jts/geom/Geometry.java
```diff
@@ -1791,8 +1791,7 @@ protected boolean isEquivalentClass(Geometry other) {
    *      but not one of its subclasses
    */
   protected void checkNotGeometryCollection(Geometry g) {
-    //Don't use instanceof because we want to allow subclasses
-    if (g.getClass().equals(GeometryCollection.class)) {
+    if (getSortIndex() == SORTINDEX_GEOMETRYCOLLECTION) {
```
dr-jts: Perhaps better to say 

`if (g.isGeometryCollection())`

That keeps the use of the SORTINDEX key localized as much as possible.

bjornharrtell: Agreed, fixed.

Issue #165, opened 2017-09-11
Perf: avoid Envelope allocation in MonotoneChainEdge


jandam:
```
  // these envelopes are created once and reused
  Envelope env1 = new Envelope();
  Envelope env2 = new Envelope();
...
    env1.init(p00, p01);
    env2.init(p10, p11);
    if (! env1.intersects(env2)) return;

```
Can be replaced with

`if (!Envelope.intersects(p00, p01, p10, p11)) return;
`

dr-jts:
Good catch.  

Did you find this by code inspection, or by noticing a hotspot in profiling?

dr-jts:
There's a couple of other places temporary envelopes are used as well.  I will try and look into removing them as well, now that the direct coordinate function is available.

dr-jts:
Design note: it would be much better if Envelopes were immutable, since changing values on possibly aliased objects is a prime source of bugs.  Originally they were made mutable due to performance concerns about object allocation, but that's much less of an issue now.  

Hard to know if it's possible to make this change to the Envelope API at this point.  But the JTS codebase could change to avoid use of mutating Envelope methods, as a first step.

jandam:
Thank you for closing this. I saw several 10 GBs allocated in TLAB during Geometry validation. I'm processing OpenStreetMap data. >400M polygons

dr-jts:
That is certainly a hotspot!  Memory use optimization is probably more important than actual performance - memory is always finite, whereas time is somewhat flexible...

cmhodgson:
> Making all geometry objects immutable is one of the goals of the JTS2 design.

This sounds fantastic. This is something I noticed when I was reviewing some of the memory-related performance problems I was having, is that everything gets copied to prevent aliasing, which could be avoided entirely be separating immutable objects from mutable ones. I do think you need to retain mutable objects for certain use cases though, as enforcing immutability in all cases can also cause excessive copying.

PR #167, opened 2017-09-11
Remove MonotoneChainOverlap temp Envelope use

dr-jts:


dr-jts:
An extension to #165

PR #184, opened 2017-09-26
Add Licensing FAQ

dr-jts:

### DIFF
--- a/FAQ-LICENSING.md
+++ b/FAQ-LICENSING.md
@@ -25,10 +25,10 @@ A2. **JTS no longer uses LGPL.  What does this mean for downstream projects?**
 JTS is now released under a dual license: EDL and EPL.  
 Dual-licensing gives downstream projects more options:
 
-* EDL is a BSD-3-style license, which is more permissive than the LGPL.  
+* The **EDL** is a BSD-3-style license, which is more permissive than the LGPL.  
   This allows a wider variety of uses; for instance, it permits iOS use of JTS for mobile development.
   
-* The EPL license provideds a similar permissiveness to LGPL with different wording around patents
+* The **EPL** provides a similar permissiveness to LGPL with different wording around patents
 
 A3. **But I thought LGPL and EPL do not mix? How does this effect derivative projects (like the LGPL GEOS project)?**
 
@@ -40,7 +40,7 @@ This will allow derivative projects like GEOS the choice of using either BSD-3-s
 A4. **How can an LGPL-licensed project (such as GEOS) contribute code to JTS?**
 
 From the JTS perspective this works in the same way as any other contribution.  
-The contributor must have an Eclipse Contributor License Agreement (CLA) in place.  
+The contributor must have an Eclipse *Contributor License Agreement* (CLA) in place.  
 For details see the [**Contributing Guide**](CONTRIBUTING.md).
 
 Code contributed to JTS must be releasable under the EDL & EPL licenses.
--- a/FAQ-LICENSING.md
+++ b/FAQ-LICENSING.md
@@ -8,10 +8,11 @@ as described [here](LICENSES.md).
 
 **Downstream projects** includes the following:
 
-* Ports of JTS to other languages 
-  * GEOS, JSTS, etc.
-* Use of JTS libraries in other codebases
-  * GeoTools, GeoServer, etc.
+* **Downstream projects** are codebases which link to JTS libraries
+  * e.g. GeoTools, GeoServer, etc.
+* **Derivative projects** are ports of JTS to other languages 
+  * e.g. GEOS, JSTS, etc.
+
 
 ## Licensing Frequently Asked Questions
 
@@ -29,18 +30,18 @@ Dual-licensing gives downstream projects more options:
   
 * The EPL license provideds a similar permissiveness to LGPL with different wording around patents
 
-A3. **But I thought LGPL and EPL do not mix? How does this effect the LGPL GEOS project?**
+A3. **But I thought LGPL and EPL do not mix? How does this effect derivative projects (like the LGPL GEOS project)?**
 
 An LGPL project (like GEOS) should use a BSD-3-style License (the BSD-3 License is compatible with the LGPL license used by GEOS).
 
 In the longer term the EPL license is being revised to work with GPL / LGPL.  
-This will allow derivative projects like GEOS the choice of either BSD-3-style or EPL license.
+This will allow derivative projects like GEOS the choice of using either BSD-3-style or EPL license.
 
 A4. **How can an LGPL-licensed project (such as GEOS) contribute code to JTS?**
 
 From the JTS perspective this works in the same way as any other contribution.  
 The contributor must have an Eclipse Contributor License Agreement (CLA) in place.  
-For details see the [Contributing Guide](CONTRIBUTING.md).
+For details see the [**Contributing Guide**](CONTRIBUTING.md).
 
 Code contributed to JTS must be releasable under the EDL & EPL licenses.
 Therefore the contributor must also have permission from the downstream project (e.g. via OSGeo CLA or similar) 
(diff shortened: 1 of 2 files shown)


dr-jts:
@jodygarnett  @jnh5y   Can you review the FAQ and see if it's accurate and if anything else should be added?

I will probably add some more links to some of the items mentioned in the FAQ.

FAQ-LICENSING.md
```diff
@@ -0,0 +1,47 @@
...
+Dual-licensing gives downstream projects more options:
+
+* The **EDL** is a BSD-3-style license, which is more permissive than the LGPL.  
+  This allows a wider variety of uses; for instance, it permits iOS use of JTS for mobile development.
+  
+* The **EPL** provides a similar permissiveness to LGPL with different wording around patents
+
+**A3. But I thought LGPL and EPL do not mix? How does this effect derivative projects (like the LGPL GEOS project)?**
+
+An LGPL project (like GEOS) should use a BSD-3-style License (the BSD-3 License is compatible with the LGPL license used by GEOS).
```
dbaston: The "BSD-3-style License" that JTS is released under is the EDL, right? Would it be correct to say:

"An LGPL project (like GEOS) should use JTS under the terms of EDL, which is compatible with the LGPL." ?

If correct, this could also avoid one misinterpreting the FAQ as suggesting that GEOS should switch its own license to BSD.

FAQ-LICENSING.md
```diff
@@ -0,0 +1,47 @@
...
+
+In the longer term the EPL license is being revised to work with GPL / LGPL.  
+This will allow derivative projects like GEOS the choice of using either BSD-3-style or EPL license.
```
dbaston: Would it be accurate to replace "BSD-3-style" with "EDL" ?

FAQ-LICENSING.md
```diff
@@ -0,0 +1,47 @@
...
+
+**A4. How can an LGPL-licensed project (such as GEOS) contribute code to JTS?**
+
+From the JTS perspective this works in the same way as any other contribution.  
+The contributor must have an Eclipse *Contributor License Agreement* (CLA) in place.  
+For details see the [**Contributing Guide**](CONTRIBUTING.md).
+
+Code contributed to JTS must be releasable under the EDL & EPL licenses.
+Therefore the contributor must also have permission from the downstream project (e.g. via OSGeo CLA or similar)
```
dbaston: Would "must have permission from the _source_ project" be more clear? (Or "contributing project"?) The term 
 "downstream" has been defined at the top of this document to mean something other than what GEOS is. And in any case, the use of "upstream" and "downstream" is a bit muddied when GEOS code is ported to JTS. More of a strongly prevailing wind than a flowing stream? :)

FAQ-LICENSING.md
dr-jts: That does sound more like the intention to me - I was wondering about the original wording myself.  @jodygarnett can you confirm this reflects the intention of this point?

FAQ-LICENSING.md
dr-jts: Yes, seems more accurate.  Although it's still a bit confusing then, because it seems to suggest that GEOS should change to use EDL or EPL, which I don't think is the intention.  

Perhaps it should probably say "*New* derivative projects" - and not mention GEOS?

FAQ-LICENSING.md
dr-jts: Yes, good idea.  I'll make this change.

dr-jts:
@dbaston thanks for the review!  Your comments are excellent and much appreciated.

FAQ-LICENSING.md
dbaston: Yes, I think it could be clarified by again replacing "the choice of using using either..." with "the choice of using JTS under the terms of either... "

Agree that the FAQ doesn't need to mention GEOS a second time here.

dr-jts:
Ok, I'm done with updates based on previous review.  Further comments welcome.

dr-jts:
Is it possible to provide some explicit wording that it is ok for LGPL and GPL projects to link to JTS?

@jodygarnett @jnh5y

jodygarnett:
I think we could provide an example of an LGPL project depending on JTS; and a GPL project linking to JTS.

> JTS has adopted a dual license strategy to allow our technology to by used by the widest audience possible, while remaining true to our open source principles. Consider the following examples:
>
> JTS can be used by GPL projects:
> - The Eclipse Distribution License (EDL) follows the Berkley Standard Distribution 3 Clause template which is compatible with the GNU Public License.
> - The GeoServer project, distributed a GPL project, makes use of JTS using the Eclipse Distribution License.
>
> JTS may be used by LGPL projects:
> - The Eclipse Public License (EPL) is compatible with Library Gnu Public License (LGPL).
> - The GeoTools project makes use of JTS, using either the Eclipse Public License (EPL) or Eclipse Distribution License.

dr-jts:
Good idea.

Just one niggly point.  Are those project *actually* using JTS 1.15 yet?  I guess if we know there are no issues with doing so it's safe to say this.  Or maybe have some weasel words like "will use..."

jodygarnett:
We should also be able to adopt the new EPL license with an explicit clause providing compatibility with GPL.

> The Eclipse Public License has been updated to EPL 2.0, which includes several improvements and the option for GPL compatibility. JTS has adopted EPL 2.0 specifically providing compatibility with GNU Public License projects.
> - https://www.eclipse.org/legal/epl-2.0/
> - Previously GPL projects were requried to use the EDL license to avoid conflict.

Let's ask on the email list and see how we can do this with the JTS committers.

FAQ-LICENSING.md
```diff
@@ -0,0 +1,53 @@
...
+  This allows a wider variety of uses.  For instance, it permits iOS use of JTS for mobile development.
+  
+* The **EPL** provides a similar permissiveness to LGPL, with different wording around patents
+
+**A3. But I thought LGPL and EPL do not mix? How does this effect derivative projects like the LGPL GEOS project?**
+
+  An LGPL project (like GEOS) can use JTS under the terms of the EDL license, which is compatible with the LGPL license.
+
+  In the longer term, the EPL license is being revised to work with GPL / LGPL.  
+  This will allow derivative projects the choice of using JTS under the terms of either the EDL or EPL license.
```
strk: I think a mention should be given here about the choice of `distributing` (rather than `using`) derivative projects under the terms of the `GPL` or `LGPL`. Ie: can a derivative project be distributed under the {GPL2,LGPL2,GPL3,LGPL3) license ?

jodygarnett: Understood, @strk can you provide an example text with your comment (trying to gather examples for discussion).

strk:
Something like this (but should be verified by others, as I'm not
a lawyer):

> Distribution of derivative projects under the terms of
> the LGPL or GPL licenses (versions 2 or 3) is permitted
> by the JTS license.

FAQ-LICENSING.md
jodygarnett: reading your note:

>  Distribution of derivative projects under the terms of  the LGPL or GPL licenses (versions 2 or 3) is permitted  by the JTS license.

To go back to the above comment you are asking to revise "is compatible" and make a longer sentence.  My understanding is there is not much difference between distributing an unmodified jar vs porting to another language - in both cases JTS is a "dependency" being used via BSD license.

Here is revised txt:

```
  An LGPL project (like GEOS or GeoTools) can use JTS as a dependency, under the terms of the EDL license. The three clauses (common to all BSD-style licenses) allows GEOS to port the work to C and distribute the resulting work. This is possible because the LGPL license used by GEOS contains additional restrictions (notably a requirement to share modifications to GEOS with end-users) that exceed what is required by JTS, but do not conflict with what is required by JTS.

For a greater discussion on open source licenses and the compatibility between BSD and LGPL expressed above we encourage you to read the [FSF](https://www.gnu.org/licenses/license-list.en.html) commentary. 
```

I have removed the point about EPL because we are not taking advantage optional clause at present.

PR #198, opened 2017-10-21
Deprecate clone in favor of copy method

bjornharrtell:
The built in `Object.clone` is not suitable to rely on for portable implementation. With a original method for copying there is also the added benefit of getting the correct type instead of `Object`.


bjornharrtell:
Rebased with some additional obsolete casts removed.

dr-jts:
Seems like a good idea.

PR #200, opened 2017-10-22
Move operation methods implementation into operation classes

bjornharrtell:
This is the first step in decoupling Geometry from Operations.

In JSTS transpilation this has been a patch on a fork to allow Geometry operation methods to be mitted because the coupling makes for some difficult circular dependencies.


bjornharrtell:
I would also suggest to move the documentation to the Operation classes, and use @see tags in the delegating methods on Geometry. Let me know if that is desired and I'll update this PR when possible.

dr-jts:
I think it's still good to have full Javadoc on the Geometry class, to provide a single point of reference for the semantics of the OGC API.

bjornharrtell:
Ok, I will add see tags in the introduced methods instead then.

bjornharrtell:
This is now ready @dr-jts

bjornharrtell:
Rebased

dr-jts:
I see a couple of issues with some of the changes in this PR:

* There is now a reference loop between OverlayOp and SnapIfNeededOverlayOp, which is ugly and confusing.  Probably need a new class carrying just the static methods
* RelateOp was really intended to just have the code for the full computation of the relate matrix.  So might be clearer if the code with the various short cuts was moved to a new class.

The other changes are fine.  How about separating them into a different PR, so they can be merged?

bjornharrtell:
I agree, especially with point 1. But I'm not sure what to do... inventing new names for these cases will be inconsistent with the current naming, where many similar static short cuts on Op-classes already.

bjornharrtell:
The coupling between `SnapIfNeededOverlayOp` and `OverlayOp` seems solvable, but hard without breaking API. I'd suggest introducing `OverlayUtil` or `OverlayHelpers` for the short cuts, leaning `Util` as that naming has some existing use in the library.

bjornharrtell:
With some additional thought on this I think I'd prefer this PR as is, even though the ugliness, as I think doing anything more clean would need more API breakage than anyone wants to see until next major version.

dr-jts:
I will give some thought to naming of new classes.  Not sure I like using `Util `- a bit vague.

bjornharrtell:
This is now rebased on master and is ready for another review round. My comment in https://github.com/locationtech/jts/pull/200#issuecomment-340284721 still holds.

bjornharrtell:
@dr-jts this is now rebased against master and I've checked that the moved code has no changes from when I did it and what's in master now.

modules/core/src/main/java/org/locationtech/jts/triangulate/VoronoiDiagramBuilder.java
```diff
@@ -171,7 +172,7 @@ private static Geometry clipGeometryCollection(Geometry geom, Envelope clipEnv)
 			if (clipEnv.contains(g.getEnvelopeInternal()))
 					result = g;
 			else if (clipEnv.intersects(g.getEnvelopeInternal())) {
-				result = clipPoly.intersection(g);
+				result = OverlayOp.intersection(clipPoly, g);
```
dr-jts, line 175: Is this needed to avoid a dependency cycle?

Seems unfortunate if so, in that it's odd that library classes can't use it's actual API.

dr-jts:
Other than the slight ugliness of not being able to call the Geometry API from internal classes, I'm good with this change.

I still think there might be a better design than all those static methods on the Op classes, but am not totally sure of the details yet.  Something that is more dynamic (along the lines of a  registry for operations, allowing them to be configured at runtime.

There is likely to be new codebases for both overlay and relate in the near term (now, for overlay).  Integrating those might motivate a more dynamic strategy for calling operations.

modules/core/src/main/java/org/locationtech/jts/triangulate/VoronoiDiagramBuilder.java
bjornharrtell: Yes it's to avoid dependency cycle. Java is lenient on circular dependencies (by making it mandatory that Java compilers to do multi-pass compilation) but that doesn't make it a good idea. That said, I'm also not sure what a good design would be.

dr-jts: > That said, I'm also not sure what a good design would be.

You mean a better design to avoid a dependency cycle?

This seems like the simplest way to avoid a cycle.  But it introduces a different dependency, on an implementation class which might change (e.g. if OverlayNG is used as the new overlay code).  And if a dynamic method registry is used this would have be changed as well.

bjornharrtell: I mean I do not have a clear picture of a better design to decouple geometry and operations that would work well even with transpilation in mind. I don't know how to do "runtime dynamic method registry" without reflection in Java. In this case in particular reflection not good for portability/transpilation, but I also have a general problem with it as it bypasses the security of Java as a static typed language and makes it harder to do safe refactoring.

bjornharrtell:
I just thought about that it was quite a long time ago I was blocked by the circular dependency and the tooling involved with JSTS has improved since. I should try to transpile without this patch and see how that goes.

modules/core/src/main/java/org/locationtech/jts/triangulate/VoronoiDiagramBuilder.java
dr-jts: I'm thinking that geometry operations would be reified (turned into classes with a generic interface for running operations).  The current `Op` class convention was a non-functional step in that direction back before I really had a solid idea about what to do. 

Alternatively, each general operation (overlay, relate, etc) will provide an interface exposing the operations it provides.  A registry will hold instances of these interfaces.

Would one of these work for transpilation?  Although not sure I want to make that big a change right now.

bjornharrtell: I can see where you are going and it might be a sound design, but I can't imagine it at a good enough level to say which direction would be nicest. As for meaningful transpilation of  interfaces it's undiscovered country (for me).

dr-jts:
> Update: Not so easy. Some Java 8 features are now used fx. generics and lambdas. Transpilation tool needs some work...

Er, yes...  I caved in to the (not-so-) new world of Java.  I don't think I'm using lambas though?  And generics only in a way that can be dropped safely (e.g. for type safety only).

bjornharrtell:
> > Update: Not so easy. Some Java 8 features are now used fx. generics and lambdas. Transpilation tool needs some work...
> 
> Er, yes... I caved in to the (not-so-) new world of Java. I don't think I'm using lambas though? And generics only in a way that can be dropped safely (e.g. for type safety only).

I'll see what I can do. :) I think this one here https://github.com/locationtech/jts/blob/511c3d938145cb3601ef3bcda8601ea99018f39f/modules/core/src/main/java/org/locationtech/jts/algorithm/InteriorPointArea.java#L259 is equivalent to a lambda.

dr-jts:
>  I think this one here
> 
> https://github.com/locationtech/jts/blob/511c3d938145cb3601ef3bcda8601ea99018f39f/modules/core/src/main/java/org/locationtech/jts/algorithm/InteriorPointArea.java#L259
> 
> is equivalent to a lambda.

Yes, probably.  That can probably be changed if needed.

bjornharrtell:
I've been able to improve the transpiler to handle generics but the circular dependency still causes problem. Perhaps future transpilation improvements could help (I suspect certain cases of static members is a trigger) however I'm not sure. @dr-jts, if you can live with this PR as is it would definitely help me keep up to date with JTS 1.16 and the forthcoming 1.17.

PR #202, opened 2017-10-22
Remove explicit super call and explicit add method

bjornharrtell:
This should not affect the Java code at all but helps avoiding two difficult transpilation cases.


dr-jts:
I regret making CoordinateList extend List, since it has too wide an interface.  This change seems ok, although I wonder about removing the call to the super constructor.

bjornharrtell:
In Java super is called implicitly with no arguments, so no worries. :)

PR #211, opened 2017-10-29
Implement overrides make overload resolution more explicit

bjornharrtell:
This is to solve similar problems as in https://github.com/locationtech/jts/pull/213 but is less invasive, should not be breaking anything.


dr-jts:
Since these changes introduce overhead which is not required by Java, it would be worth adding a comment to indicate that these are in place to support JSTS.  Otherwise they might get "cleaned up" in a future release (and also alerts readers that there is a reason why standard Java practice is not being followed).

bjornharrtell:
Hmm didn't realize they cause overhead. Will see if I can find a solution on my end instead.

bjornharrtell:
Closing until further notice.

PR #212, opened 2017-10-29
Add specifically named create for MultiPoint from coords

bjornharrtell:
This one is problematic for overload resolution when transpiling to JSTS because there are two variants taking array input with different types, but that information is erased and the arrays might be empty.

Not breaking API for Java as the existing methods are kept. in JSTS createMultiPoint will only work with Point arrays and `createMultiPointFromCoords` is needed for Coordinate arrays.

Also removes one more case of createMultiPoint with null argument.

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/geom/GeometryFactory.java
+++ b/modules/core/src/main/java/org/locationtech/jts/geom/GeometryFactory.java
@@ -384,6 +384,12 @@ public class GeometryFactory
                               : null);
   }
 
+  public MultiPoint createMultiPointFromCoords(Coordinate[] coordinates) {
+      return createMultiPoint(coordinates != null
+                              ? getCoordinateSequenceFactory().create(coordinates)
+                              : null);
+  }
+
   /**
    * Creates a {@link MultiPoint} using the 
    * points in the given {@link CoordinateSequence}.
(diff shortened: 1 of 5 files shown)


dr-jts:
Need Javadoc for new createMultiPointFromCoords method, please.

Should probably deprecate createMultiPoint(Coordinate[]) method, since it's not ideal having two different methods with identical semantics.

bjornharrtell:
Improved docs as requested.

modules/core/src/main/java/org/locationtech/jts/geom/GeometryFactory.java
```diff
@@ -377,13 +377,21 @@ public MultiPoint createMultiPoint(Point[] point) {
    *
    * @param coordinates an array (without null elements), or an empty array, or <code>null</code>
    * @return a MultiPoint object
+   * @deprecated Use {@link GeometryFactory#createMultiPointFromCoords} instead
```
dr-jts: Actually, don't want to deprecate this.  It is useful for programmatic usage, to not need a separate code path just for empty coordinates.

bjornharrtell: But `createMultiPointFromCoords` is exactly the same?

dr-jts: Ah right...  cancel that, was thinking of something else.

PR #213, opened 2017-10-29
Rename coordinate edit method

bjornharrtell:
This is to solve a difficult overload situation that I haven't been able to solve in transpiling. It's a case of relying on inheritance and method overload at the same time.

As this is a breaking change so I understand if it's not acceptable. Wanted to suggest it anyway if it's known that this API is rarely used and could perhaps be changed.


dr-jts:
What about the overloaded edit method in CoordinateSequenceOperation?  

That should probably be renamed for consistency, if nothing else.

bjornharrtell:
Agreed, it looks a bit inconsistent. As mentioned, already a bit hesistant on this change in the first place.. and the rename is required only because the input is an array causing a hard time resolving the overload like in some previous cases. So, I will instead see if I can work out a solution to avoid this change.

bjornharrtell:
Closing until further notice.

PR #215, opened 2017-11-10
Makes BasicPreparedGeometry extend Geometry

atallahhezbor:
Signed-off-by: Atallah Hezbor <atallahhezbor@gmail.com>

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/geom/prep/BasicPreparedGeometry.java
+++ b/modules/core/src/main/java/org/locationtech/jts/geom/prep/BasicPreparedGeometry.java
@@ -15,9 +15,7 @@ import java.util.Iterator;
 import java.util.List;
 
 import org.locationtech.jts.algorithm.PointLocator;
-import org.locationtech.jts.geom.Coordinate;
-import org.locationtech.jts.geom.Geometry;
-import org.locationtech.jts.geom.GeometryCollection;
+import org.locationtech.jts.geom.*;
 import org.locationtech.jts.geom.util.ComponentCoordinateExtracter;
 
 
@@ -31,7 +29,7 @@ import org.locationtech.jts.geom.util.ComponentCoordinateExtracter;
  * @author Martin Davis
  *
  */
-class BasicPreparedGeometry 
+class BasicPreparedGeometry extends Geometry
   implements PreparedGeometry
 {
   private final Geometry baseGeom;
@@ -39,8 +37,33 @@ class BasicPreparedGeometry
 
   public BasicPreparedGeometry(Geometry geom) 
   {
+    super(geom.getFactory());
     baseGeom = geom;
     representativePts = ComponentCoordinateExtracter.getCoordinates(geom);
+  }
+
+  public Geometry copy() {
+    return baseGeom.copy();
+  }
+
+  public void normalize() {
+    baseGeom.normalize();
+  }
+
+  protected Envelope computeEnvelopeInternal() {
+    return baseGeom.getEnvelopeInternal();
+  }
+
+  protected int compareToSameClass(Object o) {
+    return baseGeom.compareTo(o);
+  }
+
+  protected int compareToSameClass(Object o, CoordinateSequenceComparator comp) {
+    return baseGeom.compareTo(o, comp);
+  }
+
+  protected int getSortIndex() {
+    return 0;
   }
 
   public Geometry getGeometry() { return baseGeom; }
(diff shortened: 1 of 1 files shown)


atallahhezbor:
The idea here would be to allow UserDefinedTypes that accept Geometry to accept PreparedGeometry as well.

@jnh5y I can't assign you, but here

dr-jts:
The clean way to do this is to make Geometry an interface, and then have BasicPreparedGeometry implement it.  But that is a pretty severe breaking change, so will have to wait. 

I suppose this at least gets part of the way there (although it still usually required casts, right?)

dr-jts:
Looks to me like more methods should be overridden for full compatibility?  E.g. intersection etc.

atallahhezbor:
In our case we did not need to cast.

And BasicPreparedGeometry still implements the PreparedGeometry interface, so it already implements intersects, within, overlaps, etc. by delegating to the base geometry.

modules/core/src/main/java/org/locationtech/jts/geom/prep/BasicPreparedGeometry.java
```diff
@@ -31,18 +29,43 @@
...
+  public Geometry copy() {
+    return baseGeom.copy();
+  }
+
+  public void normalize() {
+    baseGeom.normalize();
+  }
+
+  protected Envelope computeEnvelopeInternal() {
+    return baseGeom.getEnvelopeInternal();
```
elahrvivaz: should this be `baseGeom.computeEnvelopeInternal` ?

modules/core/src/main/java/org/locationtech/jts/geom/prep/BasicPreparedGeometry.java
```diff
@@ -31,18 +29,43 @@
...
+  protected int compareToSameClass(Object o) {
+    return baseGeom.compareTo(o);
+  }
+
+  protected int compareToSameClass(Object o, CoordinateSequenceComparator comp) {
+    return baseGeom.compareTo(o, comp);
+  }
+
+  protected int getSortIndex() {
+    return 0;
```
elahrvivaz: should this delegate to baseGeom?

modules/core/src/main/java/org/locationtech/jts/geom/prep/BasicPreparedGeometry.java
atallahhezbor: Should but. But it that method is protected

modules/core/src/main/java/org/locationtech/jts/geom/prep/BasicPreparedGeometry.java
atallahhezbor: Same as above. It's protected in Geometry

elahrvivaz: wonder if it will cause any issues?

elahrvivaz: i'd at least add a comment/fixme type message

atallahhezbor: It might also work to have subclasses of BasicPreparedGeometry override it with their own sort index.

elahrvivaz: yeah, i don't know enough about it to say one way or another

modules/core/src/main/java/org/locationtech/jts/geom/prep/BasicPreparedGeometry.java
dr-jts: But can't a subclass override a protected method in the parent class?

modules/core/src/main/java/org/locationtech/jts/geom/prep/BasicPreparedGeometry.java
```diff
@@ -31,18 +29,54 @@
...
  *
  */
-class BasicPreparedGeometry 
+class BasicPreparedGeometry extends Geometry
   implements PreparedGeometry
 {
   private final Geometry baseGeom;
   private final List representativePts;  // List<Coordinate>
 
+  static final int SORTINDEX_POINT = 0;
```
elahrvivaz: can you add a comment on where you got these values from?

modules/core/src/main/java/org/locationtech/jts/geom/prep/BasicPreparedGeometry.java
```diff
@@ -31,18 +29,55 @@
...
  *
  */
-class BasicPreparedGeometry 
+class BasicPreparedGeometry extends Geometry
   implements PreparedGeometry
 {
   private final Geometry baseGeom;
   private final List representativePts;  // List<Coordinate>
 
+  // Aligned with sort indexes in Geometry
```
dr-jts: It is too brittle to replicate these constants.

modules/core/src/main/java/org/locationtech/jts/geom/prep/PreparedPoint.java
```diff
@@ -45,4 +45,7 @@ public boolean intersects(Geometry g)
   	 */
     return isAnyTargetComponentInTest(g);
   }  
+  protected int getSortIndex() {
+      return BasicPreparedGeometry.SORTINDEX_POINT;
```
dr-jts: Can't this just delegate to the parent geometry `getSortIndex `?  Then the explicit reference to the constant is not needed.

elahrvivaz: the problem is that `BasicPreparedGeometry` is trying to delegate to it's wrapped geometry, but it can't access the protected methods in the wrapped geometry in order to invoke them... without having `PreparedGeometry` extend `Geometry`, I think the other option would be to use reflection to access the protected method...

dr-jts: Right, understood.    It's a fundamental problem with this approach.  I'm not keen on injecting this into the core JTS codebase.  If you really need this can't you create a wrapper object external to JTS?

jnh5y: @dr-jts, as an FYI, I'm the one pushing for this.:)  

I'm seeing the PreparedGeometry as a good story for working with optimizations from JTS through GeoTools/GeoServer to end applications.  In GeoTools, it seems like there should be a 'prepareFilter' method/visitor would optimize spatial filters involving complex geometries.

Issue #205, opened 2017-10-28
Move RobustDeterminant back from test to core


jnh5y:
GeoTools was using RobustDeterminant.

jnh5y:
@dr-jts @jodygarnett if RobustDeterminant shouldn't be part of JTS, perhaps we could deprecate it and remove it as part of 1.16 or 2.0?  In that event, GeoTools could roll the code into its codebase?

dr-jts:
Fixed by #216.  

@jnh5y can you confirm this and close issue if so?

dr-jts:
@jnh5y  should probably also delete branch `revertRobust` ?

PR #216, opened 2017-11-10
Restore RobustDeterminant to core/main

dr-jts:
Until clients can remove usage.  At that point can be moved back to core/test fork


dr-jts:
Fixes #205

PR #219, opened 2017-11-13
STRtree explicit internal query

bjornharrtell:


bjornharrtell:
Rebased on master, no longer dependent on other PRs.

PR #220, opened 2017-11-13
Rename private methods to avoid complex overload resolution

bjornharrtell:


modules/core/src/main/java/org/locationtech/jts/algorithm/PointLocator.java
```diff
@@ -156,7 +156,7 @@ private int locate(Coordinate p, Point pt)
     return Location.EXTERIOR;
   }
 
-  private int locate(Coordinate p, LineString l)
+  private int locateOnLineString(Coordinate p, LineString l)
```
jodygarnett: So this is a private method, should not have any lasting API change

modules/core/src/main/java/org/locationtech/jts/algorithm/PointLocator.java
```diff
@@ -146,7 +146,7 @@ private void updateLocationInfo(int loc)
     if (loc == Location.BOUNDARY) numBoundaries++;
   }
 
-  private int locate(Coordinate p, Point pt)
+  private int locateOnPoint(Coordinate p, Point pt)
```
jodygarnett: So this is a private method, should not have any lasting API change.

PR #221, opened 2017-11-13
Use alternative method to change PM

bjornharrtell:


modules/core/src/main/java/org/locationtech/jts/precision/GeometryPrecisionReducer.java
```diff
@@ -173,8 +173,8 @@ private Geometry fixPolygonalTopology(Geometry geom)
   	
   	Geometry finalGeom = bufGeom;
   	if (! changePrecisionModel) {
-  	  // a slick way to copy the geometry with the original precision factory
-  		finalGeom = geom.getFactory().createGeometry(bufGeom);
+  		finalGeom = bufGeom.copy();
```
dr-jts: Don't think this copy is needed - the GeometryEditor will copy.

bjornharrtell: Hmm ok, I thought GeometryEditor with NoOpGeometryOperation did not.

bjornharrtell: Regardless, I see now I get two failing tests so will have to rethink this.

dr-jts: yes, I think you are right - using NoOpGeometryOperation does not copy the coordinates (but does copy the Geometry objects using the new factory holding the new PM).

Too bad - seems like extra work to copy the entire geometry just to copy the coordinates.
Sigh... GeometryEditor is pretty complicated, and easy to get wrong.

bjornharrtell: Indeed. Would it make sense to have a setter for PrecisionModel on the Geometry to use instead?

bjornharrtell: Or rather, set the PrecisionModel on the GeometryFactory.. but I tried that and the failures are still there, and reading you right this is what the NoOpGeometryOperation does too. So I'm confused about what the original code actually does now, it seems it does more than change the PrecisionModel.

dr-jts: GeometryFactory must be immutable, so cannot have setter for PM.  Can just make a new factory if needed.  This is what the original code did. 

What is the problem with the original code?

bjornharrtell: I get into problems with the transpilation of the anonymous GeometryEditor.CoordinateSequenceOperation, here https://github.com/locationtech/jts/blob/6b5829735c7b03b283f910cd88257da950d8f8d9/modules/core/src/main/java/org/locationtech/jts/geom/GeometryFactory.java#L607. The overload resolution comes out wrong for unknown reasons as of yet.

Other than that, I fail to understand what the original code does more than the comment `a slick way to copy the geometry with the original precision factory` appear to indicate.

bjornharrtell: (if it did just copy the geometry using the original factory, I don't see why the alternative code would not work)

dr-jts: The original code pretty much does what the comment says. But note that the createGeometry does not copy the coordinate objects, but this is ok in this context because they have already been copied by the buffer(0) operation.

dr-jts:
Trying to pick this up again, to see if there's a way to resolve this.

My current thinking is:
* `GeometryEditor.NoOpGeometryOperation` is actually broken, since it doesn't copy the new geometry factory all the way down the geometry (it does not alter atomic geometries e.g. `Linestrings` and `Points`).
* However, it does work well enough to send a new `PrecisionModel` into the `buffer(0)` operation, which is all that is needed here
* But this means that a real transformed copy of the geometry is necessary.  `GeometryFactory.createGeometry()` does this.
* So, either `GeometryFactory.createGeometry()` needs to be made portable somehow, or some other method of copying is required.

If the problem is the anonymous inner class, is there some way to change this?  Would a named inner class work?  Or a new static class (preferably in the same file) ?

PR #225, opened 2017-11-15
Upgrade JDom to version 2.0.6

mprins:
JTS is using a version of JDom that has long been EOL, having been released in Feb 2012.

This upgrades to version 2.0.6, as part of this upgrade I replaced LineNumberSAXBuilder with the regular jdom SAXBuilder configured with a LocatedJDOMFactory. This is neccesary because LineNumberSAXBuilder overrides the element methods that are now final in the class being extend.

see also:
- https://github.com/hunterhacker/jdom/wiki/JDOM2-Migration-Issues
- https://github.com/hunterhacker/jdom/wiki/JDOM2-Dependencies-and-Supported-Versions


jnh5y:
@mprins thanks for the PR.  One advantage to duplicating the JDOM code into JTS is that we avoid JTS having a dependency.  

The code is used only in the JTS test infrastructure, so that also limits some exposure.  

Is there a deficiency in the JDOM code as-is in the JTS?  (I.e. a security, correctness, or performance issue?)

jnh5y:
@mprins Mea culpea.  Jody and I went back and forth on the JDOM thing, and I was mistaken a second ago.  The class included extends base JDOM classes, and otherwise lives in a jdom-contrib jar which Eclipse was unsure about during their IP review.

After a second read and a check with Eclipse's IP database, it looks like using JDOM 2.0.6 is gonna be an improvement for us.  There are two details to note:  1.  I'll need to go through a quick approval process before we can merge, and 2. there's some documentation/license bits to clean up after/as we merge this PR.

For GeoTools/GeoServer, the recent update was to JTS 1.14.0 which has published to Maven central.  Between 1.13/1.14 and 1.15.0, the package names have changed.  To check out the work required for GT/GS to transition over, I have started branches for those projects and their dependencies here: https://github.com/jnh5y?tab=repositories

In terms of the concerns about jdom as a dependency, I ran a mvn dependency:tree for the head of JTS, and JDOM is only a dependency of jts-tests and jts-app.  From looking at the dependency trees of my current jts1.15 branches for GeoTools and GeoServer, those modules are not present.  

Basically, I *think* that JDOM won't be pulled into GeoTools/GeoServer from JTS 1.15.

Generally, I'd love some help running through some more details on the GeoTools and GeoServer builds.  I have setup CI builds on LocationTech infrastructure for Jeo, Jaitools, Jai-ext, GeoDB, GeoTools, GWC, and GeoServer...  My branches aren't quite done yet though...

mprins:
I have already got both geotools and geoserver to switch over to jdom 2.0.6, see geotools#geotools#1719 and geoserver/geoserver#2636
Since there were some api changes this war only merged to master and as such will only appear in the 19.x and up versions

mprins:
I checked that the jdom license has not changed, it's still the same as in 2000: https://github.com/hunterhacker/jdom/blob/master/LICENSE.txt

jnh5y:
Tossed in a CQ for this.  Once that is approved, we'll go from there.

jnh5y:
This has been approved.

@jodygarnett @dr-jts I wonder if we should merge this before or after the 1.15.0 release.  Thoughts?

dr-jts:
I'm fine with merging this into 1.15.

PR #226, opened 2017-11-15
Upgrade commons-lang3 (3.5 -> 3.7)

mprins:
Note that newer than 3.5 versions require java 7

see also: http://commons.apache.org/proper/commons-lang/release-notes/RELEASE-NOTES-3.7.txt

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -111,7 +111,7 @@
         <dependency>
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-lang3</artifactId>
-            <version>3.5</version>
+            <version>3.7</version>
         </dependency>
         <dependency>
           <groupId>com.oracle</groupId>


jnh5y:
@mprins the use for commons-lang3 is merely to handle some HTTP utils.  Per my comment on the JDOM dependency PR, I'd suggest that we go to the small amount of work to copy over the one function required from commons-lang3 so that we can ditch this dependency:).

As a note, any change in the dependencies used by JTS (either via Maven/pom or code re-use) does require some quick paperwork on the Eclipse/LocationTech side of things.  Once we decide on the best solution, I can help with those details.  

(The process does require us to get permission for changes, so there'll be a delay before we can merge a PR changing things.)

jnh5y:
I entered a CQ for this update.  Probably be able to close the loop on this in a few days.

jnh5y:
Good news; the IP review is complete for commons lang 3.7.

PR #228, opened 2017-11-21
Updated Licensing information after the JDOM upgrade.

jnh5y:
Signed-off-by: Jim Hughes <jnh5y@ccri.com>


lossyrob:
JTS removed code that was copied in from JDOM, and now it's just a dependency.

Looks good to me!

PR #232, opened 2017-11-21
Updates to allow for publishing to Maven central.

jnh5y:
* Creates 'release' profile and ties it to a 'release' property.

To release, call 'mvn clean install -Drelease'.

Normal, non-release builds will build the app, tests, and example modules.

### DIFF
--- a/modules/pom.xml
+++ b/modules/pom.xml
@@ -20,10 +20,24 @@
 
     <modules>
         <module>core</module>
-        <module>example</module>
         <module>io</module>
-        <module>tests</module>
-        <module>app</module>
     </modules>
-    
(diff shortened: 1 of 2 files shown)


modules/pom.xml
```diff
@@ -20,10 +20,24 @@
...
-        <module>app</module>
     </modules>
-    
+
+    <profiles>
+        <profile>
+            <id>all</id>
+            <activation>
+                <property>
+                    <name>!release</name>
```
jodygarnett: Looks good, I usually document profiles and switches like this in comments in the root pom.xml (or at the build instructions)

PR #257, opened 2018-02-03
Improve Overlay Noding validation

dr-jts:
This code extends the range of checks performed by the noding validation used in Overlay.  

The fix addresses issue #107.

This change improves the overall robustness of JTS overlay, in that it prevents some cases which otherwise fail "silently" (e.g. by producing invalid output).  However, it may actually cause a few more overlay topology exceptions, since it may flag cases which fail snapping and which previously might have made it through the overlay algorithm correctly.


dr-jts:
@strk feel free to review before I accept this.

dr-jts:
This code should perhaps have one or more unit tests which run in the core build.

strk:
I belive I tested this against GEOS testsuite and it didn't trigger any new failure there. It's great to see more unit tests :)

PR #259, opened 2018-02-06
Use internal priority queue

bjornharrtell:
See https://github.com/bjornharrtell/jsts/issues/330#issuecomment-363226574.

Need some verification, I just made it compile. :)


dr-jts:
Fixed by #391

Issue #260, opened 2018-02-21
DD (DoubleDouble) parsing issue with whole numbers


FObermaier:
Parsing negative whole numbers with DD.Parse yields wrong results.

```Java
assertEquals(DD.valueOf(1), DD.parse("1")); // DD.parse yields 0.1
assertEquals(DD.valueOf(100), DD.parse("100")); // DD.parse yields 0.1
assertEquals(DD.valueOf(-1), DD.parse("-1")); // DD.parse yields -0.1
assertEquals(DD.valueOf(-100), DD.parse("-100")); // DD.parse yields -0.1
assertEquals(DD.valueOf(-1), DD.parse("-123")); // DD.parse yields -0.123
```

Please see attached patchfile for extended DDIOTest unit test.
[Added_test_for_DD_parse_of_whole_numbers.zip](https://github.com/locationtech/jts/files/1743622/Added_test_for_DD_parse_of_whole_numbers.zip)

FObermaier:
The would be the fix

```Java
    // somewhere around line 1358 
    DD val2 = val;

    // correct number of digits before decimal sign if we have a whole negative number
    if (numBeforeDec == 0 && isNegative) numBeforeDec = 1;

    // scale the number correctly
    int numDecPlaces = numDigits - numBeforeDec - exp;

```

FObermaier:
This would be the fix:

```Java
    DD val2 = val;

    // correct number of digits before decimal sign if we have a whole negative number
    if (numBeforeDec == 0 && isNegative) numBeforeDec = numDigits;

    // scale the number correctly
    int numDecPlaces = numDigits - numBeforeDec - exp;
```

dr-jts:
Fixed by #260

PR #261, opened 2018-02-21
Fix DD.parse for whole numbers (#260)

FObermaier:
Signed-off-by: Felix Obemaier <felix.obermaier@netcologne.de>

### DIFF
--- a/modules/core/src/test/java/org/locationtech/jts/math/DDIOTest.java
+++ b/modules/core/src/test/java/org/locationtech/jts/math/DDIOTest.java
@@ -124,7 +124,8 @@ public class DDIOTest extends TestCase {
 			double relErrBound) {
 		DD xdd = DD.parse(str);
 		double err = xdd.subtract(expectedVal).doubleValue();
-		double relErr = err == 0d ? 0d : Math.abs(err / xdd.doubleValue());
+		double xddd = xdd.doubleValue();
+		double relErr = xddd == 0d ? err : Math.abs(err / xddd);
 
 		//System.out.println("Parsed= " + xdd + " rel err= " + relErr);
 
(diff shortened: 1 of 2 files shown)


modules/core/src/test/java/org/locationtech/jts/math/DDIOTest.java
```diff
@@ -88,6 +88,11 @@ private void checkSciNotation(DD x, String expectedStr) {
 	}
 
 	public void testParse() {
+		//checkParse("0", 0, 1e-32);
```
dr-jts: Any reason this is commented?

modules/core/src/test/java/org/locationtech/jts/math/DDIOTest.java
```diff
@@ -88,6 +88,11 @@ private void checkSciNotation(DD x, String expectedStr) {
...
+		checkParse("1", 1, 1e-32);
+        checkParse("100", 100, 1e-32);
```
dr-jts: Would be nice to keep indentation consistent

dr-jts:
Good catch - thanks for the patch.

modules/core/src/test/java/org/locationtech/jts/math/DDIOTest.java
FObermaier: Problem calculating relerr (div/0), fixed that.

modules/core/src/test/java/org/locationtech/jts/math/DDIOTest.java
FObermaier: Agreed

dr-jts:
@FObermaier Thanks for the last commit. Can you amend it to have a sign-off?  Then I will merge.

jnh5y:
@dr-jts we can use the 'Squash and merge' and then edit the commit message.  I can do that if you like.

dr-jts:
@jnh5y Cool!  Please do...

jnh5y:
@dr-jts @FObermaier thanks, and good job!

Issue #265, opened 2018-02-23
Possible regression in "Avoid using reflection" commit


FObermaier:
Please see ef92b816438c416303df2eecbc0f655db6e21466.

PR is #155 

Prior to this commit `protected void Geometry.checkNotGeometryCollection(Geometry g)` did the check on the argument `g`. The new version of the function ignores the argument and checks on the instance.

By looking at the way this function is used, it seems to me that the argument does matter:
* [Geometry.relate](https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/geom/Geometry.java#L1042)
* [Geometry.intersection](https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/geom/Geometry.java#L1351)
* [Geometry.union](https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/geom/Geometry.java#L1404)
* [Geometry.difference](https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/geom/Geometry.java#L1432)
* [Geometry.symDifference](https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/geom/Geometry.java#L1468).

Design notice: it seems to me that this method may be declared `static`.

FObermaier:
As an experiment I commented out the `Geometry.checkNotGeometryCollection`'s function body and ran the Junit unit tests. The only place where one of them failed was [here](https://github.com/locationtech/jts/blob/master/modules/core/src/test/java/test/jts/junit/MiscellaneousTest.java#L122) on the `GeometryCollection`.

So it may be an optionto remove the function and its usage altogether and explicitly throw an exception in GeometryCollection.getBoundary() overload. `IllegalArgumentException` might not be first choice, though.

dr-jts:
I agree with your analysis.  Thanks for noticing this.

The reason almost no unit tests fail is because they were designed to test supported functionality, not checks for disallowed inputs. That said, the NotGC check forms part of the JTS API semantics, so it should be tested by unit tests.  GCs are not supported for spatial predicates and overlay operations, and in the absence of this check incorrect results will be returned.  So it is needed.

And yes, this method could be static.

@bjornharrtell what do you think?  Is this just a slip in the refactoring?

dr-jts:
[...]

PR #266, opened 2018-02-26
Fix Geometry operation check for unsupported GeometryCollections

dr-jts:
Fix Geometry operation check for unsupported GeometryCollections
Add unit test


dr-jts:
Fix for #265

PR #268, opened 2018-03-03
Fixes for PackedCoordinateSequenceFactory.create methods

FObermaier:
The `GeometryFactoryTest.testCopyWithNonDefaultDimensions` fails when using `PackedCoordinateSequenceFactory.DOUBLE_FACTORY` as the default geometry factory.
This is because the `create(CoordinateSequence)` function uses `CoordinateSequence.toCoordinateArray` to create the new sequence.

This PR includes:
* Add constructor with a `CoordinteSequence` argument to `PackedCoordinateSequence.Double` and `PackedCoordinateSequence.Float`
* Added missing or completed code documentation to `PackedCoordinateSequence`.
* Fixed GeometryImplTest.testInvalidateEnvelope to use `CoordinateSequenceFilter` as the `CoordinateFilter` did not work with PackedCoordinateSequence.
* Fixed initialisation of `create(size, dimension)` to initialize dimensions >`CoordinateSequence.Z` to set these values to `java.lang.Double.NaN`.

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/geom/GeometryFactory.java
+++ b/modules/core/src/main/java/org/locationtech/jts/geom/GeometryFactory.java
@@ -18,6 +18,7 @@ import java.util.Collection;
 import java.util.Iterator;
 
 import org.locationtech.jts.geom.impl.CoordinateArraySequenceFactory;
+import org.locationtech.jts.geom.impl.PackedCoordinateSequenceFactory;
 import org.locationtech.jts.geom.util.GeometryEditor;
 import org.locationtech.jts.util.Assert;
 
--- a/modules/core/src/test/java/org/locationtech/jts/geom/GeometryImplTest.java
+++ b/modules/core/src/test/java/org/locationtech/jts/geom/GeometryImplTest.java
@@ -150,14 +150,18 @@ public class GeometryImplTest extends TestCase {
     public void testInvalidateEnvelope() throws Exception {
         Geometry g = reader.read("POLYGON ((0 0, 0 50, 50 50, 50 0, 0 0))");
         assertEquals(new Envelope(0, 50, 0, 50), g.getEnvelopeInternal());
-        g.apply(new CoordinateFilter() {
-                public void filter(Coordinate coord) {
-                    coord.x += 1;
-                    coord.y += 1;
+        g.apply(new CoordinateSequenceFilter() {
+                public void filter(CoordinateSequence seq, int i) {
+                    seq.setOrdinate(i, CoordinateSequence.X, seq.getOrdinate(i, CoordinateSequence.X) + 1);
+                    seq.setOrdinate(i, CoordinateSequence.Y, seq.getOrdinate(i, CoordinateSequence.Y) + 1);
                 }
+
+                public boolean isDone() { return false; }
+
+                public boolean isGeometryChanged() { return true;}
             });
-        assertEquals(new Envelope(0, 50, 0, 50), g.getEnvelopeInternal());
-        g.geometryChanged();
+        //assertEquals(new Envelope(0, 50, 0, 50), g.getEnvelopeInternal());
+        //g.geometryChanged();
         assertEquals(new Envelope(1, 51, 1, 51), g.getEnvelopeInternal());
     }
 
--- a/modules/core/src/main/java/org/locationtech/jts/geom/impl/PackedCoordinateSequence.java
+++ b/modules/core/src/main/java/org/locationtech/jts/geom/impl/PackedCoordinateSequence.java
@@ -18,6 +18,7 @@ import java.lang.ref.SoftReference;
 import java.util.Arrays;
 
 import org.locationtech.jts.geom.Coordinate;
+import org.locationtech.jts.geom.CoordinateArrays;
 import org.locationtech.jts.geom.CoordinateSequence;
 import org.locationtech.jts.geom.CoordinateSequences;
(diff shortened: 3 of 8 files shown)


modules/core/src/main/java/org/locationtech/jts/geom/GeometryFactory.java
```diff
@@ -100,7 +101,8 @@ public GeometryFactory() {
 
   private static CoordinateSequenceFactory getDefaultCoordinateSequenceFactory()
   {
-    return CoordinateArraySequenceFactory.instance();
```
dr-jts: This is a MAJOR change.  It is not desirable IMO.  

Instead, the `PackedCoordinateSequence `implementation should be fixed, but not used as the default.  It can be tested directly.

dr-jts:
@FObermaier Changing the default CoordinateSequence implementation is a major change.  Was this your intent, or were you just trying to test the fixes to PackedCoordinateSequence ?

FObermaier:
This was just to prove my point and to show that the fix is correct.

modules/core/src/main/java/org/locationtech/jts/geom/GeometryFactory.java
FObermaier: You are quick assigning WON'T FIX.
The fact, that the PackedCoordinateSequenceFactory uses toCoordinateArray() to create a CoordinateSequence off another is a major bug.

WIthout fixing that shifting WKT/WKBReader -Writer is next to useless.

dr-jts:
Ok, that's good to know.  However, it makes impossible to merge this PR as it stands.  

It should be quite possible to fix the issues and test the changes without changing the GeometryFactory default `CoordinateSequence`.  And agreed, it would be nice to be able to run all unit tests with different `CoordinateSequence `implementations.  Not sure of the best way to do this, or whether it can be done automatically, or will require multiple runs with some sort of JVM parameter to switch the default CS.  

It might be more feasible to run the XML tests with different CS implementations - and they should provide pretty good coverage (and if there are gaps they can be fixed).

modules/core/src/main/java/org/locationtech/jts/geom/GeometryFactory.java
dr-jts: Yes, agreed, this is a major bug and needs to be fixed.  The WON'T FIX was just to indicate current status - was not meant to be a permanent thumbs-down!

But it's awkward to avoid merging the GeometryFactory change (and if I hadn't noticed that it would have been a problem!).  If you revert that we can discuss how to test your fix without such a major change.

modules/core/src/test/java/org/locationtech/jts/geom/GeometryImplTest.java
```diff
@@ -150,14 +150,18 @@ private void doTestEquals(Geometry a, Geometry b, boolean equalsGeometry,
     public void testInvalidateEnvelope() throws Exception {
         Geometry g = reader.read("POLYGON ((0 0, 0 50, 50 50, 50 0, 0 0))");
         assertEquals(new Envelope(0, 50, 0, 50), g.getEnvelopeInternal());
-        g.apply(new CoordinateFilter() {
-                public void filter(Coordinate coord) {
-                    coord.x += 1;
-                    coord.y += 1;
+        g.apply(new CoordinateSequenceFilter() {
```
dr-jts: I think this change can be reverted as well.  The original code is is fine for testing the default implementation.  While it might be nice to test with a different CS representation, this should be done somewhere else.

FObermaier: The [documentation to the CoordinateFilter](https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/geom/CoordinateFilter.java#L25) discurages using it for modifying sequences.

dr-jts: Good point!  Ok, let it stand...

dr-jts:
@FObermaier  Did you really mean to add those extraneous merge commits to this PR?  It would be best to remove them, to keep the PR focussed.

FObermaier:
@dr-jts sometimes __git__ is beyond my comprehension. Don't know how this happened and how I revert that.

dr-jts:
Yeah, it's tricky.  You should be able to back out those commits from your repo - perhaps using git revert?  Perhaps @jnh5y  or @lossyrob can help?

FObermaier:
Did get it fixed. `git revert ...` and `git push --force` did the trick.

jodygarnett:
I hope #293 can take over from this branch?

FObermaier:
Would have been easier the other way around. But go ahead, I'm on vacation

dr-jts:
@FObermaier There's a lot of good fixes and Javadoc in this PR, but the codebase has changed quite a bit with #293 and follow-up bug fixes. 

Do you feel like redoing this to match the current code base?

FObermaier:
@dr-jts could you elaborate on the parts you're interested in, besides the code documentation?
I am reluctant to doing it all over for the third time.

dr-jts:
@FObermaier Sure... things that would be nice to have are:

**PackedCoordinateSequence:**
* checkDimension method, with constant names
* use of named constants for ordinate indexes
* error check for invalid array length
* Constructors from CoordinateSequence

And some of the unit tests (recent PR might have added some already)

Issue #188, opened 2017-10-02
Use CoordinateSequence accessors instead of Coordinate arrays


dr-jts:
To improve the performance of using non-array based `CoordinateSequence `implementations the JTS core should be changed to using coordinate access based on `CoordinateSequence ` accessors, instead of forcing the creation of Coordinate arrays.

Some places where this is needed:

* uses of `PointLocation ` `isInRing `and `locateInRing`
* `Centroid.add(Polygon)`

See also:
* https://github.com/NetTopologySuite/NetTopologySuite/pull/193
* https://github.com/NetTopologySuite/NetTopologySuite/pull/71

FObermaier:
All usages of `org.locationtech.jts.geom.CoordinateArrays` class potentially benefit from a utility class that works on `org.locationtech.jts.geom.CoordinateSequence` instead of `org.locationtech.jts.geom.Coordinate[]`

Not explicitly performance, but making the `[WKT|WKB]Reader` classes not work on `org.locationtech.jts.geom.Coordinate[]` can help to improve handling of measure ordinate values. I have an implementation using a `CoordinateBuffer` class at hand.

dr-jts:
Good recommendations @FObermaier .

What is the nature of your `CoordinateBuffer` class?

FObermaier:
> What is the nature of your `CoordinateBuffer` class?

It encapsulates an `ArrayList` to store 3DM coordinate values. You can add ordinate values calling `add2d`, `add3d`, `add2dm`, `add3dm` functions. While adding the following is taken care of:
* Keep track of extent (x, y) and possibly the interval of z- and m- values.
* Check for "No data values" and transform them to and from `Coordinate.NULL_ORDINATE`
* Create a `CoordinateSequence` using any factory
* Get array of ordinate values (`getXY()`, `getZ()`, `getM()`)

You can have a look at it [here](https://github.com/FObermaier/jts/blob/3DM/modules/lab/src/main/java/org/locationtech/jtslab/geom/CoordinateBuffer.java).

PR #271, opened 2018-03-13
ExtendableCoordinateSequence

FObermaier:
This is a wrapper around the `CoordinateSequence` interface.   

Internally it has access to a `CoordinatSequenceFactory` that it uses to create a buffer sequence in which it stores the ordinate values. If the capacity of the sequence is used up, a new sequence with extended capacity is created and the present values are stored in it.

Extensions to the public interface
* several `add` methods
* several `insertAt` methods
* `getCapacity` function
* `truncated` function

This PR comes with a unit test and a performance comparing `ExtendableCoordinateSequence` to creating several 1-coordinate sequences and merging those into one.

This class is convenient wherever the final size of a `CoordinateSequence` is not known, e.g. when parsing WKT (WKTReader) or when building buffers. It might help with issue #188.

I am not sure if the wording is correct, maybe `GrowableCoordinateSequence` or `BufferedCoordinateSequence` is a better name.

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/geom/ExtendableCoordinateSequence.java
+++ b/modules/core/src/main/java/org/locationtech/jts/geom/ExtendableCoordinateSequence.java
@@ -280,7 +280,7 @@ public class ExtendableCoordinateSequence implements CoordinateSequence {
    * @param x a x-ordinate
    * @param y a y-ordinate
    * @param z a z-ordinate
-   * @param m a z-ordinate
+   * @param m a m-ordinate
    */
   public void add(double x, double y, double z, double m) {
     // get the index for the new sequence
@@ -336,7 +336,7 @@ public class ExtendableCoordinateSequence implements CoordinateSequence {
    * @param x a x-ordinate
    * @param y a y-ordinate
    * @param z a z-ordinate
-   * @param m a z-ordinate
+   * @param m a m-ordinate
    */
   public void insertAt(int index, double x, double y, double z, double m) {
 
--- a/modules/core/src/main/java/org/locationtech/jts/geom/ExtendableCoordinateSequence.java
+++ b/modules/core/src/main/java/org/locationtech/jts/geom/ExtendableCoordinateSequence.java
@@ -23,7 +23,7 @@ import org.locationtech.jts.geom.impl.PackedCoordinateSequence;
 public class ExtendableCoordinateSequence implements CoordinateSequence {
 
   /** The default initial capacity */
-  static final int INITIAL_CAPACITY = 12;
+  static final int INITIAL_CAPACITY = 10;
 
   /** The factory to use when extending the current sequence*/
   private final CoordinateSequenceFactory csFactory;
@@ -145,7 +145,7 @@ public class ExtendableCoordinateSequence implements CoordinateSequence {
   public void setOrdinate(int index, int ordinate, double value) {
 
     // ensure capacity
-    ensureCapacity(index);
+    ensureCapacity(index + 1);
 
     // set value in sequence
     this.sequence.setOrdinate(index, ordinate, value);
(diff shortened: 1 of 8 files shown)


modules/core/src/main/java/org/locationtech/jts/geom/ExtendableCoordinateSequence.java
```diff
@@ -0,0 +1,425 @@
...
+import org.locationtech.jts.geom.impl.PackedCoordinateSequence;
+
+import java.io.Serializable;
+
+/**
+ * A wrapper around a coordinate sequence that always
+ * ensures that a sequence is large enough to store ordinate
+ * values.
+ */
+public class ExtendableCoordinateSequence implements CoordinateSequence, Serializable {
```
jodygarnett: This class name kind of conflicts with [ExtendedCoordinateSequence.java](https://github.com/locationtech/jts/blob/master/modules/example/src/main/java/org/locationtech/jtsexample/geom/ExtendedCoordinateSequence.java) which we have noticed is in common use by downstream projects

jodygarnett: Perhaps we could use the word delegate since it is delegating storage to another csFactory?

jodygarnett:
This is really interesting functionality, see comment for question about the naming of this class.

modules/core/src/main/java/org/locationtech/jts/geom/ExtendableCoordinateSequence.java
airbreather: - `DynamicResizingCoordinateSequence`?
- `AutoResizingCoordinateSequence`?
- `ResizingCoordinateSequence`?
- `ExpandableCoordinateSequence`?
- `ExpandingCoordinateSequence`?
- `VariableSizeCoordinateSequence`?

FObermaier:
Is there an interest in persuing this any further?

jnh5y:
@FObermaier I think one of the philosophical questions at play is whether CoordinateSequences should be mutable or immutable.  I know this has come up in conversations with @jodygarnett and @dr-jts, but my memory is faulty and don't remember which position they'd take or if we've reached a conclusion.

I do know that Jody worked with Martin to add the interface so that downstream users of the JTS could do things which make sense.  

That said, I'd like to see the JTS library have some 'collections-like api' support for common operations with CoordinateSequences.  For the case of immutable coordinatesequences, some obvious operations like concatenating and reversing sequences could be implemented.  

I think the next step is Jody and Martin weighing in on mutability.  If that goes, then I think could make sense.

dr-jts:
Generally speaking I think it is less error-prone to have Geometry be immutable.  That said, it seems a bit draconian to enforce this 100%.

And there are other uses for mutable CoordinateSequences. A prime example being when Geometry is being constructed from a stream of incoming data (as occurs for example in some of the Readers).

Perhaps it might have been better to split the read and update accessors for CoordinateSequence into two separate interfaces, so that immutable usage was more obvious (and enforceable).

FObermaier:
As far as I recall, the `CoordinateSequence` interface always allowed modification of its values. 

This PR only addresses that currently adding `Coordinate`s is cumbersome. It comes in handy for streaming or WKT parsing.

dr-jts:
Is this really the most efficient way to support streaming construction of `CoordinateSequence`s when reading?  Would a purpose-built class be simpler or more efficient?

airbreather:
> Is this really the most efficient way to support streaming construction of CoordinateSequences when reading? Would a purpose-built class be simpler or more efficient?

Before I noticed that Felix had started work on this, I had started work on an alternative way to build immutable (well, fixed-length) coordinate sequences using a stream of coordinate data.  My solution looked [like this (C#)](https://github.com/NetTopologySuite/NetTopologySuite/blob/6bea54a02cdf0c64c2e67f6ce744dbc2cf750dce/NetTopologySuite/Utilities/CoordinateBuffer.cs).  I have a few reasons to believe that that solution is going to be more efficient, **in the .NET version**, than creating and expanding coordinate sequences using a factory, but I admittedly haven't run any numbers to compare the two ideas, and I don't know how much of my reasons would be equally valid on the JVM.

I don't really want to volunteer to test the Java side of things, but if it helps, I could try to port this to .NET and compare the performances of the two.

dr-jts:
@airbreather That seems along the lines of what I was thinking about.   

Is it more efficient to create the separate ordinate lists, rather than a list of Coordinate objects?  I guess it would be if they are optimized to store doubles as native values.

dr-jts:
The design that @airbreather referenced seems like a more efficient and safer way of creating an arbitrary `CoordinateSequence` from streamed data.  It avoids any issues with trying to use an arbitrary `CoordinateSequenceFactory` to repeatedly create new sequences.  After all, there is no guarantee that a `CoordinateSequenceFactory` is particularly efficient at creating sequences, or even that they are modifiable.

The only downside might be that the CoordinateBuffer implementation is limited to a max of 4 dimensions.  But this should be adequate to support WKT and WKB reading, which are primary use cases.

airbreather:
> Is it more efficient to create the separate ordinate lists, rather than a list of Coordinate objects?

The main reason I did it this way was so that users don't have to pay for Z or M values that aren't actually used.  Here, if all you need is XY, then you only pay 16 bytes per coordinate (plus the extra room in the array list).

> The only downside might be that the CoordinateBuffer implementation is limited to a max of 4 dimensions.

There's no fundamental reason why it couldn't contain an arbitrary number of dimensions, it just seemed like the extra flexibility wasn't worth the cost when I wrote this up.

FObermaier:
I admit that [`GrowableCoordinateSequence`](https://github.com/geoserver/geoserver/blob/master/src/community/geogig/src/main/java/org/geogig/geoserver/gwc/GrowableCoordinateSequence.java) of GeoServer performs better in most cases, especially when `CoordinateArraySequenceFactory` is being used to get a result geometry 

You can read the figures of a comparison test I created in this file: [ExpandableVsGrowable.txt](https://github.com/locationtech/jts/files/2339811/ExpandableVsGrowabl.txt)
The perfomance comparison measures adding 1500x varying numbers of coordinates to either `GrowableCoordinateSequence`s or `ExtendableCoordinateSequence`s and returning a usable sequence using the observed `CoordinateSequenceFactory`.

Annotations:
* Both sequences are started with an initial capacity of 50.
* The above implementation of `GrowableCoordinateSequence` is limited in that it 
  * only supports 2 dimensions.
  * saves ordinate values as `float`s

dr-jts:
@FObermaier what is your view of the status of this?  

The @airbreather design of [CoordinateBuffer](https://github.com/NetTopologySuite/NetTopologySuite/blob/6bea54a02cdf0c64c2e67f6ce744dbc2cf750dce/NetTopologySuite/Utilities/CoordinateBuffer.cs) seems like a better alternative - do you agree?

Would be nice to close this PR if no longer active.

FObermaier:
IMHO there is no need for a CoordinateBuffer to parse WKT anymore since there are Coordinate classes for all possible ordinates in WKT (`XY[Z|M|ZM]`) defined ordinates. Plain `ArrayList` or `ArrayList<Coordinate>` will suffice.

PR #274, opened 2018-03-23
Add module names for better Java 9/Jigsaw support

lossyrob:
The Jigsaw module system requires that downstream users of JTS using the module system declare JTS as a module requirement. JTS acts as an [automatic module](http://openjdk.java.net/projects/jigsaw/spec/sotms/#automatic-modules), since it is not compiled with a `module-info.java`. Without this change, module users would have to reference the requirement for JTS by specifying the JAR name with the fully qualified version; this is unwieldy, will trigger necessary updates when updating the JTS version, and could cause errors when transitive dependencies rely on other versions of JTS.

This change utilizes the technique of putting the [module name in the manifest](http://openjdk.java.net/projects/jigsaw/spec/issues/#ModuleNameInManifest), so that libraries declaring modules can refer to JTS with more friendly, version-less names, e.g.:

```java
// module-info.java for project org.foo.baz

module org.foo.baz {
  requires org.locationtech.jts;            // jts-core
  requires org.locationtech.jts.io;         // jts-io-common
  requires org.locationtech.jts.io.oracle;  // jts-io-ora
  requires org.locationtech.jts.io.sde;     // jts-io-sde
}
```

## Testing Instructions

To verify this work, I defined a Foo.java file and a manifest-info.java file, as follows:

```java
// Foo.java

package org.foo.baz;

import org.locationtech.jts.JTSVersion;
import org.locationtech.jts.io.ParseException;
import org.locationtech.jts.algorithm.distance.DistanceToPoint;
import org.locationtech.jts.io.geojson.GeoJsonConstants;

public class Foo {
  public static void main(String[] args)
  {
    ParseException e = new ParseException("asdf");
    DistanceToPoint d = new DistanceToPoint();
    System.out.println(JTSVersion.CURRENT_VERSION);
    System.out.println(GeoJsonConstants.NAME_GEOMETRIES);
  }
}

```

```java
// module-info.java

module org.foo.baz {
  requires org.locationtech.jts;
  requires org.locationtech.jts.io;
}
```

I then compiled and ran Foo as a module:

```shell
> mkdir mods
> javac --module-path ./modules/core/target/jts-core-1.15.1-SNAPSHOT.jar:./modules/io/common/target/jts-io-common-1.15.1-SNAPSHOT.jar \
        -d mods/foo \
        module-info.java \
        Foo.java
> java --module-path mods:./modules/core/target/jts-core-1.15.1-SNAPSHOT.jar:./modules/io/common/target/jts-io-common-1.15.1-SNAPSHOT.jar -m org.foo.baz/org.foo.baz.Foo
```

Expected output:
```shell
1.15.0
geometries
```

This is building JTS with java 8 against commit 3fb74b65e142bc166f58c9105f9dc43e911c7d72

### DIFF
--- a/modules/io/ora/pom.xml
+++ b/modules/io/ora/pom.xml
@@ -10,6 +10,30 @@
     <artifactId>jts-io-ora</artifactId>
     <name>${project.groupId}:${project.artifactId}</name>
     <packaging>jar</packaging>
+
+    <build>
+        <plugins>
+            <plugin>
+                <artifactId>maven-jar-plugin</artifactId>
+                <configuration>
+                  <archive>
+                    <manifestEntries>
+                      <Automatic-Module-Name>org.locationtech.jts.io.oracle</Automatic-Module-Name>
+                    </manifestEntries>
+                    <manifestSections>
+                      <manifestSection>
+                        <name>org.locationtech.jts.io.oracle</name>
+                        <manifestEntries>
+                          <Sealed>true</Sealed>
+                        </manifestEntries>
+                      </manifestSection>
+                    </manifestSections>
+                  </archive>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+
     <dependencies>
        <dependency>
           <groupId>com.oracle</groupId>
--- a/USING.md
+++ b/USING.md
@@ -82,3 +82,18 @@ Older versions of JTS are available on Maven Central.
     <version>1.13</version>
 </dependency>
 ```
+
+## Using JTS with Jigsaw
+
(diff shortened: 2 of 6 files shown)


modules/io/ora/pom.xml
```diff
@@ -10,6 +10,30 @@
...
     <packaging>jar</packaging>
+
+    <build>
+        <plugins>
+            <plugin>
+                <artifactId>maven-jar-plugin</artifactId>
+                <configuration>
+                  <archive>
+                    <manifestEntries>
+                      <Automatic-Module-Name>org.locationtech.jts.io.orcale</Automatic-Module-Name>
```
dsmiley: typo

lossyrob: Thanks!

Issue #273, opened 2018-03-20
Geometry.copy() should copy SRID


SPalominos:
With the last (1.15) release, it seems that the abstract method `Geometry.copy()` implementation doesn't copy the Geometry SRID.

As example, the following test on the Polygon class : 
```java
@Test
    public void testSRID(){
        GeometryFactory factory = new GeometryFactory();
        Coordinate[] coords = new Coordinate[]{
                new Coordinate(0, 0),
                new Coordinate(1, 0),
                new Coordinate(1, 1),
                new Coordinate(0, 1),
                new Coordinate(0, 0),
        };
        Polygon polygon = factory.createPolygon(coords);
        polygon.setSRID(4326);
        Polygon copy = polygon.copy();
        Assert.assertEquals(polygon.getSRID(), copy.getSRID());
    }
```
returns an error : `java.lang.AssertionError: Expected :4326 Actual   :0`
Is it a normal behaviour ? Did I miss something and the SRID should be transmitted in an other way (maybe by setting the SRID on the `GeometryFactory`) ?

SPalominos:
As discussed in the issue orbisgis/h2gis#865, it's not clear if it is a normal behaviour or not. Does someone has any clue or link to the documentation where it is explained ?

ebocher:
Thanks @dr-jts 
setUserData must  also be tested.

```java
GeometryFactory factory = new GeometryFactory();       
Point point = factory.createPoint(new Coordinate(0, 0));
point.setUserData(12);
Point copy = (Point) point.copy();
Assert.assertEquals(point.getUserData(), copy.getUserData());
```

dr-jts:
Yes, agreed.  All fields in the Geometry object should be copied, and the unit test should cover these.

My plan is to create a GeometryCopyTest unit test class to include these tests.

dr-jts:
@bjornharrtell any comments?  I assume that adding in the previous semantics will not break anything in JSTS?

bjornharrtell:
Should be fine and the correct semantics, the problem for JSTS was reliance on built in Java object cloning.

dr-jts:
@bjornharrtell do you have any suggestions/constraints for how to implement this?  

At the moment I am thinking of providing an implementation for `Geometry.copy()` to copy the base fields.  Subclasses will have to call `super.copy()`.

Alternatively can leave `Geometry.copy()` as `abstract`, and create a `Geometry.copyBase()`.

ebocher:
[...]

PR #280, opened 2018-05-11
About #273

ebocher:
This PR proposes a fix about the issue #273


dr-jts:
I actually did some work on this in https://github.com/locationtech/jts/compare/master...dr-jts:fix-geometry-copy

I prefer the approach in my patch, since it better encapsulates the copying logic.  Can you review and see if it fixes the problem?  If so I will make a PR

ebocher:
Thanks a lot @dr-jts 
We really appreciate your work.
A real architect's point of view.
Best regards

dr-jts:
Thank you, @ebocher !  

So will you drop this PR in favour of one from my patch?

ebocher:
yep sure.

PR #284, opened 2018-05-23
Fix Geometry.copy() to copy instance fields

dr-jts:


modules/core/src/main/java/org/locationtech/jts/geom/Geometry.java
```diff
@@ -1648,15 +1648,26 @@ public Object clone() {
...
    *
-   * @return a clone of this instance
+   * @return a deep copy of this geometry
    */
-  abstract public Geometry copy();
-
+  public Geometry copy() {
+    Geometry copy = copyInternal();
+    copy.SRID = this.SRID;
+    copy.userData = this.userData;
```
jnh5y: Is there an easy way to copy the userData?

dr-jts: Not that I can see.  The userData is really under client control, so they need to copy it if needed. 

I will add a comment about this to the copy() Javadoc.

jnh5y: +1, documenting it makes sense to me.

dr-jts:
Fixes #273

bjornharrtell:
@dr-jts it's unfortunate that copy was made into internal use only here. It was intentionally public and both solves the issue that clone is not suitable for transpilation (as it has very specific Java semantics) and for type safety. See https://github.com/locationtech/jts/pull/198.

bjornharrtell:
Ahm, sorry I failed to see that there is a public copy on Geometry which delegates to a copyInternal which should work and my troubles must be coming from another source.

PR #290, opened 2018-06-12
Remove LineString upcast when accessing Polygon rings

dbaston:
This makes it simpler for a user to construct new Polygons from one or more rings of an existing Polygon.

### DIFF
--- a/modules/app/src/main/java/org/locationtech/jtstest/function/GeometryFunctions.java
+++ b/modules/app/src/main/java/org/locationtech/jtstest/function/GeometryFunctions.java
@@ -59,13 +59,13 @@ public class GeometryFunctions
   public static Geometry getPolygonShell(Geometry g)
   {
     if (g instanceof Polygon) {
-      LinearRing shell = (LinearRing) ((Polygon) g).getExteriorRing();
+      LinearRing shell = ((Polygon) g).getExteriorRing();
       return g.getFactory().createPolygon(shell, null);
     }
     if (g instanceof MultiPolygon) {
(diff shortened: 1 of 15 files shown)


dr-jts:
The use of LineString as the return type was done to follow the OGC Simple Features Specification (see [OGC SFS 1.1.0](http://portal.opengeospatial.org/files/?artifact_id=13227) Section 6.1.11.2).  Not sure why they used LineString rather than LinearRing.

Any comments on whether it's better to be spec-compliant, or easier to use?

dbaston:
Hmm, I figured as much. I'll weasel around the tough question and make the argument that the PR _is_ spec-compliant, because a `LinearRing` is indeed a `LineString`.

dr-jts:
That seems reasonable to me.  Still odd that they didn't do this in the spec - but I can't see a reason for this, so perhaps it just slipped through the cracks (it's the kind of thing that might not be noticed until people actually started using the spec).

dbaston:
@dr-jts can this be merged, or should I close it out?

jodygarnett:
This has broken geotools build:

jodygarnett:
Looking at failure in downstream application:

jodygarnett:
Okay because this was an API change we should of updated the version number to `1.17.0-SNAPSHOT`.

I am going to see if I can pull over everything except this to the 1.16.x branch

dr-jts:
@jodygarnett There have been a lot of changes since 1.16.  To cut a 1.16.1 should probably just pull over the commits fixing CoordSequences and leave it at that.

PR #291, opened 2018-06-13
WKTReader/-Writer enhancement

FObermaier:
The current implementation of `WKTReader` uses `Coordinate` class to read the geometries' coordinates. This is not useful when it comes to reading `GEOMETRY M` or `GEOMETRY ZM` WKTs.   
This PR changes the behavior by creating a `CoordinateSequence` for every coordinate read and merges these sequences for the construction for the actual geometry type.   

Further enhancements:
* `WKTReader` can be configured to read `GEOMETRY M` with `CoordinateArraySequence` storing measure values at `Coordinate.Z` index (`setMeasureIsZ(boolean)`).
* `WKTWriter` can be configured to write `GEOMETRY M` off geometries using `CoordinateArraySequence` as store by writing ordinate values at `Coordinate.Z` as measure values (`setZToMeasure(boolean)`).
* The ordinates used for output can be configured for the `WKTWriter`.
* Automatic use of `PackedCoordinateSequenceFactory.Double` as coordinate sequence factory when reading `GEOMETRY ZM` WKT.
* Code cleanup

The behavior of the current `WKTReader` and `WKTWriter` class is preserved and can be configured.

Note: This can be further optimized by using [PR271](https://github.com/locationtech/jts/pull/271)


jodygarnett:
Thanks for separating this out from #271.

Can I take a moment to focus the use of coordinate sequence with z and m ordinates? 

a) JTS provides an example [ExtendedCoordinateSequence.java](https://github.com/locationtech/jts/blob/master/modules/example/src/main/java/org/locationtech/jtsexample/geom/ExtendedCoordinateSequence.java) that supports XYZM, but this is only an example and not intended for use.

b) Even though it was just an example we found JAI-EXT was using ExtendedCoordinateSequence, GeoTools also ended up using it as a default.

c) GeoTools has an interface [CoordinateAccess.java](https://github.com/geotools/geotools/blob/master/modules/plugin/jdbc/jdbc-oracle/src/main/java/org/geotools/data/oracle/sdo/CoordinateAccess.java) that extends CoordianteSequence to provide both getDimension() and getNumAttributes().

Your pull request also highlights the need for dimension+attribute coordinate sequences. 

Can I ask what you think of CoordinateAccess linked to above?

jodygarnett:
If I make use of Java 8 default methods we could update CoordinateSequence and Coordinate in place:

```
  default int  getNumberAttributes(){
       return 0;
  }
  double getAttribute(int coordinate, int attribute){
     if( attribute > 0 && attribute < getNumberAttributes()){
         return Double.NaN;
     }
     throw new IllegalArgumentException("Coordinate attribute "+attribute+" illegal, "+ getNumberAttributes() + " attributes stored");
  }
```

jodygarnett:
Also noting [WKTReader2](https://github.com/geotools/geotools/blob/master/modules/library/main/src/main/java/org/geotools/geometry/jts/WKTReader2.java) here.

FObermaier:
> Can I ask what you think of CoordinateAccess linked to above?

The approach is certainly valid but it would change the behavior of the existing `getOrdinate` `setOrdinate` function on `CoordinateSequence` at least for sequences that can handle more than x-, y- and z- ordinates.

FObermaier:
Regarding `WKTReader2`, the focus for your is on the support for Curves?

jodygarnett:
CoordinateAccess: It provides getAttribute and setAttribute (in order not to change behaviour).

WKTReader2 supports curves (using a subclass of linestring that define a coordinate sequence using control points).

FObermaier:
Taking the current implementation of `ExtendedCoordinateSequence` or `PackedCoordinateSequence`with a dimension of 4 (with measures), I understand your approach would return Double.NaN for calls to `getOrdinate(index, Coordinate.M)` because Coordinate.M is not "spatial"?

jodygarnett:
@jnh5y has encouraged me to review a lot of what is out there (PostGIS, GML, SDO, etc...) and it looks like having M as a *double* is fine and matches what other technologies are doing. 

GML3 has a complicated way of solving storing attributes on coordinates for those that need it, and it is the only example I found that was not XYZM.

FObermaier:
@jodygarnett, this should do.

modules/core/src/main/java/org/locationtech/jts/io/WKTOrdinates.java
```diff
@@ -0,0 +1,35 @@
...
+   * Measure-ordinate
+   */
+  M;
+
+  private static final EnumSet<WKTOrdinates> XY = EnumSet.of(X, Y);
+  private static final EnumSet<WKTOrdinates> XYZ = EnumSet.of(X, Y, Z);
+  private static final EnumSet<WKTOrdinates> XYM = EnumSet.of(X, Y, M);
+  private static final EnumSet<WKTOrdinates> XYZM = EnumSet.of(X, Y, Z, M);
+
+  public static EnumSet<WKTOrdinates> getXY() { return XY.clone(); }
```
jodygarnett: Do you just want this to be unmodifiable?

modules/core/src/test/java/org/locationtech/jts/io/WKTReadWriteTest.java
```diff
@@ -0,0 +1,91 @@
+package org.locationtech.jts.io;
```
jodygarnett: header needed

modules/core/src/main/java/org/locationtech/jts/io/WKTReader.java
```diff
@@ -222,170 +437,255 @@ public Geometry read(Reader reader) throws ParseException {
...
-		}
-		Coordinate[] array = new Coordinate[coordinates.size()];
-		return (Coordinate[]) coordinates.toArray(array);
-	}
-
-  private Coordinate getPreciseCoordinate()
+  private Coordinate getPreciseCoordinate(StreamTokenizer tokenizer)
       throws IOException, ParseException
   {
     Coordinate coord = new Coordinate();
```
jodygarnett: Should this method consider making different kinds of Coordinates? The CoordinateSequence implementations have a createCoordainte() method that may help, also the Coordinates utility class.

modules/core/src/main/java/org/locationtech/jts/geom/CoordinateSequence.java
```diff
@@ -57,6 +57,7 @@
    * <p>This constant assumes XYZM coordinate sequence definition, please check this assumption
    * using {@link #getDimension()} and {@link #getMeasures()} before use.
    */
+  /** Standard z-ordinate index */
   int Z = 2;
```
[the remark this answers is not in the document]
FObermaier: It might have slipped in when resolving conficts.

modules/core/src/main/java/org/locationtech/jts/io/WKTOrdinates.java
```diff
@@ -0,0 +1,35 @@
+package org.locationtech.jts.io;
+
+import java.util.EnumSet;
+
+/**
+ * An enumeration of possible Well-Known-Text ordinates
+ */
+public enum WKTOrdinates {
```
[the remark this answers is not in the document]
FObermaier: Should perhaps be renamed to just WKOrdinates since they might apply to WKB as well.

modules/core/src/main/java/org/locationtech/jts/io/WKTOrdinates.java
FObermaier: Yes, is there a better way?

modules/core/src/main/java/org/locationtech/jts/io/WKTOrdinates.java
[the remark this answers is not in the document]
jodygarnett: Ordinates sounds like a utility class, how about OrdinateSet ?

modules/core/src/main/java/org/locationtech/jts/io/WKTOrdinates.java
jodygarnett: Collections.unmodifiableSet( XY ) is what I am used to. Not perfect as it masks that it is an EnumSet, but good enough to prevent modification.

Guava provides an Sets.immutableEnumSet(enumSet) but I do not want the dependency.

jodygarnett: Recommending calling these createXY, createXYZ, etc... so it is clear a copy is being produced?

jodygarnett:
I have addressed the feedback above

modules/core/src/main/java/org/locationtech/jts/io/WKTReader.java
```diff
@@ -587,51 +889,12 @@ private MultiPoint OLDreadMultiPointText() throws IOException, ParseException {
...
-    points.add(point);
-    nextToken = getNextCloserOrComma();
-    while (nextToken.equals(COMMA)) {
-    	point = readPointText();
-      points.add(point);
-      nextToken = getNextCloserOrComma();
-    }
-    Point[] array = new Point[points.size()];
-    return geometryFactory.createMultiPoint((Point[]) points.toArray(array));
+    return geometryFactory.createMultiPoint(
```
dr-jts, line 894: @FObermaier This change causes the issue reported in #563.  Is there any reason reading EMPTY Points in a MULTIPOINT was removed?

FObermaier: It was not removed on purpose. It was changed to streamline reading Coordinate[Sequence]s and use the factory methods that take CoordinateSequences as arguments.
The unit test present did not sugguest that a `MULTIPOINT` containing EMPTY points is a case to be considered.

dr-jts: Fair enough.  The WKT handling of EMPTY is complex.  I will fix and add a unit test.

PR #293, opened 2018-07-01
Explicit CoordinateSequence support for XYZM

jodygarnett:
Checklist:

* [x] CoordinateSequence.getMeasures() default method
* [x] Update to PackedCoordinateSequence
* [x] discussion and feedback
* [x] Introduce Coordinate accessors for getX(), getY(), getZ(), getM()
* [x] Introduce explicit CoordinateXY, CoordinateXYM, CoordinateXYZM
* [x] Updated CoordinateTest to cover subclasses
* [x] Added CoordinateSequence hasZ and hasM default methods to perform common dimension and measures checks
* [x] Updated PackedCoordinateSequenceTest to try out XY, XYM, XYZ, XYZM functionality
* [x] Updated CoordinateArraySequence as well
* [x] Added CoordinateSequence.createCoordinate() default method, to assist with client use
* [x] Reworked ExtendedCoordinateSequence example

Ideally:

* [ ] proof of fitness with feedback from https://github.com/locationtech/jts/pull/291 WKT Reader/Writer enhancement

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/geom/Coordinate.java
+++ b/modules/core/src/main/java/org/locationtech/jts/geom/Coordinate.java
@@ -37,7 +37,7 @@ import org.locationtech.jts.util.NumberUtil;
  *
  *@version 1.7
  */
-public class Coordinate implements Comparable, Cloneable, Serializable {
+public class Coordinate implements Comparable<Coordinate>, Cloneable, Serializable {
   private static final long serialVersionUID = 6683108902428366910L;
   
   /**
@@ -258,7 +258,7 @@ public class Coordinate implements Comparable, Cloneable, Serializable {
    *@return    -1, zero, or 1 as this <code>Coordinate</code>
    *      is less than, equal to, or greater than the specified <code>Coordinate</code>
    */
-  public int compareTo(Object o) {
+  public int compareTo(Coordinate o) {
     Coordinate other = (Coordinate) o;
 
     if (x < other.x) return -1;
@@ -350,7 +350,7 @@ public class Coordinate implements Comparable, Cloneable, Serializable {
    * or 3-dimensional comparison, and handling NaN values correctly.
    */
   public static class DimensionalComparator
-      implements Comparator
+      implements Comparator<Coordinate>
   {
     /**
      * Compare two <code>double</code>s, allowing for NaN values.
@@ -407,11 +407,8 @@ public class Coordinate implements Comparable, Cloneable, Serializable {
      * equal to, or greater than 02
      *
      */
-    public int compare(Object o1, Object o2)
+    public int compare(Coordinate c1, Coordinate c2)
     {
-      Coordinate c1 = (Coordinate) o1;
-      Coordinate c2 = (Coordinate) o2;
-
       int compX = compare(c1.x, c2.x);
       if (compX != 0) return compX;
 
--- a/modules/core/src/main/java/org/locationtech/jts/geom/Coordinate.java
+++ b/modules/core/src/main/java/org/locationtech/jts/geom/Coordinate.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016 Vivid Solutions.
+ * Copyright (c) 2018 Vivid Solutions, and others.
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
@@ -19,13 +19,14 @@ import org.locationtech.jts.util.NumberUtil;
 
 
 /**
- * A lightweight class used to store coordinates
- * on the 2-dimensional Cartesian plane.
+ * A lightweight class used to store coordinates on the 2-dimensional Cartesian plane.
+ * <p>
  * It is distinct from {@link Point}, which is a subclass of {@link Geometry}. 
  * Unlike objects of type {@link Point} (which contain additional
  * information such as an envelope, a precision model, and spatial reference
  * system information), a <code>Coordinate</code> only contains ordinate values
  * and accessor methods. <P>
+ * <p>
  *
  * <code>Coordinate</code>s are two-dimensional points, with an additional Z-ordinate. 
  * If an Z-ordinate value is not specified or not defined, 
@@ -35,7 +36,7 @@ import org.locationtech.jts.util.NumberUtil;
  * Apart from the basic accessor functions, JTS supports
  * only specific operations involving the Z-ordinate. 
  *
- *@version 1.7
+ * @version 1.7
  */
 public class Coordinate implements Comparable<Coordinate>, Cloneable, Serializable {
   private static final long serialVersionUID = 6683108902428366910L;
@@ -47,37 +48,55 @@ public class Coordinate implements Comparable<Coordinate>, Cloneable, Serializab
    */
   public static final double NULL_ORDINATE = Double.NaN;
   
-  /**
-   * Standard ordinate index values
-   */
+  /** Standard ordinate index value for, where X is 0 */
   public static final int X = 0;
+
+  /** Standard ordinate index value for, where Y is 1 */
   public static final int Y = 1;
+  
+  /**
+   * Standard ordinate index value for, where Z is 2.
+   *
+   * <p>This constant assumes XYZM coordinate sequence definition, please check this assumption
+   * using {@link #getDimension()} and {@link #getMeasures()} before use.
+   */
   public static final int Z = 2;
 
   /**
-   *  The x-coordinate.
+   * Standard ordinate index value for, where M is 3.
+   *
+   * <p>This constant assumes XYZM coordinate sequence definition, please check this assumption
+   * using {@link #getDimension()} and {@link #getMeasures()} before use.
+   */
+  public static final int M = 3;
+  
+  /**
+   * The x-ordinate.
    */
   public double x;
+  
   /**
-   *  The y-coordinate.
+   * The y-ordinate.
    */
   public double y;
+  
   /**
-   *  The z-coordinate.
+   * The z-ordinate.
+   * @deprecated Recommend {@link #getZ()}
    */
   public double z;
 
   /**
    *  Constructs a <code>Coordinate</code> at (x,y,z).
    *
-   *@param  x  the x-value
-   *@param  y  the y-value
-   *@param  z  the z-value
+   *@param  x  the x-ordinate
+   *@param  y  the y-ordinate
+   *@param  z  the z-ordinate
    */
   public Coordinate(double x, double y, double z) {
     this.x = x;
     this.y = y;
-    this.z = z;
+    this.setZ(z);
   }
 
   /**
@@ -94,7 +113,7 @@ public class Coordinate implements Comparable<Coordinate>, Cloneable, Serializab
    *@param  c  the <code>Coordinate</code> to copy.
    */
   public Coordinate(Coordinate c) {
-    this(c.x, c.y, c.z);
+    this(c.x, c.y, c.getZ());
   }
 
   /**
@@ -115,12 +134,54 @@ public class Coordinate implements Comparable<Coordinate>, Cloneable, Serializab
   public void setCoordinate(Coordinate other) {
     x = other.x;
     y = other.y;
-    z = other.z;
+    setZ(other.getZ());
   }
 
   /**
+   *  The x-ordinate.
+   */
+  public double getX() {
+    return x;
+  }
+
+  public void setX(double x) {
+    this.x = x;
+  }
+  /**
+   *  The y-ordinate.
+   */  
+  public double getY() {
+      return y;      
+  }
+
+  public void setY(double y) {
+    this.y = y;
+  }
+  /**
+   *  The z-ordinate.
+   */  
+  public double getZ() {
+      return z;      
+  }
+  /**
+   * The z-ordinate.
+   *
+   * @param z
+   */
+  public void setZ(double z) {
+    this.z = z;
+  }
+  /**
+   *  The m-measure, if available.
+   */  
+  public double getM() {
+      return Double.NaN;     
+  }
+  
+  /**
    * Gets the ordinate value for the given index.
-   * The supported values for the index are 
+   * 
+   * The base implementation supports values for the index are 
    * {@link X}, {@link Y}, and {@link Z}.
    * 
    * @param ordinateIndex the ordinate index
@@ -132,7 +193,7 @@ public class Coordinate implements Comparable<Coordinate>, Cloneable, Serializab
     switch (ordinateIndex) {
     case X: return x;
     case Y: return y;
-    case Z: return z;
+    case Z: return getZ(); // sure to delegate to subclass rather than offer direct field access
     }
     throw new IllegalArgumentException("Invalid ordinate index: " + ordinateIndex);
   }
@@ -140,7 +201,8 @@ public class Coordinate implements Comparable<Coordinate>, Cloneable, Serializab
   /**
    * Sets the ordinate for the given index
    * to a given value.
-   * The supported values for the index are 
+   * 
+   * The base implementation supported values for the index are 
    * {@link X}, {@link Y}, and {@link Z}.
    * 
    * @param ordinateIndex the ordinate index
@@ -157,7 +219,7 @@ public class Coordinate implements Comparable<Coordinate>, Cloneable, Serializab
         y = value;
         break;
       case Z:
-        z = value;
+        setZ(value); // delegate to subclass rather than offer direct field access
         break;
       default:
         throw new IllegalArgumentException("Invalid ordinate index: " + ordinateIndex);
@@ -209,8 +271,8 @@ public class Coordinate implements Comparable<Coordinate>, Cloneable, Serializab
    */
   public boolean equals3D(Coordinate other) {
     return (x == other.x) && (y == other.y) &&
-               ((z == other.z) ||
-               (Double.isNaN(z) && Double.isNaN(other.z)));
+               ((getZ() == other.getZ()) ||
+               (Double.isNaN(getZ()) && Double.isNaN(other.getZ())));
   }
   
   /**
@@ -221,7 +283,7 @@ public class Coordinate implements Comparable<Coordinate>, Cloneable, Serializab
    * @return true if the Z ordinates are within the given tolerance
    */
   public boolean equalInZ(Coordinate c, double tolerance){
-    return NumberUtil.equalsWithTolerance(this.z, c.z, tolerance);
+    return NumberUtil.equalsWithTolerance(this.getZ(), c.getZ(), tolerance);
   }
   
   /**
@@ -274,7 +336,7 @@ public class Coordinate implements Comparable<Coordinate>, Cloneable, Serializab
    *@return    a <code>String</code> of the form <I>(x,y,z)</I>
    */
   public String toString() {
-    return "(" + x + ", " + y + ", " + z + ")";
+    return "(" + x + ", " + y + ", " + getZ() + ")";
   }
 
   public Object clone() {
@@ -316,7 +378,7 @@ public class Coordinate implements Comparable<Coordinate>, Cloneable, Serializab
   public double distance3D(Coordinate c) {
     double dx = x - c.x;
     double dy = y - c.y;
-    double dz = z - c.z;
+    double dz = getZ() - c.getZ();
     return Math.sqrt(dx * dx + dy * dy + dz * dz);
   }
 
@@ -417,7 +479,7 @@ public class Coordinate implements Comparable<Coordinate>, Cloneable, Serializab
 
       if (dimensionsToTest <= 2) return 0;
 
-      int compZ = compare(c1.z, c2.z);
+      int compZ = compare(c1.getZ(), c2.getZ());
       return compZ;
     }
   }
--- a/modules/core/src/main/java/org/locationtech/jts/geom/Coordinate.java
+++ b/modules/core/src/main/java/org/locationtech/jts/geom/Coordinate.java
@@ -99,7 +99,7 @@ public class Coordinate implements Comparable<Coordinate>, Cloneable, Serializab
   public Coordinate(double x, double y, double z) {
     this.x = x;
     this.y = y;
-    this.setZ(z);
+    this.z = z;
   }
 
   /**
@@ -137,7 +137,7 @@ public class Coordinate implements Comparable<Coordinate>, Cloneable, Serializab
   public void setCoordinate(Coordinate other) {
     x = other.x;
     y = other.y;
-    setZ(other.getZ());
+    z = other.getZ();
   }
 
   /**
--- a/modules/core/src/main/java/org/locationtech/jts/geom/Coordinate.java
+++ b/modules/core/src/main/java/org/locationtech/jts/geom/Coordinate.java
@@ -25,18 +25,21 @@ import org.locationtech.jts.util.NumberUtil;
  * Unlike objects of type {@link Point} (which contain additional
  * information such as an envelope, a precision model, and spatial reference
  * system information), a <code>Coordinate</code> only contains ordinate values
- * and accessor methods. <P>
+ * and accessor methods. </p>
  * <p>
- *
  * <code>Coordinate</code>s are two-dimensional points, with an additional Z-ordinate. 
  * If an Z-ordinate value is not specified or not defined, 
  * constructed coordinates have a Z-ordinate of <code>NaN</code>
  * (which is also the value of <code>NULL_ORDINATE</code>).  
  * The standard comparison functions ignore the Z-ordinate.
  * Apart from the basic accessor functions, JTS supports
- * only specific operations involving the Z-ordinate. 
+ * only specific operations involving the Z-ordinate.</p> 
+ * <p>
+ * Implementations may optionally support Z-ordiante and M-measure values
+ * as appropriate for a CoordinateSeqeunce. Use of {@link #getZ()} and {@link #getM()}
+ * accessors, or {@link #getOrdinate(int)} are recommended.</p> 
  *
- * @version 1.7
+ * @version 1.16
  */
 public class Coordinate implements Comparable<Coordinate>, Cloneable, Serializable {
   private static final long serialVersionUID = 6683108902428366910L;
@@ -175,7 +178,16 @@ public class Coordinate implements Comparable<Coordinate>, Cloneable, Serializab
    *  The m-measure, if available.
    */  
   public double getM() {
-      return Double.NaN;     
+    return Double.NaN;     
+  }
+  
+  /**
+   * The m-measure, if supported.
+   *
+   * @param m
+   */
+  public void setM(double m) {
+    throw new IllegalArgumentException("Invalid ordinate index: " + M);
   }
   
   /**
--- a/modules/core/src/main/java/org/locationtech/jts/geom/Coordinate.java
+++ b/modules/core/src/main/java/org/locationtech/jts/geom/Coordinate.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018 Vivid Solutions, and others.
+ * Copyright (c) 2018 Vivid Solutions
  *
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
--- a/modules/core/src/main/java/org/locationtech/jts/geom/CoordinateSequence.java
+++ b/modules/core/src/main/java/org/locationtech/jts/geom/CoordinateSequence.java
@@ -60,6 +60,14 @@ public interface CoordinateSequence
    * @return the dimension of the sequence.
    */
   int getDimension();
+  
+  /**
+   * Returns the number of measures in each coordinate for this sequence.
+   * @return
+   */
+  default int getNumberOfMeasures() {
+	  return 0;
+  }
 
   /**
    * Returns (possibly a copy of) the i'th coordinate in this sequence.
@@ -119,7 +127,7 @@ public interface CoordinateSequence
    * (for instance, they may contain other dimensions or measure values).
    *
    * @param index  the coordinate index in the sequence
-   * @param ordinateIndex the ordinate index in the coordinate (in range [0, dimension-1])
+   * @param ordinateIndex the ordinate index in the coordinate (in range [0, dimension-numberOfMeasures-1])
    */
   double getOrdinate(int index, int ordinateIndex);
 
@@ -133,7 +141,7 @@ public interface CoordinateSequence
    * Sets the value for a given ordinate of a coordinate in this sequence.
    *
    * @param index  the coordinate index in the sequence
-   * @param ordinateIndex the ordinate index in the coordinate (in range [0, dimension-1])
+   * @param ordinateIndex the ordinate index in the coordinate (in range [0, dimension-numberOfMeasures1])
    * @param value  the new ordinate value
    */
   void setOrdinate(int index, int ordinateIndex, double value);
@@ -164,7 +172,7 @@ public interface CoordinateSequence
    * Called by Geometry#clone.
    *
    * @return a copy of the coordinate sequence containing copies of all points
-   * @deprecated
+   * @deprecated Recommend {@link #copy()} 
    */
   Object clone();
   
--- a/modules/core/src/main/java/org/locationtech/jts/geom/CoordinateSequence.java
+++ b/modules/core/src/main/java/org/locationtech/jts/geom/CoordinateSequence.java
@@ -116,6 +116,18 @@ public interface CoordinateSequence
   }
 
   /**
+   * Creates a coordinate for use in this sequence.
+   * <p>
+   * The coordinate is created supporting the same number of {@link #getDimension()} and {@link #getMeasures()}
+   * as this sequence and is suitable for use with {@link #getCoordinate(int, Coordinate)}.
+   * </p>
+   * @return coordinate for use with this sequence
+   */
+  default Coordinate createCoordinate() {
+    return Coordinates.create(getDimension(), getMeasures());
+  }
+  
+  /**
    * Returns (possibly a copy of) the i'th coordinate in this sequence.
    * Whether or not the Coordinate returned is the actual underlying
    * Coordinate or merely a copy depends on the implementation.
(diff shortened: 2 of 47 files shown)


jodygarnett:
So reading some of the javadocs indicates that it is up to the client code to assign a meaning to dimensions beyond 2.

Son in GeoTools if we have a SRS with dimension 2 and a geometry with dimension 3 we would know we are looking at XYM data.

modules/core/src/main/java/org/locationtech/jts/geom/CoordinateSequence.java
```diff
@@ -60,6 +60,14 @@
    * @return the dimension of the sequence.
    */
   int getDimension();
+  
+  /**
+   * Returns the number of measures in each coordinate for this sequence.
+   * @return
+   */
+  default int getNumberOfMeasures() {
```
jodygarnett: here is the default interface

bjornharrtell:
Not sure, should be possible to fix in case of troubles, so don't let it be a blocker.

basisbit:
could you please specify what Z and M should represent? I guess for Z this is obvious for many use cases, some clarification would still be nice.

FObermaier:
Does the introduction of `getNumberOfMeasures` imply changing the meaning of `getDimensions` (as in GeoTools' CoordinateAccess interface? If so this would impose a breaking change for  `CoordinateSequence` implementations like `PackedCoordinateSequence` that are used with a dimension greater than 3.

dr-jts:
@FObermaier my thoughts on answers to your questions:

* the semantics of `CoordinateSequence.getDimension` should not change. It should continue to return the total number of ordinates available.
* I think @jodygarnett  meant `CoordinateSequence.getZ`.  This returns NaN if no Z dimension was defined.  
* Again, I think the suggestion is to add `CoordinateSequence.getM`.  

The Z and M accessors are suggested in order to make the support for Z and M explicit in the API (and to match the explicit support in the SFSQL spec).  

It might be good to provide a `CoordinateSequence.hasZ` method, in order to encapsulate the logic of determining this (and this can be made a default interface method).  `hasM `could be provided as well, but does have the fairly trivial equivalent of `getNumberOfMeasures == 1`

jodygarnett:
Guess I am always confused if M counts as a ordinate with respect to getDimension().

dr-jts:
Yeah, it's confusing.  In JTS dimension has always referred to the total number of ordinates available.  If we change that not sure what will happen...  

If `getDimension `is changed to be just the number of geometric dimensions, then pretty sure that a `getNumberOfOrdinates `should be available as well, for ease of iteration.

dr-jts:
Further to the question of the semantics of getDimension, my read of the SFSQL (below) is that getDimension should be the total number of ordinates, including the M value(s) if present, based on the second sentence.

> COORD_DIMENSION — the number of ordinates used in the complex, usually corresponds to the number of dimensions in the spatial reference system. **If an "M" ordinate is included it shall be one greater than the number of dimensions of the spatial reference system**.

jodygarnett:
You can see my quote above about GEOMETRY_TYPE having different values for POINT POINTZ POINTM POINTZM. If we went that approach and reported back the correct SFSQL type .. it would be a pain and not be visible at the CoordianteSequence level.

I would like to keep getDimension() in agreement with the SFSQL spec if we can:

> dimension = spatialDimension + numberOfOrdinates 

Would rather see an approach using numbers, rather than hasM() and hasZ() methods .. to allow for XYZMN etc...

dr-jts:
I don't understand `dimension = spatialDimension + numberOfOrdinates`. What is `numberOfOrdinates `?

Agreed that a numeric approach is needed (as you say, to allow for measures > 1).  But doesn't preclude  having the predicates too.

jodygarnett:
@dr-jts and @jnh5y can I ask for a review, I have responded I think to most of the feedback.

I ended up having CoordianteSequence.getDimension() as described in the conversation above, and CoordianteSequence.getMeasures() being the number of measures (so if we have XYZM dimension is 4 and measures is 1.

I also removed the mutability from PackedCoordianteSequence as the factory methods were provided, and there was no need for PackagedCoordinateSequence to hold on to state (especially when their were static constances for PackagedCoordinateSequence.DOUBLE and PackagedCoordinateSequence.FLOAD.

@jnh5y I have introduced the specific Coordiante subclasses as you requested, so we end up with:

* Coordinate - stores XYZ, the getM() method always returns Double.NaN
* CoordinateXY - both getZ() and getM() return Double.NaN. The field "z" is still visible from superclass but not used
* CoordinateXYM
* CoordinateXYZM

modules/core/src/main/java/org/locationtech/jts/geom/Coordinate.java
```diff
@@ -47,37 +48,55 @@
...
+   *
+   * <p>This constant assumes XYZM coordinate sequence definition, please check this assumption
+   * using {@link #getDimension()} and {@link #getMeasures()} before use.
+   */
+  public static final int M = 3;
+  
+  /**
+   * The x-ordinate.
    */
   public double x;
```
elahrvivaz: would it make sense to deprecate direct access to x and y as well? It seems potentially confusing to have `coord.x`, `coord.y` and `coord.getZ()` be the established access pattern

modules/core/src/main/java/org/locationtech/jts/geom/CoordinateSequence.java
```diff
@@ -111,15 +145,48 @@
...
 
+  /**
+   * Returns ordinate M of the specified coordinate if available.
+   * 
+   * @param index
+   * @return the value of the Z ordinate in the index'th coordinate, or Double.NaN if not defined.
+   */
+  default double getM(int index) {
+     if( getDimension() > 2 && getMeasures() > 0 ) {
+	 final int mIndex = getDimension()-getDimension();
```
elahrvivaz: should this be `getDimension() - getMeasures()`?

modules/core/src/main/java/org/locationtech/jts/geom/Coordinates.java
```diff
@@ -0,0 +1,93 @@
...
+     * 
+     * @param coordiante
+     * @return dimension 
+     */
+    public static int getDimension(Coordinate coordiante) {
+	if(coordiante instanceof CoordinateXY) {
+	    return 2;
+	}
+	else if(coordiante instanceof CoordinateXYM) {
+	    return 2;
```
elahrvivaz: should this be 3?

modules/core/src/main/java/org/locationtech/jts/geom/impl/PackedCoordinateSequence.java
```diff
@@ -250,19 +270,23 @@ public Double(float[] coordinates, int dimensions) {
...
       coords = new double[coordinates.length * this.dimension];
       for (int i = 0; i < coordinates.length; i++) {
         coords[i * this.dimension] = coordinates[i].x;
         if (this.dimension >= 2)
           coords[i * this.dimension + 1] = coordinates[i].y;
         if (this.dimension >= 3)
-          coords[i * this.dimension + 2] = coordinates[i].z;
+          coords[i * this.dimension + 2] = coordinates[i].getOrdinate(2); // Z or M
+        if (this.dimension >= 4)
+            coords[i * this.dimension + 2] = coordinates[i].getOrdinate(3); // M
```
elahrvivaz: should this be `+ 3`

modules/core/src/main/java/org/locationtech/jts/geom/CoordinateSequence.java
jodygarnett: yes, yes it should. And I should have test cases before this goes live :)

modules/core/src/main/java/org/locationtech/jts/geom/Coordinate.java
jodygarnett: I would like to do so, but discussion with @jnh5y and @dr-jts has me considering if having explicit classes, with access to the primitives, is required. Martin quite correctly focuses on these being primary a data structures, rather than an object.

With java value types on the horizon we are likely to change to the `struct` idea from C. Really want to have coordinates represent a block of [xy], [xyz], [xym], [xyzm] if continuous memory, so when we make an array of them everything is co-located in memory.

jodygarnett: So my design hat agrees with you, prefer getX() and getY(), make them final so the JRE can inline them to avoid performance penalty which is a key concern ...

dr-jts: Have to keep public access to Coordinate x and y fields for backwards compatibility.  These could be removed in a future non-compatible version (i,e, JTS 2)

dr-jts: Well, unless a big refactoring is done (as I see @jodygarnett has already done for the z field).

jodygarnett: just because refactor is done for JTS codebase, does not mean I want to accidentally break client code.

jodygarnett:
Okay folks test case written, this has collected all feedback provided.

modules/core/src/main/java/org/locationtech/jts/geom/Coordinate.java
jodygarnett: I really am not sure value types is on the java roadmap ...

dr-jts: The problem with marking as Deprecated is that that triggers compiler errors, which can be annoying.  Can comment them as deprecated, though.

jodygarnett: really? I thought it just made compiler warnings?

From the annotation javadocs:

> A program element annotated &#64;Deprecated is one that programmers
> are discouraged from using, typically because it is dangerous,
> or because a better alternative exists.  Compilers warn when a
> deprecated program element is used or overridden in non-deprecated code.

In anycase I have it commended it as deprecated as the javadoc approach allows a message to be provided

dr-jts: Well, yes, warnings not errors.  Still, it's nice to allow compiling warning-free.

elahrvivaz: my $0.02 - probably a good idea to comment and also mark them. the compiler warnings are useful to raise awareness that users should be moving away from the methods, whereas most wouldn't notice a comment change and then face an unexpected compile failure whenever they actually go away.

dsmiley: LOL emitting warnings so people change their code is entirely the point.  Then they change their code and the warnings go away

dr-jts: I have been in situations where compile-time warnings trigger a rejection of the update to the library, and it is not possible to change the client code.  So I'm very wary of deprecating things until at least a few releases containing the alternative functionality.

jratike80:
I suppose we do not want to read literally the citation from SFSQL because it seems to say that for example EPSG:4326 that is Ellipsoidal 2D CS it is not allowed to use XYZ or XYZM coordinates while XYM is OK with COORD_DIMENSION=3.

> COORD_DIMENSION — the number of ordinates used in the complex, usually corresponds to the number of dimensions in the spatial reference system. If an "M" ordinate is included it shall be one greater than the number of dimensions of the spatial reference system.

jratike80:
What is supposed to happen when XY(Z)M coordinates are passed to JTS operations which likely destroy the meaning of measures? For example, most often M is used for linear referencing and M does not make sense if geometries are buffered.

modules/core/src/main/java/org/locationtech/jts/geom/CoordinateSequence.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016 Vivid Solutions.
+ * Copyright (c) 2018 Vivid Solutions, and others.
```
dr-jts: I don't think this is a valid copyright?

modules/core/src/main/java/org/locationtech/jts/geom/CoordinateSequence.java
```diff
@@ -45,22 +45,76 @@
...
+   * <li>For XY sequence measures is zero</li>
+   * <li>For XYM sequence measure is one<li>
+   * <li>For XYZ sequence measure is zero</li>
+   * <li>For XYZM sequence measure is one</li>
+   * <li>Values greater than one are supported</li>
+   * </ul>
+   *
+   * @return the number of measures included in dimension
+   */
+  default int getMeasures() {
```
dr-jts: This might be better named getNumberOfMeasures, or getMeasureCount

dr-jts:
What are the use cases for CoordinateXY, CoordinateXYM, etc?  

Internally JTS just uses Coordinate as a intermediate object to allow easy computing and reference to coordinate values.  This can be served by Coordinate (and perhaps Coordinate with an added M value if and when measure-based algorithms are provided).

Otherwise it is encourage to just use the CoordinateSequence API.

If the other classes are not used, then perhaps there is no need to provide them?

Another issue is that they are hard to use programmatically, since the dimension is statically encoded in the class.

jodygarnett:
jratike80 we found inconsistent descriptions on the meaning of dimension (both including "m" and excluding "m" as described in your quote). We decided to go with including M, so that dimension describes the limit accepted by the getOrdiante method, if not client code is always stuck doing math...

jodygarnett:
@dr-jts the use case for CoordinateXY, CoordinateXYM etc... is from @jnh5y who wanted explicit implementations to better control memory use. If we had CoordinateXY with only an x and a y field it would be ideal, or as an interface to produce flyweight objects for CoordinateSequence access.

modules/core/src/main/java/org/locationtech/jts/geom/CoordinateSequence.java
jodygarnett: It is the format used by eclipse source code so I figured it was good:

modules/core/src/main/java/org/locationtech/jts/geom/CoordinateSequence.java
jodygarnett: I settled on shorter, after trying getNumberOfMeasures() initially.

jnh5y: Yeah, I'll say that I'd expect getMeasures to return an Array/List of the measures rather than a count.  Can we change the name?

modules/core/src/main/java/org/locationtech/jts/geom/CoordinateSequence.java
```diff
@@ -45,22 +45,76 @@
...
+  }
+  
+  /**
+   * Checks {@link #getDimension()} and {@link #getMeasures()} to determine if {@link #getZ(int)}
+   * is supported.
+   * 
+   * @return true if {@link #getZ(int)} is supported.
+   */
+  default boolean hasZ() {
+      return (getDimension()-getMeasures()) > 2;
```
jnh5y: Is there a situation where getDimension()-getMeasures() >= 2?

modules/core/src/main/java/org/locationtech/jts/geom/CoordinateSequence.java
```diff
@@ -45,22 +45,76 @@
...
+  }
+  
+  /**
+   * Checks {@link #getMeasures()} to determine if {@link #getM(int)}
+   * is supported.
+   * 
+   * @return true if {@link #getM(int)} is supported.
+   */
+  default boolean hasM() {
+      return getDimension()>2 && getMeasures() > 0;
```
jnh5y: @jodygarnett is there any used case for 1-d geometries with a measure?  Is checking getMeasures() > 0 sufficient?

modules/core/src/main/java/org/locationtech/jts/geom/CoordinateSequence.java
```diff
@@ -111,15 +165,49 @@
...
    * Ordinate indices 0 and 1 are assumed to be X and Y.
+   * <p>
    * Ordinates indices greater than 1 have user-defined semantics
-   * (for instance, they may contain other dimensions or measure values).
+   * (for instance, they may contain other dimensions or measure
+   * values as described by {@link #getDimension()} and {@link #getMeasures()}).
    *
    * @param index  the coordinate index in the sequence
-   * @param ordinateIndex the ordinate index in the coordinate (in range [0, dimension-1])
+   * @param ordinateIndex the ordinate index in the coordinate (in range [0, dimension-numberOfMeasures-1])
```
jnh5y: I'm not sure about the range here.  It seems to me that the range would still be [0, dimension-1].  

The bit from [0, dimension-numberOfMeasures-1] would be the XY(Z) values.  For numberOfMeasures > 0, the bit from [dimension-numberOfMeasures, dimension-1] would the measures.

modules/core/src/main/java/org/locationtech/jts/geom/CoordinateSequence.java
```diff
@@ -133,7 +221,7 @@
    * Sets the value for a given ordinate of a coordinate in this sequence.
    *
    * @param index  the coordinate index in the sequence
-   * @param ordinateIndex the ordinate index in the coordinate (in range [0, dimension-1])
+   * @param ordinateIndex the ordinate index in the coordinate (in range [0, dimension-numberOfMeasures1])
```
jnh5y: numberOfMeasures instead of numberOfMeasures1

modules/core/src/main/java/org/locationtech/jts/geom/Coordinate.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016 Vivid Solutions.
+ * Copyright (c) 2018 Vivid Solutions, and others.
```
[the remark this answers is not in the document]
basisbit: how about `Copyright (c) 2018 Vivid Solutions, and contributors.`? That is what most projects do and you don't have the overhead of contacting every previous contributor and hope they sign a copyleft contract.

modules/core/src/main/java/org/locationtech/jts/geom/CoordinateXY.java
```diff
@@ -0,0 +1,98 @@
...
+      throw new IllegalArgumentException("CoordianteXY dimension 2 does not support z-ordinate");
+  }  
+  
+  @Override
+  public double getOrdinate(int ordinateIndex) {
+      switch (ordinateIndex) {
+      case X: return x;
+      case Y: return y;
+      }
+      throw new IllegalArgumentException("Invalid ordinate index: " + ordinateIndex);
```
jnh5y: Should this be the 'default' case in the switch?

modules/core/src/main/java/org/locationtech/jts/geom/CoordinateXYZM.java
```diff
@@ -0,0 +1,96 @@
...
+        break;
+      case M:
+        m = value;
+        break;
+      default:
+        throw new IllegalArgumentException("Invalid ordinate index: " + ordinateIndex);
+    }
+  }
+  
+  public String toString() {
```
jnh5y: On the topic of the 'toString' methods...  It seems that the original Coordinate.toString has ()'s.  Should we be careful here?

modules/core/src/main/java/org/locationtech/jts/geom/Coordinates.java
```diff
@@ -0,0 +1,90 @@
...
+    if (dimension == 2) {
+      return new CoordinateXY();
+    } else if (dimension == 3 && measures == 0) {
+      return new Coordinate();
+    } else if (dimension == 3 && measures == 1) {
+      return new CoordinateXYM();
+    } else if (dimension == 4 && measures == 1) {
+      return new CoordinateXYZM();
+    }
+    return new Coordinate();
```
jnh5y: Hmm...  This seems a tough case to handle properly...

modules/core/src/main/java/org/locationtech/jts/geom/impl/CoordinateArraySequence.java
```diff
@@ -153,7 +201,12 @@ public Coordinate getCoordinateCopy(int i) {
   public void getCoordinate(int index, Coordinate coord) {
```
jnh5y: As a sanity check, this seems like an easy place for a user to do something silly.  If they pass in the wrong kind of Coordinate, they could get a no-Z/M-support error message, right?  

In JTS 2.0, it'd be great to see if we can be clever enough to handle this at the type-system.

modules/core/src/main/java/org/locationtech/jts/geom/impl/CoordinateArraySequence.java
```diff
@@ -170,6 +223,31 @@ public double getY(int index) {
...
+   * @see org.locationtech.jts.geom.CoordinateSequence#getZ(int)
+   */
+  public double getZ(int index)
+  {
+    if (hasZ()) {
+      return coordinates[index].getZ();
+    } else {
+      return Double.NaN;
+    }
+
```
jnh5y: Extra newline.

modules/core/src/main/java/org/locationtech/jts/geom/impl/CoordinateArraySequence.java
```diff
@@ -178,9 +256,9 @@ public double getOrdinate(int index, int ordinateIndex)
     switch (ordinateIndex) {
       case CoordinateSequence.X:  return coordinates[index].x;
       case CoordinateSequence.Y:  return coordinates[index].y;
-      case CoordinateSequence.Z:  return coordinates[index].z;
+      default:
+	      return coordinates[index].getOrdinate(ordinateIndex);
```
jnh5y: Sanity check, does one get Double.NaN or an exception if they go out of bounds?

modules/core/src/main/java/org/locationtech/jts/geom/impl/CoordinateArraySequenceFactory.java
```diff
@@ -78,4 +78,21 @@ public CoordinateSequence create(int size, int dimension) {
...
+      //throw new IllegalArgumentException("dimension must be <= 4");
+    }
+    if (measures > 1) {
+      measures = 1;
+      //throw new IllegalArgumentException("measures must be <= 1");
+    }
+    if (dimension < 2)
+      dimension = 2; // handle bogus dimension
+    if (dimension - measures < 2) {
+      throw new IllegalArgumentException("max spatial dimension 2 required");
```
jnh5y: Do you mean 'Minimum spatial dimension of 2 required'?  (I'm also a fan of dumping the inputs.  E.g. "Minimum spatial dimension of 2 required.  Input dimension was $dimension, input number of measures was $measures.")

modules/core/src/main/java/org/locationtech/jts/geom/impl/PackedCoordinateSequence.java
```diff
@@ -271,14 +300,14 @@ public Double(Coordinate[] coordinates, int dimension) {
      * @param coordinates
      */
     public Double(Coordinate[] coordinates) {
-      this(coordinates, 3);
+      this(coordinates, 3,0);
```
jnh5y: (Picky) space after the comma.

modules/core/src/test/java/org/locationtech/jts/geom/CoordinateTest.java
```diff
@@ -163,6 +163,82 @@ public void testDistance3D() {
...
+      Coordinate coord = new Coordinate(xym); // copy
+      assertEquals( xym, coord );
+      assertTrue( !xym.equalInZ(coord,0.000001) );
+      
+      coord = new Coordinate(1.0,1.0,1.0); // 2.5d
+      xym = new CoordinateXYM( coord ); // copy
+      assertEquals( xym, coord );
+      assertTrue( !xym.equalInZ(coord,0.000001) ); 
+  }
+  public void testCoordinateXYMZ() {
```
jnh5y: Should be  testCoordinateXYZM

modules/core/src/test/java/org/locationtech/jts/geom/impl/PackedCoordinateSequenceTest.java
```diff
@@ -36,5 +41,103 @@ public PackedCoordinateSequenceTest(String name)
...
+  
+  public void testDimensionAndMeasure()
+  {
+    CoordinateSequenceFactory factory = getCSFactory();
+    CoordinateSequence seq = factory.create(5, 2);
+    CoordinateSequence copy;
+    Coordinate coord;
+    Coordinate[] array;
+    
+    initProgression(seq);
```
jnh5y: Is there an easy way to share this test code with the other class?

modules/core/src/test/java/org/locationtech/jts/io/WKBTest.java
```diff
@@ -151,7 +151,7 @@ private void runWKBTest(String wkt) throws IOException, ParseException
 
 	private void runWKBTestPackedCoordinate(String wkt) throws IOException, ParseException {
 		GeometryFactory geomFactory = new GeometryFactory(
-				new PackedCoordinateSequenceFactory(PackedCoordinateSequenceFactory.DOUBLE, 2));
+				new PackedCoordinateSequenceFactory(PackedCoordinateSequenceFactory.DOUBLE));
```
jnh5y: What's the default dimension of the PackedCoordinateSequenceFactory?

pom.xml
```diff
@@ -193,7 +193,7 @@
                 <version>2.15</version>
                 <configuration>
                     <includes>
-                        <include>**/*Test.java</include>
+                        <include>**/*Test.`java`</include>
```
jnh5y: Does this affect anything?

modules/core/src/test/java/org/locationtech/jts/io/WKBTest.java
jodygarnett: dimension 3, measures 0, you can see in the constructors.

pom.xml
jodygarnett: eek! probably.

modules/core/src/test/java/org/locationtech/jts/io/WKBTest.java
jnh5y: Then this is a change, no?  Here we are testing getting a PCSF which creates things of dimension 2, right?

Also, a WKBReader should be able to return values which have any 2, 3, and 4 dimensions...

modules/core/src/test/java/org/locationtech/jts/geom/impl/PackedCoordinateSequenceTest.java
jodygarnett: Could place it in the superclass, the two tests are slightly different though. Since one store the coordinate instances (so == returns true) and one does not (so == returns false).

jnh5y:
@jodygarnett ok, I finished a read-through.  Looks pretty close.

modules/core/src/test/java/org/locationtech/jts/geom/impl/PackedCoordinateSequenceTest.java
jnh5y: Could have an override-able method around that check?

modules/core/src/main/java/org/locationtech/jts/geom/impl/PackedCoordinateSequence.java
jodygarnett: aside: I could not find any code formatting guidelines for the project, I setup some defaults that kind of matched in my IDE an hoped for the best.

modules/core/src/main/java/org/locationtech/jts/geom/Coordinates.java
jodygarnett: I think I should make a CoordinateSequence.createCoordiante() method, it will make the ExtendedCoordianteSequence example easier to work with as well. If not we are demanding that these subclasses be used which is poor design.

modules/core/src/main/java/org/locationtech/jts/geom/CoordinateXYZM.java
jodygarnett: So original is Coordinate:

```
    return "(" + x + ", " + y + ", " + z + ")";
```

And ExtendedCoordiante:

```
    String stringRep = x + " " + y + " m=" + m;
    return stringRep;
```

I am happy to go with '(', kind of like the use of 'm=' so it is clear when measures are in play.

modules/core/src/main/java/org/locationtech/jts/geom/CoordinateXYM.java
```diff
@@ -0,0 +1,121 @@
...
+   *
+   * <p>This constant assumes XYM coordinate sequence definition, please check this assumption using
+   * {@link #getDimension()} and {@link #getMeasures()} before use.
+   */
+  public static final int M = 2;
+
+  /** Default constructor */
+  public CoordinateXYM() {
+    super();
+    this.m = 0.0;
```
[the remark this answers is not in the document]
jodygarnett: Default constructor is:

> Constructs a <code>Coordinate</code> at (0,0,NaN).

modules/core/src/main/java/org/locationtech/jts/geom/CoordinateXY.java
jodygarnett: that would also be a valid way to represent it.

modules/core/src/test/java/org/locationtech/jts/io/WKBTest.java
jodygarnett: Oh here is the deal, the previous implementation of the factory was modal! So it held on to the init value `2` and would use it as a default for some of the create methods.

I removed this modal functionality since it introduced the possibility for error, especially with the static final factory implementations. Any thread could change the "default" messing up another thread.

Sot the above change is on purpose.

modules/core/src/main/java/org/locationtech/jts/geom/impl/CoordinateArraySequenceFactory.java
jodygarnett: fixed

modules/core/src/main/java/org/locationtech/jts/geom/impl/CoordinateArraySequence.java
jodygarnett: found it was a little inconsistent. My understanding is exception if you are out of bounds, DnD if you are in bounds buy the value is undefined.

modules/core/src/main/java/org/locationtech/jts/geom/impl/CoordinateArraySequence.java
jodygarnett: In this case we trust the `coord` set methods to handle things appropriately.

modules/core/src/main/java/org/locationtech/jts/geom/CoordinateXYM.java
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  public CoordinateXYM(double x, double y, double m) {
+    super(x, y, Coordinate.NULL_ORDINATE);
+    this.m = m;
+  }
+
+  public CoordinateXYM(Coordinate coord) {
+    super(coord.x,coord.y);
+    m = getM();
```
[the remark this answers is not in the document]
jodygarnett: inconsistent, `0` for a spatial dimension (XY), DnD for an optional value (Z)

modules/core/src/main/java/org/locationtech/jts/geom/Coordinate.java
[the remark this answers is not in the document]
jodygarnett: I don't mind either way, "and others" was how the eclipse projects were doing it.

Really we can ask @jnh5y as a locaitontech committer representative to look this up for us and get back with an answer.

jodygarnett:
Thanks @jnh5y I will update based on your feedback, and then we really have to call it done as I do not have more time to put towards it.

Based on the above how do you feel about:

* CoordianteSequence.createCoordiante()
* Either going ahead with `(c) Vivid Solutions, and others` or reverting to `(c) Vivid Solutions`. Can you ask and let us know what to do?

jratike80:
I can read "coordiante" in quite many places, like CoordianteSequence.createCoordiante(). Are they typos or something intentional?

jodygarnett:
typos, responding on my phone auto corrects

modules/core/src/main/java/org/locationtech/jts/geom/Coordinate.java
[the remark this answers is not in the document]
jnh5y: From looking at https://www.eclipse.org/projects/handbook/#ip-copyright-headers, it looks like "and Contributors to the Eclipse Foundation" may be acceptable.  (I'd be comfortable with that since it is listed in the Eclipse documentation.)

modules/core/src/main/java/org/locationtech/jts/geom/CoordinateSequence.java
jnh5y: Should you update this line of documentation then?

modules/core/src/main/java/org/locationtech/jts/geom/CoordinateSequence.java
jnh5y: Remaining typo.

modules/core/src/main/java/org/locationtech/jts/geom/CoordinateXYZM.java
jnh5y: +1, I like it.

modules/core/src/main/java/org/locationtech/jts/geom/Coordinates.java
```diff
@@ -0,0 +1,90 @@
+/*
+ * Copyright (c) 2016 Vivid Solutions
```
jnh5y: Copyright clean-up.

modules/core/src/test/java/org/locationtech/jts/io/WKBTest.java
jnh5y: +1

jnh5y:
@dr-jts Any more comments on this?  I'd like to merge this PR later this week if there are no understanding issues.

modules/core/src/main/java/org/locationtech/jts/geom/Coordinate.java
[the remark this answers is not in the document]
jodygarnett: I am just reverting it, we can sort out headers another time without delaying this PR.

PR #309, opened 2018-09-04
Revert use of string.format in WKTWriter

### LINKED ISSUE #308: WKTWriter toLineString methods do not output full precision
A recent commit to WKTWriter breaks the toLineString method output, because the use of Java formatting only outputs a limit number of decimals of precision.

This should be fixed by either reverting to the original code or using a different format.

It would be helpful to factor the coordinate string conversion into a common method.

Would also be good to have some unit tests covering this code.

FObermaier:
* Add unit test of static WKTWriter.toPoint and WKTWriter.toLineString
  functions.
* fixes #308

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/io/WKTWriter.java
+++ b/modules/core/src/main/java/org/locationtech/jts/io/WKTWriter.java
@@ -18,9 +18,21 @@ import java.io.Writer;
 import java.text.DecimalFormat;
 import java.text.DecimalFormatSymbols;
 import java.util.EnumSet;
-import java.util.Locale;
 
-import org.locationtech.jts.geom.*;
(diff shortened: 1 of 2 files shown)


dr-jts:
Thanks, Felix.  

Just one thing - it would be nice to see a unit test which includes explicit constant data.  This does two things: (1) makes the effect of the function being tested very clear, and (2) since no randomization is used it is highly repeatable.  So using a constant such as 

LINESTRING (1.1234567890123456 1.1234567890123456, 2.9999999999 2.0000000001)

where the decimals are chosen to stress the edge cases of the numeric conversion.  (It may be that the default Java double to string conversion is actually subject to roundoff error or undesired truncation.  The only way to find out is to test it.  This is outside the scope of your change, but getting a test harness in place with some "normal" data is a good step towards this).

PR #326, opened 2018-10-09
AffineTransformation.rotate uses the wrong rotationInstance method

zhenyanghua:


dr-jts:
Good find.  A unit test for this would be great to have as well.

dr-jts:
The unit tests look good. But the last commit doesn't look right, and it's now failing IP validation.  Can you remove the last commit?

zhenyanghua:
That seems like an eclipse sign-off issue. any hint?

dr-jts:
Yes, it's a sign-off issue. Did you sign off both commits?

zhenyanghua:
I did for the first commit directly via github editor. As for the second one, I pushed through my local repo, which failed the check.

dr-jts:
Perhaps you have not signed an ECA as describe in the [Contributing guide](https://github.com/locationtech/jts/blob/master/CONTRIBUTING.md)?

zhenyanghua:
Well, I know that I had to amend my commit in order to do an empty push to trigger the hooks. It seems working now.

Issue #329, opened 2018-10-11
Simplifications  & Enhancements to Coordinate classes


dr-jts:
It seems like it would be reasonable to:

* Do **not** specify X and Y ordinate indices in subclasses of `Coordinate`, since they are inherited
* Do **not** specify ordinate indices in subclasses if they are not valid for that class (e.g. Z and M)
* Do **not** set the values of unsupported ordinates in `setCoordinate `overrides (or maybe set them to `NaN`)

None of these will change the API, so can be fixed now.

Also it might be nice to add `getDimension ()`and `getMeasures() `methods to `Coordinate `and subclasses, since those are known statically from the Coordinate definition.

Although one possible point of confusion is that if a `CoordinateSequence `does not use an appropriate Coordinate subclass then the dimension and measures count would not be in synch.

See also #372 and #349

dr-jts:
@jodygarnett  thoughts?

dr-jts:
@jnh5y thoughts?

dr-jts:
Also wondering if it would make sense to add `getDimension `and `getMeasures `methods to Coordinate and subclasses?

jnh5y:
@dr-jts my main thought/consideration about simplifying Coordinate would be any options that'd let us 'combine' Coordinate and Point.  This especially gets more interesting when we consider Value Classes coming to the JVM sometime...

If we could get that right, we'd could likely make some stronger claims about memory layout and push some optimizations there.

jodygarnett:
GML3 combines Coordinate and Point (or rather lets a Point be used as a Coordinate) never found the extra expressability worth the trade off in complexity.

jodygarnett:
Returning to @dr-jts question:

```
public class Coordinate implements Comparable<Coordinate>, Cloneable, Serializable {
public static final int X = 0;
public static final int Y = 1;
public static final int Z = 2;
...
}
public class CoordinateXY extends Coordinate {
  public static final int X = 0;
  public static final int Y = 1;
  public static final int Z = -1;
}
```

Static constants have no concept of inheritance (indeed you will often get a warning when using a constant via a subclass reference since the result is brittle if anyone later introduces a static field with the same name).

In this case we are defining `X`,`Y`,`Z` constants for `CoordianteXY` to both avoid this warning and document the fact that `CoordianteXY.Z` is unused and unusable.

jodygarnett:
[...]

PR #333, opened 2018-10-28
Fix inconsistency with CoordinateArraySequence using M measures

mukoki:
Throws exception if CoordinateArraySequence has M values and coord is just an instance of Coordinate in
CoordinateArraySequence#getCoordinate(index,coord)
This happens, for example, in PointLocation code, but this may not be the only place.
The solution in this PR is not nice (instanceof tests), but I miss a hasM() method in Coordinate class to do it more properly. Any reason why such a method has not been added with all the "measure" stuff ?

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/geom/impl/CoordinateArraySequence.java
+++ b/modules/core/src/main/java/org/locationtech/jts/geom/impl/CoordinateArraySequence.java
@@ -220,14 +220,7 @@ public class CoordinateArraySequence
    * @see org.locationtech.jts.geom.CoordinateSequence#getX(int)
    */
   public void getCoordinate(int index, Coordinate coord) {
-    coord.x = coordinates[index].x;
-    coord.y = coordinates[index].y;
-    if( hasZ()) {
-      coord.setZ(coordinates[index].getZ());
-    }
-    if( hasM() && (coord instanceof CoordinateXYM || coord instanceof CoordinateXYZM)) {
-      coord.setM(coordinates[index].getM());
-    }    
+    coord.setCoordinate(coordinates[index]);
   }
 
   /**
--- a/modules/core/src/main/java/org/locationtech/jts/geom/impl/CoordinateArraySequence.java
+++ b/modules/core/src/main/java/org/locationtech/jts/geom/impl/CoordinateArraySequence.java
@@ -13,12 +13,7 @@ package org.locationtech.jts.geom.impl;
 
 import java.io.Serializable;
 
-import org.locationtech.jts.geom.Coordinate;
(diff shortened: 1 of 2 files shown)


dr-jts:
Are you saying that PointLocation.isOnLine will fail if a CoordinateArraySequence withM provides a Coordinate with no M?  

Or is this just an extra consistency check?  If so, it seems like it would be better done at creation time?

dr-jts:
As for the idea of adding `Coordinate.hasM()`, that is a good idea.  I proposed a similar improvement in #329 (but slightly more indirectly, as `getDimension` and `getMeasures` - so `hasM = getMeasures == 1`

dr-jts:
I see the Travis CI is failing because of some test code I commited to the TestBuilder.  I will fix this ASAP.

mukoki:
As demonstrated by the unit test, PointLocation.isOnLine fails with a Geometry containing XYZM coordinates because it internally uses CoordinateArraySequence.getCoordinate(index, coord) which copy each original ordinate (including M) to a new Coordinate() (which does not support M).

dr-jts:
And probably an even better solution is to allow creating new instances of the actual `Coordinate `subclass used by the `CoordinateSequence`.  This probably  needs to be done by a method on the CoordinateSequence interface.  

It's tempting to rely on the `Coordinate.copy()` method, but that won't work if the sequence does not have any coordinates, and it makes for slightly obscure code (i.e. copying an existing coordinate even if there is no intention to use its value).

mukoki:
1 - Not throwing an exception for Coordinate.setM() seems acceptable to me.
2 - If we had a hasM() method in Coordinate, we could implement CoordinateArraySequence.getCoordinate(index,coord) in a more robust way, using the smallest dimension used by CoordinateSequence and by the second coord argument. Seems the option with the least side effect.
3 - Not sure if Coordinate.copy can save us : coord throwing exception is a method argument, I think we cannot change it (it is passed by PointLocation.isOnLine algo in my case). Or did I miss something ?

dr-jts:
Another idea:  `Coordinate.setCoordinate()` handles setting the fields correctly.   And `Coordinate `ordinate accessors do not throw if an ordinate does not exist (they return NaN as per design).   So should be able to use `Coordinate.setCoordinate()` safely in `CoordinateArraySequence.getCoordinate` ?

And agreed that `Coordinate.set` methods should handle missing ordinates silently, without throwing.

mukoki:
You mean
public void getCoordinate(int index, Coordinate coord) {
    coord.setCoordinate(coordinates[index]);
}
Seems the best idea so far ;-)

dr-jts:
Excellent.  Surprisingly tricky to get the Coordinate subclasses right - as I expected.

PR #335, opened 2018-10-30
added point basic example

henriguy1210:
There was only a check of the intersection of two lines that are exactly the same. So I added point on a line example.


dr-jts:
Can you remove the doc change as per comment?

henriguy1210:
removed doc change commit.

PR #340, opened 2018-11-06
Fix method spelling: tesWritetRepeatedSqrt -> testWritetRepeatedSqrt

schwehr:


schwehr:
I just signed the ECA.  I'm guessing it takes a few for the check to register the change.

modules/core/src/test/java/org/locationtech/jts/math/DDIOTest.java
```diff
@@ -175,7 +175,7 @@ private void checkParseError(String str) {
 		assertTrue(foundParseError);
 	}
 
-	public void tesWritetRepeatedSqrt()
+	public void testWritetRepeatedSqrt()
```
dr-jts: Actually this should be `testWriteRepeatedSqrt`.  (note removal of erroneous 3rd 't').

If you change and commit that might clear up the IP validation error too.

schwehr:
Forced a signed single commit over the two prior commits.

PR #341, opened 2018-11-06
GeometryStrategies: Fix check of empty arg.text and replaceAll escaping

schwehr:
Found by the compiler:
  [EqualsIncompatibleType] Calling equals on incompatible types String and StringBuffer
  [InvalidPatternSyntax] Invalid syntax used for a regular expression: Unexpected internal error near index 1

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/io/gml2/GeometryStrategies.java
+++ b/modules/core/src/main/java/org/locationtech/jts/io/gml2/GeometryStrategies.java
@@ -310,7 +310,7 @@ public class GeometryStrategies{
 				// one child, either a coord
 				// or a coordinate sequence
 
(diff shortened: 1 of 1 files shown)


jnh5y:
@schwehr that error is coming up since you didn't 'sign' your commit with 'git commit -s'.  It is a silly thing, but Eclipse requires it.  (Signing off a commit like that is a way for you to signify that the changes are yours and you can submit them to the project.)

schwehr:
@jnh5y Thanks!  I'll redo the commits as signed and repush them.  Thanks!  And now I even see it in the contributing doc.  Sorry for the noise.

jnh5y:
No worries; thanks for sticking with it and knocking out those changes!

PR #348, opened 2018-11-13
Removed some unused local variables

schwehr:


schwehr:
Forgot to sign the commit

schwehr:
Overwrote with a signed commit

modules/core/src/main/java/org/locationtech/jts/triangulate/quadedge/QuadEdgeSubdivision.java
```diff
@@ -803,12 +803,12 @@ public void visit(QuadEdge[] triEdges) {
 
 		private void checkTriangleSize(Coordinate[] pts)
 		{
-			String loc = "";
+			// String loc = "";
 			if (pts.length >= 2)
-				loc = WKTWriter.toLineString(pts[0], pts[1]);
+				/* loc = */ WKTWriter.toLineString(pts[0], pts[1]);
```
dr-jts: Entire line should be commented

modules/core/src/main/java/org/locationtech/jts/triangulate/quadedge/QuadEdgeSubdivision.java
```diff
@@ -803,12 +803,12 @@ public void visit(QuadEdge[] triEdges) {
...
 			else {
 				if (pts.length >= 1)
-					loc = WKTWriter.toPoint(pts[0]);
+					/* loc = */ WKTWriter.toPoint(pts[0]);
```
dr-jts: Comment entire line

modules/core/src/main/java/org/locationtech/jts/geomgraph/Edge.java
```diff
@@ -178,9 +178,8 @@ public void addIntersection(LineIntersector li, int segmentIndex, int geomIndex,
       /**
       * Add the intersection point to edge intersection list.
       */
-      EdgeIntersection ei = eiList.add(intPt, normalizedSegmentIndex, dist);
```
dr-jts: Don't see need to remove this

modules/core/src/main/java/org/locationtech/jts/triangulate/quadedge/QuadEdgeSubdivision.java
schwehr: It then looks like the entire checkTriangleSize should be commented out, yes?  Unless you have a better suggestion, that's what my next commit will do. Or what is the check supposed to do?  The assert is already commented out.  Is this somehow relying upon side effects of something?

I had assumed that the toLineString and toPoint were there for side effects.  Now that I look at them, I don't see any checks.

Pardon my confusion as I don't understand what checkTriangleSize is supposed to do.

modules/core/src/main/java/org/locationtech/jts/geomgraph/Edge.java
schwehr: Doh!  That's not what I meant to do.  Only the ei local var is unused.  I presume the add critical.  I'll push a fix in a bit.  And I could use guidance on how you'd like the style.  e.g. I'm mixing `/* */` and `//` here.  That ok?  Just use `/* */` or remove them all?

```java
      /* EdgeIntersection ei = */ eiList.add(intPt, normalizedSegmentIndex, dist);                                                                               
      // ei.print(System.out);                                                                                                                                   
```

modules/core/src/main/java/org/locationtech/jts/triangulate/quadedge/QuadEdgeSubdivision.java
dr-jts: It was used for debugging.  See line above with a call to it commented out.

It could be commented out, but I don't see a problem with just leaving it.

dr-jts:
Yes, the add is critical.

I would prefer to have debugging code left in place, even if commented
out.  In this case, I don't see the harm in simply leaving the assignment,
since it allows the debug code to be easily uncommented and used.

schwehr:
@dr-jts Thanks for the feedback.  I'm fairly new to JTS and GeoTools (and Java) so I'm trying out changes to see if I can find a path that works for you that reduces the toolchain complaints that I'm getting.  I'm currently getting ~200 complaints.  Many of them are along the lines of unused or unread things.  With GDAL, I found that working through all of the complaints like this found a significant number of actual issues (maybe 5% of the tool complaints when I started).  A couple of us are trying to do a review of the JTS changes since the switch to GIT.   Hopefully, we can quickly iterate towards a style of pull request suggestions that isn't too much work.  And being used to C++, my brain is trying to push me to do `#if`, which isn't going to fly.  I would love a solution that makes the tools understand that this isn't a bug, makes it obvious that these are just for debugging, and doesn't cause performance issues.  If that's not possible, let me know and I'll revert that part of the change.

dr-jts:
Agreed that it is good to review warnings, since they sometimes uncover bugs.

What is your toolchain?   Adding a source-level ignore for warnings is likely to be tool-dependent.  Which for obvious reasons doesn't seem like a good idea.

Failing that, the approach of commenting out debug code, or removing it if it is just unused, seems like the best approach.

JTS does now have it's own `Debug `framework, and I try and use that if it doesn't cause performance issues.  (It also makes it easy to find debugging code and comment it out.)

schwehr:
I'm mostly working inside of Google, so it's bazel based builds on linux in lots of containers in borg.  e.g.  https://cacm.acm.org/magazines/2016/7/204032-why-google-stores-billions-of-lines-of-code-in-a-single-repository/fulltext  I do a lot of back and forth between the piper vcs and git (often with copybara).

dr-jts:
What is complaining about the dead code?  Javacc?  Or some linting tool?

schwehr:
It's a linting tool.  Why don't I back this PR up a bit and go for the simple cleanups that don't look like they are have anything really interesting for debugging?  Should just be a couple things.

dr-jts:
Sounds good.

schwehr:
@dr-jts I think I've got it, but please give it a double check.

dr-jts:
Looks good.  I can merge tomorrow.

schwehr:
Awesome.  Thanks!

PR #349, opened 2018-11-14
Simplify logic in CoordinateSequence.java:hasM.

### LINKED ISSUE #347: Confusing logic in hasM
Why is `getMeasures() > 0` not sufficient?  Can someone who knows add a comment as to why?

https://github.com/locationtech/jts/blob/8ef27d434edbdebf96d9b61be1810319a435ae7d/modules/core/src/main/java/org/locationtech/jts/geom/CoordinateSequence.java#L114

```java
  /**
   * Checks {@link #getMeasures()} to determine if {@link #getM(int)}
   * is supported.
   * 
   * @return true if {@link #getM(int)} is supported.
   */
  default boolean hasM() {
      return getDimension()>2 && getMeasures() > 0;
  }
```

schwehr:
- The getDimension()>2 was redundant.
- Improve javadoc with a suggestion by dr-jts.

Fixes #347

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/geom/CoordinateSequence.java
+++ b/modules/core/src/main/java/org/locationtech/jts/geom/CoordinateSequence.java
@@ -105,15 +105,19 @@ public interface CoordinateSequence
   default boolean hasZ() {
       return (getDimension()-getMeasures()) > 2;
   }
-  
+
   /**
-   * Checks {@link #getMeasures()} to determine if {@link #getM(int)}
-   * is supported.
-   * 
+   * Tests whether the coordinates in the sequence have measures associated with them. Returns true
+   * if {@link #getMeasures()} > 0. See {@link #getMeasures()} to determine the number of measures
+   * present.
+   *
    * @return true if {@link #getM(int)} is supported.
+   *
+   * @see #getMeasures()
+   * @see #getM(int)
    */
   default boolean hasM() {
-      return getDimension()>2 && getMeasures() > 0;
+      return getMeasures() > 0;
   }
 
   /**


modules/core/src/main/java/org/locationtech/jts/geom/CoordinateSequence.java
```diff
@@ -105,15 +105,16 @@ default int getMeasures() {
...
       return (getDimension()-getMeasures()) > 2;
   }
-  
+
   /**
-   * Checks {@link #getMeasures()} to determine if {@link #getM(int)}
-   * is supported.
-   * 
+   * Tests whether the coordinates in the sequence have measures associated with them. Returns true
+   * if {@link #getMeasures()} > 0. See {@link #getMeasures()} to determine the number of measures
```
dr-jts: Actually can also use @see Javadoc for this

schwehr: @dr-jts : It seems like `@see` does not cope with text after?  So I think `@link` might be best.  If you have a way that `@see` will work, can you tell me what it is?

https://docs.oracle.com/javase/7/docs/technotes/tools/windows/javadoc.html#see

schwehr:
I tried again, but don't feel particularly confident about the javadoc as I'm still getting to know the basics of JTS.

modules/core/src/main/java/org/locationtech/jts/geom/CoordinateSequence.java
```diff
@@ -105,15 +105,19 @@ default int getMeasures() {
...
+   * present.
+   *
    * @return true if {@link #getM(int)} is supported.
+   *
+   * @see #getMeasures()
+   * @see #getM(int)
    */
   default boolean hasM() {
-      return getDimension()>2 && getMeasures() > 0;
+      return getMeasures() > 0;
```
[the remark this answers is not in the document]
jodygarnett: Think I was trying to put as many constraints as I could on this (there is a pathological case of a 1 dimensional coordinate measuring linear distance).

getMeasures() > 0 is fine and nice and readable.

PR #363, opened 2019-01-08
Improvements to IndexedFacetDistance

dr-jts:
* Add nearestLocations and nearestPoints methods
* Make thread-safe


airbreather:
Should this have also updated [`DistanceTest`](https://github.com/locationtech/jts/blob/e59cf2cb9e14edb878261a7c0abe13b0be689efb/modules/core/src/test/java/org/locationtech/jts/operation/distance/DistanceTest.java) to extend `AbstractDistanceTest`?  Otherwise it looks kinda silly having this abstract class that only `IndexedFacetDistanceTest` implements...

dr-jts:
Yes, probably.  I was following a countervailing principle of software development: If it ain't broke, don't fix it.

PR #368, opened 2019-01-28
Hilbert Curve and Morton Curve

dr-jts:
Hilbert Curve and ordering
Morton Curve and ordering

### DIFF
--- a/doc/JTS_Version_History.md
+++ b/doc/JTS_Version_History.md
@@ -22,7 +22,8 @@ Distributions for older JTS versions can be obtained at the
 
 ### Functionality Improvements
 
-* Added `HilbertCurveBuilder`.
+* Added `HilbertCode` and `HilbertCurveBuilder`.
+* Added 'MortonCode` and `MortonCurveBuilder`.
 
 ### Bug Fixes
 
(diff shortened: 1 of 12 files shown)


jnh5y:
@dr-jts nice!  You know about SFCurve as a project, right?  https://github.com/locationtech/sfcurve

Shout if I can ever help with space-filling curves for JTS.:)

jnh5y:
Interesting... are you looking at Hilbert R-Trees as static or dynamic data structures?  (I skimmed the Wikipedia article, and I haven't looked enough to understand the benefits.)

Incidentally, if the Hilbert implementation matters, @cne1x could come and share what he knows about 'compact' Hilbert curves.  (They are useful if the magnitude of the various dimensions are unequal.  E.g., if you have 10 bits of info for x and 6 bits for y.)

dr-jts:
As a alternate packing method for the current static R-tree structure.  (Which incidentally really needs to be refactored to support this, i.e. to separate the packing strategy from the actual R-tree structure and algorithms.  I'm also going to dump the SIRtree complexity, since it's no longer used in JTS having been superceded by the `SortedPackedIntervalRtree`.

PR #374, opened 2019-02-15
Change CoordinateArraySequence to check consistency of array Coordinates rather than fixing

dr-jts:
Signed-off-by: Martin Davis <mtnclimb@gmail.com>

PR #376, opened 2019-02-18
Functionality change for ArrayCoordinateSequence with inconsistent coordinate types

jodygarnett:
Two changes made after discussion of #374 

* [x] ArrayCoordinateSequence will leave null values in place
* [x] ArrayCoordinateSequence will create a copy of the array if modification is required

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/geom/impl/CoordinateArraySequence.java
+++ b/modules/core/src/main/java/org/locationtech/jts/geom/impl/CoordinateArraySequence.java
@@ -12,6 +12,7 @@
 package org.locationtech.jts.geom.impl;
 
 import java.io.Serializable;
+import java.lang.reflect.Array;
 
 import org.locationtech.jts.geom.*;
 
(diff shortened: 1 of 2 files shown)


jodygarnett:
aside: This will need to be applied to master also :)

modules/core/src/main/java/org/locationtech/jts/geom/impl/CoordinateArraySequence.java
```diff
@@ -82,13 +83,14 @@ public CoordinateArraySequence(Coordinate[] coordinates, int dimension) {
...
   {
-    this.coordinates = coordinates;
     this.dimension = dimension;
     this.measures = measures;
     if (coordinates == null) {
       this.coordinates = new Coordinate[0];
     }
-    enforceArrayConsistency( this.coordinates );
+    else {
+      this.coordinates = enforceArrayConsistency(coordinates);
```
jodygarnett: This provides the "make a copy if needed"

PR #382, opened 2019-02-25
degrade javadoc comment of commented out method

tyrasd:
So that the javadoc builder don't mistake it for the documentation of the `isRectangle` method: 

[selection_208](http://locationtech.github.io/jts/javadoc/org/locationtech/jts/geom/Polygon.html#isRectangle())


dr-jts:
Thanks. I'll make a PR to fix this by removing the commented method.

dr-jts:
Fixed by 9d481ab78d25f7982e4546487ee9caff2082f4a2

PR #386, opened 2019-03-01
Improve PackedCoordinateSequence and -Factory class

FObermaier:
* add checkDimension method, call it during sequence construction
* add constructor with CoordinateSequence as argument to both
  PackedCoordinateSequence.Double and .Float
* use ordinate index constants instead of plain values
* add missing integrety checks to constructors
* add INITIAL_Z_VALUE along with get- and set methods. Call
  initializeZValues where required
* share DEFAULT_DIMENSION and DEFAULT_MEASURE in factory and sequence
  class
* add DEFAULT_MEASURE_VALUE to CoordinateXYM and use it in constructors
  of CoordinateXYM and CoordinateXYZM
* add unit tests

Closes #268 

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/geom/impl/PackedCoordinateSequence.java
+++ b/modules/core/src/main/java/org/locationtech/jts/geom/impl/PackedCoordinateSequence.java
@@ -390,6 +390,11 @@ public abstract class PackedCoordinateSequence
       boolean hasZM = this.hasZ() & this.hasM();
 
       int skipJ = Math.max(0, dimension - coordDimension);
+      if (skipJ > 0) {
+        if (removeZ) skipJ +=1;
+        else if (addZ && coordMeasures>0) skipJ-=1;
+      }
+
       coords = new double[coordinates.length * this.dimension];
       for (int i = 0, j = 0; i < coordinates.length; i++, j += skipJ) {
         coords[j++] = coordinates[i].x;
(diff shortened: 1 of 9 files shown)


modules/core/src/test/java/org/locationtech/jts/io/gml2/GMLReaderTestCase.java
```diff
@@ -11,8 +11,10 @@
  */
 package org.locationtech.jts.io.gml2;
```
dr-jts: Not sure why this class is part of the PR?

modules/core/src/main/java/org/locationtech/jts/geom/impl/PackedCoordinateSequence.java
```diff
@@ -337,21 +463,18 @@ public Double(int size, int dimension, int measures) {
...
-      }
-      else if (dimension == 3 && measures == 1) {
-    double m = coords[i * dimension + 2];     
-          return new CoordinateXYM(x,y,m);          
-      }
-      else if (dimension == 4 && measures == 1) {
-    double z = coords[i * dimension + 2];
-    double m = coords[i * dimension + 3];
-    return new CoordinateXYZM(x,y,z,m);
+      if (dimension == 2/* && measures == 0*/) {
```
dr-jts: best to remove the commented code

modules/core/src/test/java/org/locationtech/jts/io/gml2/GMLReaderTestCase.java
FObermaier: All `GMLReaderTestCase` tests always fail in IntelliJ IDEA IDE, because the files required are not present. 
Annoying but only accidental part of this PR

PR #390, opened 2019-03-05
Improve Testbuilder styling

dr-jts:


dr-jts:
Adds more controls to TestBuilder Layers styling panels.

PR #392, opened 2019-03-06
Change name to EnvelopeDistance

dr-jts:
Add unit test

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/index/strtree/EnvelopeDistance.java
+++ b/modules/core/src/main/java/org/locationtech/jts/index/strtree/EnvelopeDistance.java
@@ -1,14 +1,25 @@
+/*
+ * Copyright (c) 2019 Martin Davis.
+ *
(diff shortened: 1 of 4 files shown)


modules/core/src/main/java/org/locationtech/jts/index/strtree/EnvelopeDistance.java
```diff
@@ -1,14 +1,25 @@
...
 
 /**
- * Utility functions for working with {@link Envelope}s.
+ * Functions for computing distances between {@link Envelope}s.
  * 
  * @author mdavis
  *
  */
-public class EnvelopeUtil 
+public class EnvelopeDistance
```
jnh5y: @dr-jts Is this an API change?  E.g., do we expect folks to have used EnvelopeUtil in their code?

dr-jts: No., that class was only introduced in #391 yesterday.

PR #393, opened 2019-03-12
Refactor method to change PM

bjornharrtell:
This replaces https://github.com/locationtech/jts/pull/221 and will work in transpilation to JSTS.

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/geom/GeometryFactory.java
+++ b/modules/core/src/main/java/org/locationtech/jts/geom/GeometryFactory.java
@@ -602,12 +602,12 @@ public class GeometryFactory
   public Geometry createGeometry(Geometry g)
   {
     GeometryEditor editor = new GeometryEditor(this);
-    return editor.edit(g, new CloneOp(coordinateSequenceFactory));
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/locationtech/jts/geom/GeometryFactory.java
```diff
@@ -602,11 +602,17 @@ public LineString createLineString(CoordinateSequence coordinates) {
...
     GeometryEditor editor = new GeometryEditor(this);
-    return editor.edit(g, new GeometryEditor.CoordinateSequenceOperation() {
-      public CoordinateSequence edit(CoordinateSequence coordSeq, Geometry geometry) {
-                  return coordinateSequenceFactory.create(coordSeq);
-          }
-    });
+    return editor.edit(g, new CloneOp(coordinateSequenceFactory));
+  }
+
+  private static class CloneOp extends GeometryEditor.CoordinateSequenceOperation {
```
dr-jts: Can this be called CoordSeqCloneOp, just to leave some future naming freedom?

bjornharrtell: Sure, fixed.

PR #394, opened 2019-03-24
Apply fixes to SegmentNodeList from GEOS.

saicheems:
libgeos/geos@c709406

Signed-off-by: Sai Cheemalapati <saicheems@google.com>


dr-jts:
Any idea if there is unit test cases in GEOS which exhibit this as a bug?

dr-jts:
According to the GEOS commit this fix was for a failure in test file buffer.xml. 
 This is now called https://github.com/libgeos/geos/blob/master/tests/xmltester/tests/misc/Buffer-1.xml

JTS runs all those test fine (when the BufferResultMatcher is used).

So I don't see any need for this patch, unless a failing unit test can be shown.

saicheems:
Thanks, I confirmed that the error I was expecting does not occur at HEAD. Closing.

PR #395, opened 2019-03-26
Apply fixes to SegmentNodeList from GEOS.

saicheems:
https://github.com/libgeos/geos/commit/c7094065ad1285a88071654f8b8626cc60022ff2

Signed-off-by: Sai Cheemalapati <saicheems@google.com>

### DIFF
--- a/modules/core/src/test/java/org/locationtech/jts/noding/snapround/SnapRoundingTest.java
+++ b/modules/core/src/test/java/org/locationtech/jts/noding/snapround/SnapRoundingTest.java
@@ -16,10 +16,12 @@ import java.util.ArrayList;
 import java.util.List;
 
 import org.locationtech.jts.geom.Coordinate;
+import org.locationtech.jts.geom.Geometry;
 import org.locationtech.jts.geom.LineSegment;
 import org.locationtech.jts.geom.LineString;
 import org.locationtech.jts.geom.PrecisionModel;
 import org.locationtech.jts.io.WKTReader;
+import org.locationtech.jts.noding.NodedSegmentString;
 
 import junit.framework.TestCase;
 import junit.textui.TestRunner;
@@ -39,6 +41,22 @@ public class SnapRoundingTest  extends TestCase {
   }
 
   public SnapRoundingTest(String name) { super(name); }
+
+  public void testThinTriangle() throws Exception {
+      String wkt = "LINESTRING ( 55121.54481117887 42694.49730855581,"
+          + "55121.54481117887 42694.4973085558,"
+          + "55121.458748617406 42694.419143944244,"
+          + "55121.54481117887 42694.49730855581 )";
+      Geometry g = new WKTReader().read(wkt);
+      List<NodedSegmentString> strings = new ArrayList<>();
+      strings.add(new NodedSegmentString(g.getCoordinates(), null));
+      PrecisionModel pm = new PrecisionModel(1.1131949079327356E11);
+      new MCIndexSnapRounder(pm).computeNodes(strings);
+      strings.get(0).getNodeList().addSplitEdges(strings);
+      for (NodedSegmentString s : strings) {
+        assertTrue(s.size() >= 2);
+    }
+  }
 
   public void testPolyWithCloseNode() {
     String[] polyWithCloseNode = {
--- a/modules/core/src/main/java/org/locationtech/jts/noding/SegmentNodeList.java
+++ b/modules/core/src/main/java/org/locationtech/jts/noding/SegmentNodeList.java
@@ -132,6 +132,11 @@ public class SegmentNodeList
     SegmentNode eiPrev = (SegmentNode) it.next();
     while (it.hasNext()) {
       SegmentNode ei = (SegmentNode) it.next();
+
+      if (ei.compareTo(eiPrev) == 0) {
+        continue;
+      }
+
       boolean isCollapsed = findCollapseIndex(eiPrev, ei, collapsedVertexIndex);
       if (isCollapsed)
         collapsedVertexIndexes.add(new Integer(collapsedVertexIndex[0]));
@@ -226,7 +231,12 @@ public class SegmentNodeList
     // add it to the points list as well.
     // (This check is needed because the distance metric is not totally reliable!)
     // The check for point equality is 2D only - Z values are ignored
-    boolean useIntPt1 = ei1.isInterior() || ! ei1.coord.equals2D(lastSegStartPt);
+
+    // Added check for npts being == 2 as in that case NOT using second point
+    // would mean creating a SegmentString with a single point
+    // FIXME: check with mbdavis about this, ie: is it a bug in the caller ?
+    //
+    boolean useIntPt1 = npts == 2 || (ei1.isInterior() || !ei1.coord.equals2D(lastSegStartPt));
     if (! useIntPt1) {
       npts--;
     }
@@ -277,7 +287,12 @@ public class SegmentNodeList
     // add it to the points list as well.
     // (This check is needed because the distance metric is not totally reliable!)
     // The check for point equality is 2D only - Z values are ignored
-    boolean useIntPt1 = ei1.isInterior() || ! ei1.coord.equals2D(lastSegStartPt);
+
+    // Added check for npts being == 2 as in that case NOT using second point
+    // would mean creating a SegmentString with a single point
+    // FIXME: check with mbdavis about this, ie: is it a bug in the caller ?
+    //
+    boolean useIntPt1 = npts == 2 || (ei1.isInterior() || !ei1.coord.equals2D(lastSegStartPt));
     if (! useIntPt1) {
       npts--;
     }


saicheems:
Sorry about the re-create, it seems I can't reopen my previous PR (#394)...

I had made a mistake, I was able to reproduce at HEAD. I included a test case which fails if the changes made to SegmentNodeList are reverted. The file I placed the test in was chosen somewhat arbitrarily, I'm happy to move it anywhere else that makes sense.

modules/core/src/test/java/org/locationtech/jts/noding/snapround/SnapRoundingTest.java
```diff
@@ -40,6 +42,22 @@ public static void main(String args[]) {
...
 
+  public void testThinTriangle() throws Exception {
+      String wkt = "LINESTRING ( 55121.54481117887 42694.49730855581,"
+          + "55121.54481117887 42694.4973085558,"
+          + "55121.458748617406 42694.419143944244,"
+          + "55121.54481117887 42694.49730855581 )";
+      Geometry g = new WKTReader().read(wkt);
+      List<NodedSegmentString> strings = new ArrayList<>();
+      strings.add(new NodedSegmentString(g.getCoordinates(), null));
+      PrecisionModel pm = new PrecisionModel(1.1131949079327356E11);
```
dr-jts: This is a very odd scale factor for the PrecisionModel.  Is it needed to get this test to run?

dr-jts: And it appears the answer is yes... I tried with 1e11 and 1e12 and both runs passed.

So this is an *extreme* corner case, IMO.  I'll still look at at the patch, but I would like to be sure it's patching the right thing, and won't be masking other potential errors.

That large of a PrecisionModel factor would be unusual, especially since it is greater than the actual precision of the input ordinates.

saicheems: Thanks. For some context in where that PM came from, we chose a precision which is 1e6 * the nominal scale in meters of a projection, e.g. 111319.49079327357 * 1e6, where the former is the nominal scale in meters of EPSG:4326, and the latter a micron scaling factor (snap to a micron in the projection).

saicheems: Here's another failure case with a slightly less crazy precision:

```
String wkt = "LINESTRING (-1677607.6366504875 -588231.47100446,"
    + "-1674050.1010869485 -587435.2186255794,"
    + "-1670493.6527468169 -586636.7948791061,"
    + "-1424286.3681743187 -525586.1397894835,"
    + "-1670493.6527468169 -586636.7948791061,"
    + "-1674050.1010869485 -587435.2186255795,"
    + "-1677607.6366504875 -588231.47100446)";

...
PrecisionModel pm = new PrecisionModel(1.11E10);
```

dr-jts: That looks better... 

Theoretically arbitrary scale factors should work, but they are not really what that concept was designed to handle.

dr-jts: One problem with both of these test cases is that the precision scale factor is greater than the actual precision of the data.  So the snap-rounding is doing nothing except costing cycles and possibly introducing further numerical error. 

Also, using a very large scale factor with relatively large ordinate values is probably pushing the snap-rounding code to a point where it doesn't function correctly.  More specifically, the sum of the magnitude of the scale factor + the magnitude of the data should be at least 3-4 orders less than the precision of floating point (16 digits).  In this case the sum is 10 + 7 = 17 - so that is clearly outside the valid range for snap-rounding.

dr-jts: Again, not saying this fix isn't needed, but I would MUCH prefer a test case which is valid in its own right.

dr-jts: And you should probably reconsider your use of snap-rounding..  Given these examples I would say that using it is potentially worse than useless.  Do you actually see a benefit/need for using it?

eengle: Hi Martin, long time no chat!

What we're after is constructing a noded arrangement of lines. where we want the semantics of snaprounding when nodes are created very near each other. And that's as much as possible ALL we want from snaprounding here, so a precision somewhat higher than the magnitude of the coordinates is a good thing.

If I'm reading this right, and I should stress I may not be, we appear to get a result that is wrong because the logic in Java is wrong. It's no longer wrong in GEOS, although it originally made the same mistake. And so in java we can still see single vertex strings in the output sometimes, which shouldn't happen regardless of rounding.

What do you think about the fix? I had worried a little bit that perhaps another strong expectation, such as half-open assigning of nodes to noded segments, was broken somewhere else, and perhaps this edit to snaprounding just glosses over a deeper problem. But once we found the GEOS fix we thought it was time to consult you.

Kind regards,

- Eric

dr-jts: Thanks for the background, Eric.  

It would certainly be great to have a noding algorithm which didn't change the precision of the input, but just snapped nearby nodes/segments enough to ensure the noding is correct.  (FWIW It looks like FME has [solved](https://www.safe.com/blog/2018/09/fme2018-tolerance-evangelist179/) this problem).

The problem is, the snap-rounding algorithm is really only designed to work when all coordinates are rounded to the tolerance.  And AFAIK it's only going to provide a guarantee of correctness when using a precision which actually causing rounding to take place.  So I don't really know what will happen if it's pushed beyond this point.  

Although do I understand that this is actually working for you (in GEOS at least)?  Given the previous comments about the scale factor and ordinate magnitude, I'm puzzled how any snapping can actually take place?

I wil look at the fix in more detail ASAP.  It seems like it can't hurt at least.

eengle: Well, if you use snaprounding with a precision meaningfully smaller than the input vertex spacing, a consistent noder is about what you get, right? And that side effect of snapping input vertices? It's useful, since having all vertices on the same grid greatly reduces the incidence of rare precision bugs in operations like difference.

As for whether this combination of features is working for us: Yes, over a remarkably nasty set of input datasets, many thousands of users, and a couple of years. Even without the fix. Which is to be expected, since this is still unfixed even many years after GEOS fixed it; I can only think they were just luckier in selecting their random test datasets than JTS was ;-]

Anyway, yes, this whole thing works pretty well, we just want to address this logic bug. In the most appropriate place. If this patch, based on the GEOS patch, isn't the right way, we'll just want to go back to them and do whatever you feel is best here, over there as well, to be good citizens about it.

dr-jts: I see your point.  Agreed, it makes sense to use snap-rounding to prevent too-close nodes which are created during noding.  

I'm a little surprised that the snapping can work, since the "tolerance square" of the grid you are using would seem to be at or below the limit of double-precision.  But maybe I'm missing something, if it works...

I've looked at the code more, and it does seem like it's an appropriate fix.  Although it also seems like it will create "zero-length" segments.  Do you find that?  Do you filter those out in your code?

saicheems: We have not observed the creation of zero-length segments, but it's possible that they are being created and turned into empty `LineString`s later on, and we're just allowing that.

dr-jts: I've had a look at the test situtation in detail.  It's always hard to see what's going on when working at the subatomic level, but it looks like basically like the split edge has a set of split nodes which don't make total sense topologically.  My guess this is because the snap-rounding is being pushed too far.  

So the fix solves this problem, but it's really just kicking the can down the road, since it creates a 2-point segment which has identical endpoints.  This isn't supposed to be coming out of the noder.  But maybe in your later processing this isn't as bad as a 1-point line?  Can you just drop the 1-point SegmentString?

saicheems: We run into other issues later on in our pipeline if 1-point `SegmentString`s get dropped, it would take me a while to track down the exact series of problems. We tried for a while to nail each issue caused by the 1-point string before going to the source.

I think the problem is that the resulting strings don't make a logically consistent graph:

> it creates a 2-point segment which has identical endpoints

I may be misunderstanding, but this is not what I have observed.

With this CL, the first geometry I provided produces the graph:
```
[A, B, A]
[A, B]
[B, C]
[C, A]
```

Without:
```
[A, B, A]
[A]
[B, C]
[C, A]
```

FObermaier:
> (FWIW It looks like FME has solved this problem).

cool!

modules/core/src/test/java/org/locationtech/jts/noding/snapround/SnapRoundingTest.java
dr-jts: @saicheems Agreed, I was wrong about the identical endpoints.  

But... either way, the noding is incorrect (because of the [A-B-A]).  Does this not affect things downstream?

This is a typical kind of robustness issue caused by extremely short linestrings.  It might be interesting to track down further, but really this is why it's better to use snap-rounding with a scale factor that avoids creating such short segments.

dr-jts:
In any case, I agree that this change does fix a potential contract breakage in the SegmentNodeList API, and should not cause harm in cases where the input is valid (i.e. has node coordinates assigned to exactly one segment).

I would like to update the commenting and also refactor the code to make this check once only (by having createSplitEdge and addEdgeCoordinates call the same code).  I will also move the test cases out of the SnapRoundingTest and into a more targetted test class.  

I'll make a branch for this and if it's acceptable then use it instead of this PR.

modules/core/src/test/java/org/locationtech/jts/noding/snapround/SnapRoundingTest.java
eengle: Wait, ABA *is* correct isn't it? If you snapround([A,B,A+r/2], r), I'd expect ABA. Snaprounding doesn't do graph transforms, it yields noded strings. Other JTS tools certainly exist to help cope with and clean up this kind of result, and we do use them to do so.

Now if the input falls into just two pixels, snaprounding shouldn't break connectedness despite working with only AB and BA output edges.  That is, the standalone 'a' with no route to 'bc' is the issue, because it has a different graph reachability than the input.

dr-jts: > Wait, ABA _is_ correct isn't it? If you snapround([A,B,A+r/2], r), I'd expect ABA. 

Well, A-B-A is not fully noded, and the noder contract is produce a properly noded set of edges.  (Fully noded would be one edge [A,B]).

The reason is that the SegmentNodeList is invalid - the vertex (55121.54481117887, 42694.4973085558) is assigned to both segment 0 and segment 1.  This is probably due to the extreme short length of segment 0 and some rounding error somewhere.

Anyway, I still stand by my contention that you are pushing the SnapRounding beyond it's operating envelope. But if your process is working that's great.  I'm working on committing the fix now, since it is benign.

dr-jts:
Here's the refactored fix:  https://github.com/dr-jts/jts/tree/fix-segmentnodelist

If you are able to test it that would be great.

modules/core/src/test/java/org/locationtech/jts/noding/snapround/SnapRoundingTest.java
eengle: Well, independently of the fix, let me see if I correctly understand what you're saying:

The snaprounder produces a distinct undirected graph, where 'distinct' means no two edges have internal crossings or are equal, and 'undirected' has the usual graph meaning of adding AB having exactly the same effect as adding BA.

Assuming I have this right, I have some problems with this definition. In particular, a set of distinct undirected edges contains less information about the input's connectivity than a directed graph with duplicate and duplicate reversed edges.

That extra info is very useful because it preserves connectivity. When cleaning a set of noded strings to reach some closed definition of a topology, you really can't function without this info, not without much greater supplementary complexity anyway.

Snaprounding as implemented *seems* to do things with an ~directed model, and clearly doesn't discard reversed duplicates. That might not qualify as a fully noded arrangement, per the definition you are thinking of, but do you mind if I continue to get that anyway, as it's quite useful?

dr-jts: My apology... it's been a while since I've been thinking about this code.  You are correct - the SnapRounder (and JTS Noders) in general DO operate in the way you describe.  They maintain the direction of edges and split edges, and thus can produce duplicates.  And they can produce ABA edges, due to dimensional collapse.  (In Overlay for example these are filtered out after noding has taken place, for example via [this test](https://github.com/dr-jts/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/geomgraph/Edge.java#L118).)

I think maybe we're both right, in a way.   That is in fact valid output from the Snap-rounded noding.  But also the reason that output is happening is due to a robustness issue which is producing an incorrect assignment of split nodes to the edge (as I mentioned previously).  I am going to look into why this happens - although there is probably not a simple fix (which is why snap-rounding was introduced in the first place, because with a reasonable precision tolerance this kind of inconsistency does not happen).

dr-jts:
@FObermaier Yes, very clever of them.  I'm hoping to understand how they do this at some point...

modules/core/src/test/java/org/locationtech/jts/noding/snapround/SnapRoundingTest.java
dr-jts: I think I'm going to have to eat _all_ my words...   This situation does seem to be an expected output of the SnapRounder/SegmentNodeList code, given the input provided (in particular, the very large precision scale factor).  

So the fix is not only benign but required.  I will make a PR with the amended code.

I'm still uncomfortable with pushing the snap-rounder so far into the red, for several reaons:
* it certainly wasn't designed to operate in this regime
* It has code which is likely to be non-robust when working with such large numbers (e.g. the test for [intersecting a HotPixel](https://github.com/dr-jts/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/noding/snapround/HotPixel.java#L220))
* there's probably a more efficient implementation possible for this kind of snapping

That said, if this does actually work then it offers a very nice capability of robust noding with no rounding of coordinates.

dr-jts:
I've created PR #399 to replace this.  If you feel it is correct feel free to close this one.

modules/core/src/test/java/org/locationtech/jts/noding/snapround/SnapRoundingTest.java
dr-jts: I retract my retraction...  The SegmentNodeList **is** designed to fully node, the input linework.  In particular it detects the ABA case and adds a node at B in order to **not** return an ABA edge.  This is what[ `addCollapsedNodes()`](https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/noding/SegmentNodeList.java#L89) does.

The reason there is an ABA edge appearing in the output is because the sort order of the nodes in the SegmentNodeList is not sensitive to the interior/exterior distinction of the nodes.  The A node on segment 1 is in the interior of the segment and should come **after** the B node that is the start point of segment 1.  But instead it sorts before it (for some reason), and thus causes the ABA edge to be created - followed immediately by another AB edge when the B node is finally encountered.

Not sure what the fix is at this point.

dr-jts: And the reason the sort order is wrong is because the SegmentPointComparator is relying on the octant of the segment, and assumes that both points lie "close enough" to the segment to be compared in a simple way.  The A point violates this assumption, because it doesn't really "lie along" the segment it is assigned to - or at least can't robustly be sorted using the simple sign comparison method.

A fix might be to simply force interior nodes to sort after the start node of the segment, which is exterior.  Or perhaps just check if one of the nodes is the start point, and always sort that first.

saicheems:
Thanks Martin! Closing this PR.

PR #398, opened 2019-03-27
Force StringUtil.toString(double) method to use Locale.ENGLISH

FObermaier:
Currently `StringUtil.toString(double)` function uses `Locale.getDefault()` to write a `double` to a `String`.
This function is being used from within `CoordinateSequences.toString(CoordinateSequence)` making the output un[read|use]able in environments that have a comma as a decimal seperator (like e.g. german).

This PR contains
* Fix for `StringUtil.toString(double)` method to use `Locale.ENGLISH`
* Add methods to change decimal format pattern
* Add unit test

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/util/StringUtil.java
+++ b/modules/core/src/main/java/org/locationtech/jts/util/StringUtil.java
@@ -17,9 +17,12 @@ import java.io.IOException;
 import java.io.LineNumberReader;
 import java.io.PrintStream;
 import java.io.StringReader;
+import java.math.RoundingMode;
 import java.text.DecimalFormat;
+import java.text.DecimalFormatSymbols;
 import java.text.NumberFormat;
 import java.util.ArrayList;
+import java.util.Locale;
 
 /**
  * Utility methods for working with {@link String}s.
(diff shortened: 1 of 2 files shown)


dr-jts:
Seems like a good fix.  Does this problem affect WKTWriter as well?

FObermaier:
No, `WKTWriter` uses its own `DecimalFormatter`. It is created like this:
https://github.com/locationtech/jts/blob/5ece77b631735452ac442f737124a32695f434f2/modules/core/src/main/java/org/locationtech/jts/io/WKTWriter.java#L152-L162

dr-jts:
Yes, I'm just wondering why the two approaches are so different.

FObermaier:
The only real difference I see is that the `WKTWriter` sets the number of decimal places according to a `PrecisionModel`. `StringUtil.toString(double)` lacks that, instead it always used to print out at most one decimal place ("0.#"). Now that is changable.

`WKTWriter` could reuse a static `DecimalFormatSymbols` instance.
`StringUtil` could adjust its `DecimalFormatSymbols` like `WKTWriter` does

dsmiley:
I recommend Locale.ROOT.  In Lucene/Solr we use https://github.com/policeman-tools/forbidden-apis to catch such issues at compile time in a comprehensive way.

FObermaier:
`Locale.ROOT` is fine for me, seems to be what resembles `CultureInfo.InvariantInfo` in the dotNet world most.

dr-jts:
I agree 100% that number formating needs to be done more consistently in JTS.  I just ran across [another place](https://github.com/locationtech/jts/blob/master/modules/app/src/main/java/org/locationtech/jtstest/testbuilder/geom/GeometryLocation.java#L155) where the formatting is ad hoc and hence wrong for some numbers.

But this should be done in a better place than `StringUtil` I think.  It's probably better off in NumberUtil (and the StringUtil class can delegate to that).

It might make sense for WKTWriter to then use this code as well (which will require a method to set the maximum digits as well).

dr-jts:
Is Locale.ROOT guarenteed to use a period as the decimal separator?  Wouldn't it be clearer to specify the decimal separator explicitly?

dr-jts:
Is it really necessary to be able to set the rounding mode?

dr-jts:
Hmm.. and apparently DecimalFormat is not thread-safe, which complicates this further.   This hasn't been a problem so far to my knowledge, but this kind of issue tends to be subtle.  

Perhaps the `toString(double)` method should be synchronized to avoid potential issues (in spite of the potential performance hit).

dr-jts:
This is also a bug in the `WKTWriter` static toXXX methods as well.

FObermaier:
@dr-jts, I'm fine totally fine with you creating your own PR

dr-jts:
> @dr-jts, I'm fine totally fine with you creating your own PR

Great - working on it now.

dr-jts:
@FObermaier have a look at #409 as an alternative and extension to this PR

dr-jts:
Fixed by #409

PR #403, opened 2019-04-01
Add AnchorPoint noding to jtslab

FObermaier:
An `AnchorPoint` is a coordinate that has already has been determined and is known to the process. It is kept in a `KdTree` and retrieved and reused during further computation.

This PR adds `AnchorPoint` based
* noding and 
* intersection computation

to the jts-lab module.

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/algorithm/LineIntersector.java
+++ b/modules/core/src/main/java/org/locationtech/jts/algorithm/LineIntersector.java
@@ -16,6 +16,8 @@ package org.locationtech.jts.algorithm;
  */
 import org.locationtech.jts.geom.Coordinate;
 import org.locationtech.jts.geom.PrecisionModel;
+import org.locationtech.jts.geomgraph.index.EdgeSetIntersector;
+import org.locationtech.jts.geomgraph.index.SimpleMCSweepLineIntersector;
 import org.locationtech.jts.io.WKTWriter;
 import org.locationtech.jts.util.Assert;
 
@@ -165,6 +167,15 @@ public abstract class LineIntersector
     pa = intPt[0];
     pb = intPt[1];
     result = 0;
+  }
+
+  /**
+   * Create an EdgeSetIntersector
+   *
+   * @return an EdgeSetIntersector
+   */
+  public EdgeSetIntersector create() {
+    return new SimpleMCSweepLineIntersector();
   }
 
   /**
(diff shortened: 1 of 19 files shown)


dr-jts:
There's a lot of changes to the core classes in this PR, so I don't want to merge it as is.

dr-jts:
I am curious about the algorithm though.  Is this based on some other codebase?  Does it solve some or all robustness problems?

Is this similar to what FME has [implemented](https://www.safe.com/blog/2018/09/fme2018-tolerance-evangelist179/)?

FObermaier:
>Is this based on some other codebase?

No. I developed it after you cited the above FME blog post on gitter.im with the hope to eliminate all robustness problems...

> Does it solve some or all robustness problems?

It does solve some but afaikt not all. A good example of where it works better than snap-rounding is modules/lab/src/test/java/org/locationtech/jtslab/algorithm/AnchorPointRobustLineIntersectorTest.java

FObermaier:
> There's a lot of changes to the core classes in this PR, so I don't want to merge it as is.

Most of these changes are needed to inject `AnchorPointRobustLineIntersector` instead of std. `RobustLineIntersector`.

dr-jts:
> Most of these changes are needed to inject `AnchorPointRobustLineIntersector` instead of std. `RobustLineIntersector`.

Understood.  It might be possible to do the snapping outside `RobustLineIntersector`.  Fewer changes is better (although some minor ones is ok).

Also, the new OverlayNG framework makes it a lot easier to switch in a new `Noder`.  So that should reduce the amount of injection changes required as well.

Changes to core JTS should be done as a separate PR.  Lab PRs should only have lab code inside them.  Otherwise it's hard to know what is experimental.

dr-jts:
As it turns out I was working on a similar idea, so got inspired to finish it off.  It works really well - it seems to be very robust, and is often faster than the original overlay (since the snapping heuristic used there is quite slow).  

The code is called `SnappingNoder` (for now).  It simply plugs into the OverlayNG framework, which makes it easy to try out.  OverlayNG is key here, since it handles topology collapse better.

If you are interested in trying it out, it's in a branch in my GH account.  Entry point is [here](https://github.com/dr-jts/jts/blob/overlay-sr/modules/app/src/main/java/org/locationtech/jtstest/function/OverlayNGSnappingFunctions.java).  I'm keen to get some testing done on this code.

FObermaier:
Superceeded with snap-rounding

PR #406, opened 2019-04-04
Fix LinearLocation::isEndpoint for not-normalized locations [jts-core] [bug]

piomar123:
The second condition was never actually evaluated:
https://github.com/locationtech/jts/blob/67fcc3d6f9b20cb3eecd29560f6fe9e5061804e8/modules/core/src/main/java/org/locationtech/jts/linearref/LinearLocation.java#L422-L423

Now if the location refers to the second to last `segmentIndex` but with `segmentFraction >= 1.0` it's also considered endpoint.

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/linearref/LinearLocation.java
+++ b/modules/core/src/main/java/org/locationtech/jts/linearref/LinearLocation.java
@@ -444,7 +444,7 @@ public class LinearLocation
     int nseg = lineComp.getNumPoints() - 1;
     // if not an endpoint can be returned directly
     if (segmentIndex < nseg) return this;
-    return new LinearLocation(componentIndex, nseg, 1.0, false);
+    return new LinearLocation(componentIndex, nseg - 1, 1.0, false);
   }
   
   /**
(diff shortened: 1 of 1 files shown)


dr-jts:
How did you trigger this bug?

dr-jts:
There's probably another bug of the same type in `LinearLocation.toLowest` [here](https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/linearref/LinearLocation.java#L447).  So would be best to fix them both together.  I can make a PR for this.

piomar123:
I've been just looking for a method checking whether `LinearLocation` is between startpoint and endpoint, and found this bug in the code. 
I can add the fix in `LinearLocation.toLowest` too.

piomar123:
The test for checking new implementation should be also added.

dr-jts:
> Existing tests are enough, just `LinearLocation.getSegment` was lenient enough to pass all of them with incorrect input.

Not following - I didn't see any failures in the current tests?

piomar123:
I mean `LinearLocation.getSegment` is implemented defensively that if `segmentIndex` is too big, it takes last segment anyway:
https://github.com/locationtech/jts/blob/f56998b09f9faa1dbf9ec8acebdcef6ff1ef1894/modules/core/src/main/java/org/locationtech/jts/linearref/LinearLocation.java#L283-L285

That's why I thought there are insufficient tests for that but it just worked as expected before:
https://github.com/locationtech/jts/blob/f56998b09f9faa1dbf9ec8acebdcef6ff1ef1894/modules/core/src/test/java/org/locationtech/jts/linearref/AbstractIndexedLineTest.java#L141-L149

dr-jts:
Well, it would be good to have some kind of unit test that covers these bugs.

dr-jts:
Replaced by #421

PR #409, opened 2019-04-09
Consistent ordinate formatting

dr-jts:
Encapsulate ordinate formatting for consistency and accuracy.

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/io/OrdinateFormat.java
+++ b/modules/core/src/main/java/org/locationtech/jts/io/OrdinateFormat.java
@@ -1,0 +1,122 @@
+/*
+ * Copyright (c) 2019 Martin Davis.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Eclipse Distribution License v. 1.0 which accompanies this distribution.
+ * The Eclipse Public License is available at http://www.eclipse.org/legal/epl-v10.html
+ * and the Eclipse Distribution License is available at
+ *
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ */
+
+package org.locationtech.jts.io;
+
+import java.text.DecimalFormat;
+import java.text.DecimalFormatSymbols;
+
+/**
+ * Formats numeric values for ordinates
+ * in a consistent, accurate way.
+ * <p>
+ * The format has the following characteristics:
+ * <ul>
+ * <li>It is consistent in all locales (in particular, the decimal separator is always a period)
+ * <li>Scientific notation is never output, even for very large numbers.
+ * This means that it is possible that output can contain a large number of digits. 
+ * <li>The maximum number of decimal places reflects the available precision
+ * <li>NaN values are represented as "NaN"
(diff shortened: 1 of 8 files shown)


dr-jts:
Replaces #398

FObermaier:
Fine.

modules/core/src/main/java/org/locationtech/jts/io/OrdinateFormat.java
```diff
@@ -0,0 +1,64 @@
...
+  /**
+   * The number of fraction digits to accommodate a reasonable ordinat value
+   */
+  public static final int MAX_FRACTION_DIGITS = 18;
+  
+  public static OrdinateFormat DEFAULT = new OrdinateFormat();
+
+  private DecimalFormat format;
+  
+  public static DecimalFormat createFormat(int maximumFractionDigits) {
```
FObermaier: private?

modules/core/src/main/java/org/locationtech/jts/io/WKTWriter.java
```diff
@@ -150,15 +159,7 @@ public static String toLineString(Coordinate p0, Coordinate p1)
    *      s without scientific notation.
    */
   private static DecimalFormat createFormatter(PrecisionModel precisionModel) {
```
FObermaier: return `OrdinateFormat`?

modules/core/src/main/java/org/locationtech/jts/io/OrdinateFormat.java
```diff
@@ -0,0 +1,64 @@
...
+    format.setMaximumFractionDigits(maximumFractionDigits);
+    return format;
+  }
+
+  public OrdinateFormat() {
+    format = createFormat(MAX_FRACTION_DIGITS);
+  }
+
+  public OrdinateFormat(int maximumFractionDigits) {
+    format = createFormat(maximumFractionDigits);
```
FObermaier: `format = createFormat(Math.min(MAX_FRACTION_DIGITS, maximumFractionDigits));`

dr-jts: I guess could do this, but the current code leaves the option open to force more digits if desired.

modules/core/src/main/java/org/locationtech/jts/io/WKTWriter.java
dr-jts: That would be better... It would just require a lot more changes throughout WKTWriter.  I will consider...

modules/core/src/main/java/org/locationtech/jts/io/OrdinateFormat.java
dr-jts: currently has to be `public` because it's used by `WKTWriter`.  But your suggested change below would let this become `private`.

modules/core/src/main/java/org/locationtech/jts/io/WKTWriter.java
dr-jts: I have refactored `WKTWriter` to use `OrdinateFormat`.  This removes the dependency on `DecimalFormat`.

PR #421, opened 2019-04-23
Fix LinearLocation endpoint segment index bugs

dr-jts:


dr-jts:
Replaces #406, adding unit tests.

Issue #420, opened 2019-04-23
Perf: improve UniqueCoordinateArrayFilter


jandam:
UniqueCoordinateArrayFilter can be improved by using HashSet instead of TreeSet and by removing treeSet.contains
*) HashSet is faster (up to 2x) than TreeSet in most cases
*) Set::contains + Set::add can be replaced with single Set:add that return FALSE when value is already in Set.
Proposed fix:
```
public class UniqueCoordinateArrayFilter implements CoordinateFilter 
{
  /**
   * Convenience method which allows running the filter over an array of {@link Coordinate}s.
   * 
   * @param coords an array of coordinates
   * @return an array of the unique coordinates
   */
  public static Coordinate[] filterCoordinates(Coordinate[] coords)
  {
    UniqueCoordinateArrayFilter filter = new UniqueCoordinateArrayFilter();
    for (int i = 0; i < coords.length; i++) {
      filter.filter(coords[i]);
    }
    return filter.getCoordinates();
  }
  
  Set set = new HashSet();  // changed here
  List list = new ArrayList();

  public UniqueCoordinateArrayFilter() { }

  /**
   *  Returns the gathered <code>Coordinate</code>s.
   *
   *@return    the <code>Coordinate</code>s collected by this <code>CoordinateArrayFilter</code>
   */
  public Coordinate[] getCoordinates() {
    Coordinate[] coordinates = new Coordinate[list.size()];
    return (Coordinate[]) list.toArray(coordinates);
  }

  public void filter(Coordinate coord) {
    if (set.add(coord)) {  // changed here, removed exlamation
      list.add(coord);
    }
  }
}

```

dr-jts:
Curious to know how you identified this fix.  Did you notice a performance issue with computing Convex Hulls?  Or was this flagged by an automated code style checker?

One reason for this question is that I notice a couple of potential further performance improvements:

* Use a `CoordinateSequenceFilter` instead of a `CoordinateFilter`
* Technically it is not necessary to include the vertices in polygon holes, since they should always be contained in the convex hull of the polygon shell (if it is valid).

dr-jts:
> One reason for this question is that I notice a couple of potential further performance improvements:
> 
> * Use a `CoordinateSequenceFilter` instead of a `CoordinateFilter`
> * Technically it is not necessary to include the vertices in polygon holes, since they should always be contained in the convex hull of the polygon shell (if it is valid).

@dbaston Any thoughts on these suggestions?

[...]

PR #422, opened 2019-04-25
Improve performance of UniqueCoordinateFilter

dr-jts:
Use HashSet, and boolean return value from Set.add().
Javadoc improvements

Based on suggestions in #420.

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/util/UniqueCoordinateArrayFilter.java
+++ b/modules/core/src/main/java/org/locationtech/jts/util/UniqueCoordinateArrayFilter.java
@@ -14,15 +14,17 @@
 package org.locationtech.jts.util;
 
 import java.util.ArrayList;
-import java.util.TreeSet;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
 
 import org.locationtech.jts.geom.Coordinate;
 import org.locationtech.jts.geom.CoordinateFilter;
 
 
 /**
- *  A {@link CoordinateFilter} that builds a set of <code>Coordinate</code>s.
- *  The set of coordinates contains no duplicate points.
+ *  A {@link CoordinateFilter} that extracts a unique array of <code>Coordinate</code>s.
+ *  The array of coordinates contains no duplicate points.
  *  It preserves the order of the input points.
  *
  *@version 1.7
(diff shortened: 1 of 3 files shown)


modules/core/src/main/java/org/locationtech/jts/util/UniqueCoordinateArrayFilter.java
```diff
@@ -44,8 +46,9 @@
     return filter.getCoordinates();
   }
   
-  TreeSet treeSet = new TreeSet();
-  ArrayList list = new ArrayList();
+  private Set<Coordinate> coordSet = new HashSet<Coordinate>();
+  // Use an auxiliary list as well in order to preserve coordinate order
```
dbaston: I wonder about `LinkedHashSet` here?

dr-jts: Interesting suggestion.  But I suspect that using an `ArrayList` is more efficient, since it is simpler than a doubly linked list?

`LinkedHashSet` is probably better for situations where the ordering needs to be maintained dymamically, whereas in this use case requires only a single access to the ordered points.

dbaston: Yeah, I can imagine the `ArrayList` coming out ahead too. Hard to say without measuring. Just thought I'd mention it in case it hadn't occurred to you.

dr-jts: Thanks...  good to be reminded of the ever-increasing plethora of useful Java data structures.

dbaston: You're right that the `ArrayList` is faster. Filtering a million-coordinate array ten times takes 12.1s in master, 2.5s with this PR, and 2.8s with the `LinkedHashSet`.

dr-jts: Nice!  The big win of course is the improvement over master.  But the extra few cycles saved with `ArrayLIst` are nice too.

PR #438, opened 2019-06-07
FIx deprecation warnings on Coordinate.z

dr-jts:
FIx deprecation warnings on Coordinate.z
Field is used by subclasses.
Change deprecation to discouragement.


snodnipper:
I have just been updating some code and came across this.  I was wondering whether you considered deprecating direct field access to X and Y too.

dr-jts:
Well, this PR was to **remove** the deprecation on direct Z access., for the time being.  `x` and `y` are already non-deprecated, so no change needed.

But it would be worth adding a warning that their use is discouraged.

PR #442, opened 2019-06-18
Remove unused variables

bjornharrtell:


bjornharrtell:
This actually (seemingly) caused problems in the transpiled JavaScript port, see https://github.com/bjornharrtell/jsts/pull/384.

PR #448, opened 2019-07-16
Fix Voronoi polygon generation by adding robust Triangle circumcentreDD

dr-jts:
Also:
* Add DD.determinant methods


dr-jts:
See also [GEOS 859](https://trac.osgeo.org/geos/ticket/859#comment:4), which is another example of the gridded input issue.

PR #451, opened 2019-07-18
Add CLI for running operations

dr-jts:
Add CLIs for running operations

### DIFF
--- a/bin/testbuilderop.bat
+++ b/bin/testbuilderop.bat
@@ -1,0 +1,9 @@
+@echo off
+rem Windows batch file to run the JTS TestBuilderOpCmd from the Maven target directory
+
+set JAVA_OPTS=-Xms256M -Xmx1024M
+
+REM Default app options can be added here
+set APP_OPTS=%*
+
+java %JAVA_OPTS% -cp %~dp0..\modules\app\target\JTSTestBuilder.jar org.locationtech.jtstest.testbuilder.cmd.TestBuilderOpCmd %APP_OPTS% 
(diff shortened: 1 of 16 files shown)


jnh5y:
@dr-jts This looks like fun, so I wanted to try it out.  

I think the testbuilderop.sh needs to reference JTSTestBuilder.jar instead of JTSTestRunner.jar.
(E.g., the last line should be...)
java -cp ${JTS_LIB_DIR}/JTSTestBuilder.jar org.locationtech.jtstest.testbuilder.cmd.TestBuilderOpCmd $@

Separate from that, can you give a quick example of using the new command line tool?

dr-jts:
Right, that reference to JTSTestRunner.jar is a copy-paste bug.  I'll fix it very shortly

dr-jts:
An example of running is:
```
testrunnop -op Overly.unaryUnion -afile some-file-with-geom.wkt -output wkb
```
That reads the WKT file (containing one or more geometries), executes the Overlay.unaryUnion operation (which is just a synonym for Geometry.union() in this case), and outputs the result as WKB.

dr-jts:
One slightly confusing thing is that the -op arg has to be Category.function, which are listed in the TestBuilder Spatial Function panel tree view.  

In particular, only *some* Geometry methods are actually given under the Geometry category. So that might be bit confusing.  I think I'll add all the Geometry methods there, so that it's more obvious. I will also add a dump of the full list of function names that are available (triggered by a -help argument).

Thx for testing this out - let me know how it goes.

dr-jts:
Updated to fix issue and add operations listing

PR #460, opened 2019-08-15
use mvn site to generate javadoc jar

jodygarnett:
Move javadoc to phase site:

* mvn clean install - just builds jar and source jar
* mvn site - builds javadoc jar

Note the javadoc generation has warnings and errors at present...

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -18,6 +18,9 @@
     Build everything and skip tests:
        mvn clean install -DskipTests
 
+    Build javadocs (shows up as a jar):
+       mvn site
+       
     Setup for eclipse development:
        mvn eclipse:eclipse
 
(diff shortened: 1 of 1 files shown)


jodygarnett:
@jnh5y this PR is "functional", not sure if we want it?  I believe we addressed the javadoc warning issues discussed above.

jnh5y:
@jodygarnett why do you want to leave the generation of the javadocs to mvn site?  I'm confused.

jodygarnett:
I think this is left over from when we first migrated to maven. There are two options for where to hook in javadocs... 

* We could set it up as a "report" so it would show up in `mvn site` build
* We could bind it to a stage (so for example it would show up in a full `mvn install` build)

Or we could do nothing and depend on `mvn javadoc:javadoc` like we do now?

I recommend we close this pull request, unless you are used to (or like the sound of) the `mvn site` approach?

Issue #19, opened 2016-02-02
IndexedPointInAreaLocator Goes Into Infinite Loop with Empty Geometry


airbreather:
This is [issue 51](https://sourceforge.net/p/jts-topo-suite/bugs/51/) on SourceForge.

The problem is [here](https://github.com/dr-jts/jts/blob/8503255634783684b7fda46798c8507f760c3d26/modules/core/src/main/java/org/locationtech/jts/index/intervalrtree/SortedPackedIntervalRTree.java#L91): we only check to see if the count is 1, but for an empty list of nodes, this will always be 0.

I had a patch over at SourceForge, but looking at it again, there's a more elegant solution that doesn't penalize the common case as much:
1. In [org.locationtech.jts.index.intervalrtree](https://github.com/dr-jts/jts/tree/8503255634783684b7fda46798c8507f760c3d26/modules/core/src/main/java/org/locationtech/jts/index/intervalrtree), add another subclass of `IntervalRTreeNode` called something like `IntervalRTreeNullNode` whose `query` method is a no-op.
2. Right at the [top](https://github.com/dr-jts/jts/blob/8503255634783684b7fda46798c8507f760c3d26/modules/core/src/main/java/org/locationtech/jts/index/intervalrtree/SortedPackedIntervalRTree.java#L81) of `buildRoot`, check to see if `leaves` has a `size()` of 0.  If it does, return a new instance of `IntervalRTreeNullNode`.

I signed a CLA, but I think it might be faster if I just point these things out, since I'm having difficulties setting up my local environment to do development **of** JTS.

dr-jts:
A unit test for this is to add the following to `IndexedPointInAreaLocatorTest`:

```
   public void testEmpty() throws Exception {
      runPtInRing(Location.EXTERIOR, new Coordinate(0,0), "POLYGON EMPTY");
   }
```

dr-jts:
Closed by #462

PR #462, opened 2019-08-20
Fix IndexedPointInAreaLocator and SortedPackedIntervalRTree to handle empty input

dr-jts:
`SortedPackedIntervalRTree` had a bug causing an infinite loop when building an empty tree.  This has been fixed.  
`IndexedPointInAreaLocator`depended on this, but has been fixed internally as well.

This also fixes `PreparedPolygon`.

See issue #19 


dr-jts:
Closes #19

PR #470, opened 2019-09-05
Add CascadedPolygonUnion bufferUnion

dr-jts:
Improves CascadedPolygonUnion by using a buffer(0) union if overlay union fails due to a topology error.


dbaston:
Is there a test case that throws an exception with cascaded union but goes on to succeed with the buffer union?

dr-jts:
> Is there a test case that throws an exception with cascaded union but goes on to succeed with the buffer union?

good point, I was thinking about this too.  There's definitely quite a few large datasets which fail without this and pass with it.  The trick is extracting a simple test case. I will work on this.  

In fact my hypothesis is that most union failures will actually work with buffer(0), so any of them would provide a test case.

dr-jts:
I am adding a test case for this in #471.

No guarantee it has the same failing behaviour in GEOS, unfortunately.  But hopefully it does.

PR #489, opened 2019-10-24
Enforce polygon ring orientation with GeometryFactory

FObermaier:
JTS currently does not rely on/enforce a specific polygon ring orientation.
The OGC Simple Features - Common Architecture v1.2.1 spec (6.1.11.1) specifies
that a polygons exterior ring should be CCW, whereas inner rings should be CW.
Unfortunately, for JTS it is the other way around.

Another reason for this enhancement is, that there are data sources that require
a specific ring orientation for polygons. Oracle, SqlServer (at least for geography)
require CCW, GeoJSON requires CW.

NetTopologySuite/NetTopologySuite#360

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/geom/GeometryFactory.java
+++ b/modules/core/src/main/java/org/locationtech/jts/geom/GeometryFactory.java
@@ -17,6 +17,7 @@ import java.io.Serializable;
 import java.util.Collection;
 import java.util.Iterator;
 
+import org.locationtech.jts.algorithm.Orientation;
 import org.locationtech.jts.geom.impl.CoordinateArraySequenceFactory;
 import org.locationtech.jts.geom.util.GeometryEditor;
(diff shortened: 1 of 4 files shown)


dr-jts:
I do not think that JTS should enforce ring orientation in the constructor.  It's a deliberate design choice to provide a lot of latitude in the geometry structures that can be represented.  One reason for this is to allow representing invalid geometry in order to allow it to be manipulated or rectified.  Also, see [Postel's Law](https://en.wikipedia.org/wiki/Robustness_principle).  

However, JTS does choose a canonical ring orientation for constructed geometry (also as per Postel's Law).  Unfortunately the orientation was chosen prior to being specified in the OGC standard, and the standard happened to choose the opposite orientation.  Ideally this should be rectified by changing the standard output orientation (to CCW for shells, CW for rings).  This can probably be done with few consequences, but it might be nice to control this via a global flag, to provide backwards compatibility where needed.

Issue #305, opened 2018-08-28
Geometry.reverse inconsistency


FObermaier:
The implementations of `Geometry.reverse` can be categorized by ones using `Geometry.copy` and others that don't:

* Reverse using `copy`
  * Point https://github.com/locationtech/jts/blob/8d433caaf37d0dd965af95608c23acbb9dbdb566/modules/core/src/main/java/org/locationtech/jts/geom/Point.java#L193-L196
  * Polygon https://github.com/locationtech/jts/blob/8d433caaf37d0dd965af95608c23acbb9dbdb566/modules/core/src/main/java/org/locationtech/jts/geom/Polygon.java#L432-L440

* Reverse not using `copy`
  * LineString https://github.com/locationtech/jts/blob/8d433caaf37d0dd965af95608c23acbb9dbdb566/modules/core/src/main/java/org/locationtech/jts/geom/LineString.java#L183-L189
  * LinearRing https://github.com/locationtech/jts/blob/8d433caaf37d0dd965af95608c23acbb9dbdb566/modules/core/src/main/java/org/locationtech/jts/geom/LinearRing.java#L135-L141
  * MultiPoint -> GeometryCollection
  * MultiLineString https://github.com/locationtech/jts/blob/8d433caaf37d0dd965af95608c23acbb9dbdb566/modules/core/src/main/java/org/locationtech/jts/geom/MultiPolygon.java#L123-L131
  * MultiPolygon https://github.com/locationtech/jts/blob/8d433caaf37d0dd965af95608c23acbb9dbdb566/modules/core/src/main/java/org/locationtech/jts/geom/MultiPolygon.java#L123-L131
  * GeometryCollection https://github.com/locationtech/jts/blob/8d433caaf37d0dd965af95608c23acbb9dbdb566/modules/core/src/main/java/org/locationtech/jts/geom/GeometryCollection.java#L281-L289

This has the effect that `Geometry.SRID` and `Geometry.UserData` are copied for `Polygon` and `Point` but not for the others. 

This is inconsistent. I have no preference which way is better.

dr-jts:
It's inefficient to use `copy `in most cases, because that copies the internal coordinate sequences which are going to be replaced anyway.

But the SRID should be copied.  Not sure about the userData, since that may or may not be desired depending on use case.

Probably there should be a protected copySRID method which can be called by Geometry methods which should preserve it.

dr-jts:
[...]

Issue #397, opened 2019-03-27
Add .editorconfig file


FObermaier:
The JTS codebase has code styling inconsistencies, e.g. mixed space/tab usage.

> EditorConfig helps maintain consistent coding styles for multiple developers working on the same project across various editors and IDEs. The EditorConfig project consists of a file format for defining coding styles and a collection of text editor plugins that enable editors to read the file format and adhere to defined styles. EditorConfig files are easily readable and they work nicely with version control systems.

https://editorconfig.org/

PR #513, opened 2020-01-23
Make Geometry.reverse() consistent

FObermaier:
* Add protected reverseInternal function to Geometry,
  implement in Point, LineString, LinearRing, Polygon and
  GeometryCollection
* Remove abstract from Geometry.reverse function,
  implementation calls reverseInternal and copies
  Envelope and SRID values
* To avoid breaking change make Geometry.reverse overrides
  return super.reverse()
* Add unit test

Note:
The reversal of items in MultiLineString has been removed.

closes to #305

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/geom/Geometry.java
+++ b/modules/core/src/main/java/org/locationtech/jts/geom/Geometry.java
@@ -49,7 +49,7 @@ import org.locationtech.jts.util.Assert;
  *
  * <H3>Overlay Methods</H3>
  *
- * The overlay methods 
+ * The overlay methods
  * return the most specific class possible to represent the result. If the
  * result is homogeneous, a <code>Point</code>, <code>LineString</code>, or
  * <code>Polygon</code> will be returned if the result contains a single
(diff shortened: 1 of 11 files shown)


dr-jts:
Thanks, code looks good. 

It's unfortunate that the commits are cluttered up with formatting changes (mainly whitespace removal).  Not sure if it's possible to avoid this?

FObermaier:
IntelliJ did that. I must admit I like it. See #397.

dr-jts:
Yes, I like consistent formatting as well (have been working with Go lately, and it's opinionated formatting works very well).  

It's just not ideal to mix formatting changes and code changes, since the code changes get swamped by all the trivial format changes.

FObermaier:
For the review you might want to try 
grafik

I'm not going to re-add trailing white-space to get this accepted. Take what you like.

dr-jts:
Oh, great, didn't know that was available.  That helps with reviewing for sure.

dr-jts:
@FObermaier the deprecations on the subclass `reverse` methods are causing problems, since they shadow the `Geometry.reverse` method.  Isn't it sufficient to just remove the subclass methods?  Or does that break something?

FObermaier:
I'm no java expert. I wanted to avoid introducing a breaking change. I was not aware that this causes other issues.

dr-jts:
@FObermaier understood. This is a trickier design situation than it first appears.

PR #533, opened 2020-04-27
Fixed MinimumBoundingCircle.getMaximumDiameter() to handle case where extremalPts[2]-[0] is longest

tink-expo:
Hello,

It seemed that ```MinimumBoundingCircle.getMaximumDiameter()``` returns farther pair only among ```extremalPts[0]-[1]``` and ```extremalPts[1]-[2]``` if ```extremalPts.length == 3```.

I thought this can cause unexpected result  when ```extremalPts``` consists of 3 points forming acute triangle, and the longest segment is ```extremalPts[2]-[0]```. 
(Please take a look at the case added at ```MinimumBoundingCircleTest.testMinDiameterPolygon()```.)

So I'm suggesting a change for ```farthestPoints()``` to handle this case as well by checking all three segment length.

Maybe I have misunderstood the purpose of ```getMaximumDiameter()```, but if I have understood the method description comments correctly, this fix could be helpful.
Thank you :)

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/algorithm/MinimumBoundingCircle.java
+++ b/modules/core/src/main/java/org/locationtech/jts/algorithm/MinimumBoundingCircle.java
@@ -162,10 +162,14 @@ public class MinimumBoundingCircle
   private static Coordinate[] farthestPoints(Coordinate[] pts) {
     double dist01 = pts[0].distance(pts[1]);
     double dist12 = pts[1].distance(pts[2]);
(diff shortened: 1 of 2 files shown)


dr-jts:
This is great, thanks.

Would you be able to fix the test file to say `maxDiameter` in the method names instead of `minDiameter` ?

tink-expo:
Fix applied :)

dr-jts:
Great, thanks.

PR #543, opened 2020-05-23
Make STRtree use Java PriorityQueue and deprecate internal

bjornharrtell:
Signed-off-by: Björn Harrtell <bjorn@wololo.org>


dr-jts:
Deleting the `PriorityQueue` class is a breaking change.  Is that necessary for transpilation, or could you simply blacklist it?

bjornharrtell:
Ah, didn't realize that. I just had a quick "ah this code isn't needed anymore" thought.

It's not important for JSTS. But do you agree STRtree should use Java built in instead and that we could mark the internal one as deprecated?

dr-jts:
Yes, makes sense to switch STRtree to the Java PQ.

bjornharrtell:
@dr-jts reworked and ready for review.

Issue #545, opened 2020-06-04
org.locationtech.jts.geom.impl not exposed in OSGI environment


qtran:
Hi, 
I would like to use the classes in 'org.locationtech.jts.geom.impl' in a OSGI environment, unfortunately that package is not exported. Any plans to expose this?

I tested this with jts-core 1.16.1

Kind regards

jnh5y:
Generally, this sounds reasonable.  

Speaking for myself, I'm not savvy on OSGI.  What changes would need to be made?  Are you able to contribute to the project directly (see https://github.com/locationtech/jts/blob/master/CONTRIBUTING.md for details)?

qtran:
You'll need to adapt the configuration of the org.apache.felix:maven-bundle-plugin (modules/core/pom.xml)
I should be able to contribute to the project, I'll try to open a pull request in the near future

qtran:
Fixed with 1.17.0

PR #547, opened 2020-06-08
Exposing all packages under org.locationtech.* in OSGI enviroment

qtran:
This PR exposes all packages under 'org.locationtech.*' in a OSGI enviroment, as a result the package 'org.locationtech.jts.geom.impl' is also exposed. (By default the maven-bundle-plugin is excluding all packages containing 'impl' or 'internal')

relates to #545

Signed-off-by: Quang Tran <quang.tran@mailbox.org>


qtran:
Messed up with the commit message

PR #553, opened 2020-06-22
Fix GML coordinates and srsName parsing

dr-jts:
* `coordinates` element parsing now handles spaces before and after comma separators
* `srsName` attribute parsing now extracts the final numeric value as the geometry SRID

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/io/gml2/GeometryStrategies.java
+++ b/modules/core/src/main/java/org/locationtech/jts/io/gml2/GeometryStrategies.java
@@ -336,6 +336,11 @@ public class GeometryStrategies{
 				// now to start parse
 				String t = arg.text.toString();
 				t = t.replaceAll("\\s"," ");
+				/**
+				 * Remove spaces after commas, for when they are used as separators (default).
+				 * This prevents coordinates being split by the tuple separator
+				 */
+				t = t.replaceAll(",\\s+", ",");
 				
 				Pattern ptn = (Pattern) patterns.get(toupleSeperator);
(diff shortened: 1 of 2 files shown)


modules/core/src/main/java/org/locationtech/jts/io/gml2/GeometryStrategies.java
```diff
@@ -336,6 +337,11 @@ else if(arg.attrs.getIndex(GMLConstants.GML_NAMESPACE,"ts")>=0)
 				// now to start parse
 				String t = arg.text.toString();
 				t = t.replaceAll("\\s"," ");
+				/**
+				 * Remove spaces after commas, for when they are used as separators (default).
+				 * This prevents coordinates being split by the tuple separator
+				 */
+				t = t.replaceAll(",\\s+", ",");
```
satishbhor, line 344: I think we should also handle spaces before comma as well?
```
<gml:Point gml:id="p21" srsName="http://www.opengis.net/def/crs/EPSG/0/4326">
    <gml:coordinates>45.67     ,88.56</gml:coordinates>
 </gml:Point>
```

dr-jts: yes, probably should

satishbhor: Thanks!

dr-jts: Done!

PR #556, opened 2020-06-24
Add Feature and FeatureCollection support to geojson package

icholy:


dr-jts:
I am reluctant to accept this into JTS.  Reasons are:

* JTS so far has avoided supporting anything at the Feature level
* This widens the API and hence incurs some support cost, potentially

Could this be provided as a pure extension in a separate project?  If so we could set up an area in the JTS doc to link to extension projects.  This might be nice for other people as well.

icholy:
@dr-jts I get that. The problem is that almost all geojson data in the wild wraps geometries in Features and FeatureCollections. If I wanted to leverage the geojson package in its current state, I'd need to parse as a Feature/FeatureCollection, format the geometry portion back to json and then parse that using `GeoJsonReader`. Then I'd need a second set of Feature/FeatureCollection classes that have jts Geometry's as their fields. Point being, the geojson package is essentially useless in it's current state. This is less about having a Feature representation and more about actually being able to parse geojson.

adahn6:
+1 to the Feature/FeatureCollection usage being widespread, in [Atlas](https://github.com/osmlab/atlas) we currently build our own geojson serialization to support this style, but this would be nice to have.

modules/io/common/src/test/java/org/locationtech/jts/io/geojson/GeoJsonTest.java
```diff
@@ -92,11 +94,55 @@ public void testGeometryCollectionEmpty() throws ParseException {
     runTest("GEOMETRYCOLLECTION EMPTY");
   }
 
+  public void testFeature() throws ParseException {
```
dr-jts, line 97: Shouldn't there be some tests which have actual property data, to show they can be handled?

modules/io/common/src/main/java/org/locationtech/jts/io/geojson/GeoJsonWriter.java
```diff
@@ -34,7 +34,7 @@
 
 
 /**
- * Writes {@link Geometry}s as JSON fragments in GeoJson format.
+ * Writes {@link Geometry}s, {@link Feature}s, and {@link FeatureCollection}s as JSON fragments in GeoJson format.
```
dr-jts, line 37: The Feature and FeatureCollection are full JSON documents, not just fragments I think?

modules/io/common/src/main/java/org/locationtech/jts/io/geojson/GeoJsonConstants.java
```diff
@@ -32,5 +35,7 @@
   public static final String NAME_MULTIPOLYGON = "MultiPolygon";
   public static final String NAME_MULTILINESTRING = "MultiLineString";
   public static final String NAME_MULTIPOINT = "MultiPoint";
+  public static final String NAME_FEATURE = "Feature";
```
dr-jts, line 38: Would be nice to improve the ordering of these constants.  Either alphabetically, or put the new ones required for Features at the end

modules/io/common/src/main/java/org/locationtech/jts/io/geojson/FeatureCollection.java
```diff
@@ -0,0 +1,24 @@
...
+import java.util.ArrayList;
+
+/**
+ * A FeatureCollection in GeoJSON contain a collection of zero or more features.
+ * <p>
+ * A specification of the GeoJson format can be found at the GeoJson web site:
+ * <a href='http://geojson.org/geojson-spec.html'>http://geojson.org/geojson-spec.html</a>.
+ * <p>
+ */
+public class FeatureCollection extends ArrayList<Feature> {
```
dr-jts, line 12: Not sure its good to extend ArrayList.   It introduces a big dependency which might be brittle over time.

dr-jts: Perhaps just let the API accept a raw `List<Feature>` (or even `Collection<Feature>`) for writing, and return `List<Feature>` for reading.

dr-jts:
@icholy Yes, I understand the awkwardness of handling Features and FeatureCollections.

One idea I'm leaning towards is to keep the new code in a separate class called `GeoJSONFeatureReader`.  It can just call `GeoJSONReader.create`.  (Really `GeoJSONReader` should be called `GeoJSONGeometryReader`, but too late now).  Likewise for `GeoJSONFeatureWriter`.  That partitions the code a bit to make it easier to understand, and it reduces the risk of API breakage.

icholy:
@dr-jts making `GeoJSONReader.create` public solves all my problems.

dr-jts:
Isn't the `crs` property of Geometry deprecated?!

dr-jts:
Anyway, I have no problem with exposing those methods as `public`. The same thought had occurred to me.

Still, might be nice to provide the extra code that implements Feature and FeatureCollection parsing.  This could be done somewhere else in JTS (for example, in the TestBuilder, which would be handy).  That doesn't have to be done right now though.

Are you willing to make a new PR for exposing `create` and `getGeometryFactory`?  With some unit tests...

Also, I feel like `getGeometryFactory` should really just delegate to a new method `readSRID`, which would do the work of actually extracting the SRID.  Wrapping in a `GeometryFactory` really just hides the value of interest.  And then perhaps that would be the method that is exposed?

icholy:
@dr-jts sure thing.

PR #559, opened 2020-07-06
CGAlgorithms: links to alternatives for @deprecated methods and fields

sebkur:
I came across some warnings in one of my projects using JTS warning me about deprecated methods I was using from the CGAlgorithms class. I had trouble finding the actual replacement methods. To spare other users of the library that trouble I added corresponding Javadoc tags @deprecated pointing to the alternative implementations.


dr-jts:
This is a good improvement.  Can you sign an ECA?  If not, I can replicate this PR and commit it.

sebkur:
I do have an Eclipse account, and when I log in to that it tells me <email> is not covered by the necessary legal agreements in order to proceed.

sebkur:
Planning on making more contributions so I would prefer to get this right so that it won't be a hurdle in the future

sebkur:
ah, just force-pushed to trigger Github to recheck and that did the trick

PR #564, opened 2020-07-16
Enable WKB reading/writing for POINT EMPTY

jagill:
Although the WKB spec does not support empty points, ESRI and others
serialize them as points with NaN coordinates.  Empty points are the
only valid geometry that were not serializable: this commit enables
serialization for them, so users can serialize all valid geometries
with WKB.


jagill:
The Eclipse contributor agreement requires me to allow them to publicly distribute my mailing address, which I'm not keen on.  Please feel free to use this contribution directly, or by taking all or part of the code, in any way you wish.

dr-jts:
Replaced by PR #567.

PR #582, opened 2020-08-18
Adjust reverse and reverseInternal visibility and type narrowing

jodygarnett:
This change addresses an unintended consequence where the visibility of these methods as implemented by geometry subclasses was consistent.

* [x] consistent public visibility for reverse() methods
* [x] consistent protected visibility for reverseInternal() methods
* [x] type narrowing for each subclass implementation of reverse()
* [x] subclass implementations of reverse() are not deprecated

Signed-off-by: Jody Garnett <jody.garnett@gmail.com>

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/geom/LinearRing.java
+++ b/modules/core/src/main/java/org/locationtech/jts/geom/LinearRing.java
@@ -132,7 +132,7 @@ public class LinearRing extends LineString
     return new LinearRing(points.copy(), factory);
   }
 
-  public LinearRing reverse()
+  protected LinearRing reverse()
   {
     return (LinearRing) super.reverse();
   }
--- a/modules/core/src/main/java/org/locationtech/jts/geom/LinearRing.java
+++ b/modules/core/src/main/java/org/locationtech/jts/geom/LinearRing.java
@@ -132,13 +132,12 @@ public class LinearRing extends LineString
     return new LinearRing(points.copy(), factory);
   }
 
-  /** @deprecated */
-  public Geometry reverse()
+  public LinearRing reverse()
   {
-    return super.reverse();
+    return (LinearRing) super.reverse();
   }
 
-  public Geometry reverseInternal() {
+  public LinearRing reverseInternal() {
     CoordinateSequence seq = points.copy();
     CoordinateSequences.reverse(seq);
     return getFactory().createLinearRing(seq);
--- a/modules/core/src/main/java/org/locationtech/jts/geom/MultiLineString.java
+++ b/modules/core/src/main/java/org/locationtech/jts/geom/MultiLineString.java
@@ -102,10 +102,9 @@ public class MultiLineString
    * are reversed.
    *
    * @return a {@link MultiLineString} in the reverse order
-   * @deprecated
    */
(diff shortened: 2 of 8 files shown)


modules/core/src/main/java/org/locationtech/jts/geom/LinearRing.java
```diff
@@ -132,13 +132,12 @@ protected LinearRing copyInternal() {
     return new LinearRing(points.copy(), factory);
   }
 
-  /** @deprecated */
-  public Geometry reverse()
+  public LinearRing reverse()
```
jodygarnett, line 135: This is no longer deprecated, a change that was producing warnings in client code.

modules/core/src/main/java/org/locationtech/jts/geom/MultiLineString.java
```diff
@@ -102,10 +102,9 @@ public Geometry getBoundary()
    * are reversed.
    *
    * @return a {@link MultiLineString} in the reverse order
-   * @deprecated
    */
-  public Geometry reverse() {
-    return super.reverse();
+  public MultiLineString reverse() {
```
jodygarnett, line 106: This is no longer deprecated, which was producing warnings in client code.

jodygarnett:
Discussion with @jnh5y asks about the use of `@Override` annotation. These methods are appropriate for the `@Override` annotation, but I did not see it used elsewhere in the `Geometry` implementation ...

Happy to add `@Override`, especially to the protected `reverseInternal()` method which would be easy to miss during a refactoring.

jodygarnett:
> Is it even necessary to have `reverse()` methods on subclasses, given that they all delegate to the `Geometry` method?

The one advantage is the type narrowing, in case where the geometry type is known.

jodygarnett:
> Is the type-narrowing needed? Or just the removal of `deprecated`?

I decided to be explicit and check each case, the removal and deprecated and the change of method visibility are the priority.

jnh5y:
Alright, finally had a second to look into this more.  

Jody's type narrowing is the same as how copyInternal is currently handled on master.  Given that precedent, I think the type narrowing is justified.

modules/core/src/main/java/org/locationtech/jts/geom/LinearRing.java
```diff
@@ -132,13 +132,12 @@ protected LinearRing copyInternal() {
...
   {
-    return super.reverse();
+    return (LinearRing) super.reverse();
   }
 
-  public Geometry reverseInternal() {
+  public LinearRing reverseInternal() {
```
jnh5y, line 140: I believe this should be protected; otherwise this will be the only reverseInternal which has public scope!

jodygarnett: Thank you, good catch

jodygarnett: sorry I missed applying this feedback :(

jnh5y: Actually, I think you switched reverse to protected instead of this method.

PR #584, opened 2020-08-26
Remove enforcing array consistency from CoordinateArraySequence

jodygarnett:
Removes check on array used for CoordinateArraySequence, documenting instead that the array must be consistent with the dimension and measures provided.  Refactored the protected enforceConsistency method out to the Coordintes utility class for any client code that ends up working with mixed coordines.

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/geom/Coordinates.java
+++ b/modules/core/src/main/java/org/locationtech/jts/geom/Coordinates.java
@@ -46,7 +46,7 @@ public class Coordinates {
     }
     return new Coordinate();
   }
-    
+  
   /**
    * Determine dimension based on subclass of {@link Coordinate}.
    * 
--- a/modules/core/src/main/java/org/locationtech/jts/geom/impl/CoordinateArraySequence.java
+++ b/modules/core/src/main/java/org/locationtech/jts/geom/impl/CoordinateArraySequence.java
@@ -75,8 +75,11 @@ public class CoordinateArraySequence
   
   /**
    * Constructs a sequence based on the given array 
-   * of {@link Coordinate}s (the
-   * array is not copied).
+   * of {@link Coordinate}s (the array is not copied).
+   * <p>
+   * It is your responsibility to ensure the array contains Coordinates of the
+   * indicated dimension and measures (See 
+   * {@link CoordinateArrays#enforceConsistency(Coordinate[])} ).</p>
    *
    * @param coordinates the coordinate array that will be referenced.
    * @param dimension the dimension of the coordinates
@@ -89,7 +92,7 @@ public class CoordinateArraySequence
       this.coordinates = new Coordinate[0];
     }
     else {
-      this.coordinates = enforceArrayConsistency(coordinates);
+      this.coordinates = coordinates;
     }
   }
 
@@ -158,48 +161,6 @@ public class CoordinateArraySequence
     }
   }
 
-  /**
-   * Ensure array contents of the same type, making use of {@link #createCoordinate()} as needed.
-   * <p>
-   * A new array will be created if needed to return a consistent result.
-   * </p>
-   * 
-   * @param array array containing consistent coordinate instances
-   */
-  protected Coordinate[] enforceArrayConsistency(Coordinate[] array)
-  {
-     Coordinate sample = createCoordinate();
-     Class<?> type = sample.getClass();
-     boolean isConsistent=true;
-     for( int i = 0; i < array.length; i++) {
-       Coordinate coordinate = array[i];
-       if( coordinate != null && !coordinate.getClass().equals(type)) {
-         isConsistent = false;
-         break;
-       }
-     }
-     if( isConsistent ){
-       return array;
-     }
-     else {
-       Class<? extends Coordinate> coordinateType = sample.getClass();
-       Coordinate copy[] = (Coordinate[]) Array.newInstance(coordinateType, array.length);
-       for ( int i = 0; i < copy.length; i++){
-          Coordinate coordinate = array[i];
-          if( coordinate != null && !coordinate.getClass().equals(type)){
-            Coordinate duplicate = createCoordinate();
-            duplicate.setCoordinate(coordinate);
-            copy[i] = duplicate;
-          }
-          else {
-            copy[i] = coordinate;
-          }
-       }
-       return copy;
-     }
-  }
-
-  
   /**
    * @see org.locationtech.jts.geom.CoordinateSequence#getDimension()
    */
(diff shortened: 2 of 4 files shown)


modules/core/src/main/java/org/locationtech/jts/geom/impl/CoordinateArraySequence.java
```diff
@@ -168,35 +171,7 @@ public CoordinateArraySequence(CoordinateSequence coordSeq)
...
-            duplicate.setCoordinate(coordinate);
-            copy[i] = duplicate;
-          }
-          else {
-            copy[i] = coordinate;
-          }
-       }
-       return copy;
-     }
+     return Coordinates.enforceConsistency(array, dimension, measures);
```
davidblasby, line 174: should this method be removed?  (from p2p conversation)

modules/core/src/main/java/org/locationtech/jts/geom/Coordinates.java
```diff
@@ -46,7 +48,100 @@ public static Coordinate create(int dimension, int measures)
...
+   * Utility method ensuring array contents are of the specified dimension and measures.
+   * <p>
+   * Array is modified in place if required, coordinates are replaced in the array as required
+   * to ensure all coordinates have the same dimension and measures.
+   * </p>
+   *
+   * @param array
+   * @return array returned, or copy created if required to enforce consistency.
+   */
+  public static Coordinate[] enforceConsistency(Coordinate[] array)
```
davidblasby, line 62: test case?  Or is it covered elsewhere?

modules/core/src/main/java/org/locationtech/jts/geom/Coordinates.java
```diff
@@ -46,7 +48,100 @@ public static Coordinate create(int dimension, int measures)
...
+   * Array is returned unmodified if consistent, or a copy of the array is made with
+   * each inconsistent coordinate duplicated into an instance of the correct dimension and measures.
+   * </p></>
+   *
+   * @param array coordinate array
+   * @param dimension
+   * @param measures
+   * @return array returned, or copy created if required to enforce consistency.
+   */
+  public static Coordinate[] enforceConsistency(Coordinate[] array,int dimension, int measures)
```
davidblasby, line 112: test case?  Or is it covered elsewhere?

modules/core/src/main/java/org/locationtech/jts/geom/Coordinates.java
```diff
@@ -46,7 +48,100 @@ public static Coordinate create(int dimension, int measures)
...
+   * Array is modified in place if required, coordinates are replaced in the array as required
+   * to ensure all coordinates have the same dimension and measures.
+   * </p>
+   *
+   * @param array
+   * @return array returned, or copy created if required to enforce consistency.
+   */
+  public static Coordinate[] enforceConsistency(Coordinate[] array)
+  {
+    if( array == null ){
```
davidblasby, line 64: if (array == null ) {

modules/core/src/main/java/org/locationtech/jts/geom/Coordinates.java
```diff
@@ -46,7 +48,100 @@ public static Coordinate create(int dimension, int measures)
...
+      return array;
+    }
+    // step one check
+    int dimension = 0;
+    int measures = 0;
+    boolean isConsistent=true;
```
davidblasby, line 70: the spacing isn't consistent with above ;)

modules/core/src/main/java/org/locationtech/jts/geom/impl/CoordinateArraySequence.java
jodygarnett: Thanks you are correct it moved to CoordinateArrays utility class for client code to call if needed

modules/core/src/main/java/org/locationtech/jts/geom/LinearRing.java
```diff
@@ -132,7 +132,7 @@ protected LinearRing copyInternal() {
     return new LinearRing(points.copy(), factory);
   }
 
-  protected LinearRing reverse()
+  public LinearRing reverse()
```
dr-jts, line 135: Would be better if this was not part of this PR

modules/core/src/main/java/org/locationtech/jts/geom/Coordinates.java
jodygarnett: I will add a test case, thanks! Previously it was covered by other tests

modules/core/src/main/java/org/locationtech/jts/geom/LinearRing.java
jodygarnett: yes it would, let me fix and then rebase this PR.

modules/core/src/main/java/org/locationtech/jts/geom/Coordinates.java
```diff
@@ -11,6 +11,8 @@
  */
 package org.locationtech.jts.geom;
 
+import java.lang.reflect.Array;
```
davidblasby, line 14: probably not used?

jodygarnett:
Thanks for review everyone

modules/core/src/main/java/org/locationtech/jts/geom/impl/CoordinateArraySequence.java
```diff
@@ -75,8 +75,11 @@ public CoordinateArraySequence(Coordinate[] coordinates, int dimension) {
   
   /**
    * Constructs a sequence based on the given array 
-   * of {@link Coordinate}s (the
-   * array is not copied).
+   * of {@link Coordinate}s (the array is not copied).
+   * <p>
+   * Ensure the array is of the indicated dimension and measures. The method
```
jnh5y, line 80: Hmm... I wonder if this explanation should be expanded.  

The word "Ensure" is throwing me here.  Read quickly, could one think that the method is ensuring consistency?  (Like it was previously?)  

Also, if the requirement is on the inputs, it seems like it'd be better on the description of the coordinates variable.

Honestly, mixing Coordinates and using the common ordinates ought to be fine.  Right?

As a concrete suggestion how about something this:

"CoordinateArray should likely contain consistent implementations of the Coordinate for most uses.  Mixing types may lead to issues.  Callers can create such consistent arrays using {@link CoordinateArrays#enforceConsistency(Coordinate[])}."

jodygarnett:
@jnh5y I have improved the javadoc, David also found a some code improvements to be made. Please merge when the automated checks are passed.

jodygarnett:
Hold up I see Martin marked this as draft, sadly I removed that before noticing he just did so 10 mins ago.

@dr-jts what is up?

dr-jts:
> Hold up I see Martin marked this as draft, sadly I removed that before noticing he just did so 10 mins ago.
> 
> @dr-jts what is up?
Well it appears to still be changing.  But I guess as a result of review.  Is there a way to sign off on review?

Issue #581, opened 2020-08-15
Issue with RayCrossingCounter and coordinates with Measures


mukoki:
Some algo may fail with coordinates with measures.
For example, with RayCrossingCounter.locatePointInRing(coord, coordSequence)
If CoordinateSequence is
dim = 4, measures = 0 (see previous bug report) : OK
dim = 4, measures = 1 (normal XYZM coordinate) : throws exception at ring.getCoordinate(i, p1)

Problem is that RayCrossingCounter.locatePointInRing tries to set the M of the XYZM coordinate to a default 3D Coordinate without M initialized in the method. I suppose XYM coordinates has the same problem in this case.

dr-jts:
What previous bug report?
Can you supply a stack trace?
What kind of CoordinateSequence?
Can you supply code reproducing the problem?

mukoki:
I added a PR with a test building a 4D XYZM CoordinateSequence which makes RayCrossingCounter#locatePointInRing fail
The fix I propose makes another test fail in Distance3DOp
I propose another fix for this failure consecutive to my first fix, but this is about 3d polygons projected to the XY, XZ or YZ plan, and I have not a clear idea about what all this code does.

dr-jts:
Fixed by #589

PR #586, opened 2020-08-27
Bug 581 ray crossing with 4d coord

mukoki:
Test to demonstrate #581 bug and proposal to fix it


dr-jts:
The fix looks good to me. Are you going to sign an ECA?

dr-jts:
Actually I'm thinking a better fix might be to change `RayCrossingCounter.countSegment` to use ordinate values instead of coordinates.

Thoughts?  I can make a PR for this.

mukoki:
> The fix looks good to me. Are you going to sign an ECA?

Wow, lot of ceremony. I just filled the ECA form, but I don't know how to link my commits to this document.

mukoki:
> Actually I'm thinking a better fix might be to add a `RayCrossingCounter.countSegment` that accepts ordinate values instead of coordinates.
> 
> Thoughts? I can make a PR for this.

I think you're right. We don't really need to get coordinates with the exact coord dimension. We rather need x,y from both coordinates. In this case, you can also remove p1 and p2 coordinate creation and ring.getCoordinate(i, p1) call which is the place where the error occured.

dr-jts:
> Wow, lot of ceremony. I just filled the ECA form, but I don't know how to link my commits to this document.

Yes, the bureaucracy is annoying.  AFAIK you can't amend this PR - you have to create another one.

dr-jts:
Another possibility - just use `CoordinateSequence.getCoordinate`.  Pretty sure modern JVMs optimize small object creation highly, so perhaps this is performant enough to use.  Confirming this with a performance test would be good.

mprins:
> > Wow, lot of ceremony. I just filled the ECA form, but I don't know how to link my commits to this document.
> 
> Yes, the bureaucracy is annoying. AFAIK you can't amend this PR - you have to create another one.

when you go to https://accounts.eclipse.org/legal/eca/validation/45250 you cab see that a signed commit is needed, use `git commit -a -s`

mukoki:
> Another possibility - just use `CoordinateSequence.getCoordinate`. Pretty sure modern JVMs optimize small object creation highly, so perhaps this is performant enough to use. Confirming this with a performance test would be good.

Do you mean as I did in the commit (bf88792abecf26538e15cf9583bd95e9b8a1a825) ?
Seems to work, but probably do more work than necessary compared to your idea to just use x,y ordinates. I'll try to test the later idea. Not sure I can easily compare performance though.

dr-jts:
> > Another possibility - just use `CoordinateSequence.getCoordinate`. Pretty sure modern JVMs optimize small object creation highly, so perhaps this is performant enough to use. Confirming this with a performance test would be good.
> 
> Do you mean as I did in the commit ([bf88792](https://github.com/locationtech/jts/commit/bf88792abecf26538e15cf9583bd95e9b8a1a825)) ?

I'm seeing that commit use `createCoordinate`?

I mean using `CoordinateSequence.getCoordinate(i)`. That results in a Coordinate being created for every access, but this should be fairly efficient.

dr-jts:
Superceded by #589

PR #593, opened 2020-09-03
Add KMLReader for reading KML geometry elements with attributes

mpanchekhin:
Signed-off-by: Maksim Panchekhin <maksimpanchekhin@gmail.com>

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/io/kml/KMLReader.java
+++ b/modules/core/src/main/java/org/locationtech/jts/io/kml/KMLReader.java
@@ -1,0 +1,358 @@
+/*
+ * Copyright (c) 2016 Vivid Solutions.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * and Eclipse Distribution License v. 1.0 which accompanies this distribution.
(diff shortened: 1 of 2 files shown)


modules/core/src/main/java/org/locationtech/jts/io/kml/KMLReader.java
```diff
@@ -0,0 +1,358 @@
...
+import javax.xml.stream.XMLInputFactory;
+import javax.xml.stream.XMLStreamException;
+import javax.xml.stream.XMLStreamReader;
+import java.io.StringReader;
+import java.lang.reflect.Array;
+import java.util.*;
+
+/**
+ * Constructs a {@link Geometry} object from the OGC KML representation.
+ * Works only with KML geometry fragments
```
dr-jts, line 27: Perhaps say "elements", and mention that it also parses attributes within that element.

dr-jts:
This looks very good - thanks!

modules/core/src/main/java/org/locationtech/jts/io/kml/KMLReader.java
mpanchekhin: Fixed

PR #594, opened 2020-09-08
FAQ: repair links to javadocs; mention correct Java version (1.8)

sebkur:
I think the reference to Java 1.4.2 is outdated and 1.8 is probably the right version, correct?

Also the links to the methods in the Javadocs do not work, I hope they do now.

Also, I updated the 'Last Update' time...


sebkur:
Okay I think Java 8 should not be referenced as Java 1.8

Issue #602, opened 2020-09-24
StackOverflowError in KdTree running InteriorPointAreaPerfTest


FObermaier:
Running the `InteriorPointAreaPerfTest` leads to the following output:
```
Interior Point Area perf test
SineStar: origin: (100.0, 100.0)  size: 100.0  # arms: 20  arm ratio: 0.3
# Iterations: 100

Running with # pts 11
runTest1 : 59 ms

Running with # pts 101
runTest1 : 3 ms

Running with # pts 1001
runTest1 : 5 ms

Running with # pts 10001
runTest1 : 9 ms

Running with # pts 100001
runTest1 : 65 ms
Exception in thread "main" java.lang.StackOverflowError
	at org.locationtech.jts.index.kdtree.KdTree.queryNodePoint(KdTree.java:330)
	at org.locationtech.jts.index.kdtree.KdTree.queryNodePoint(KdTree.java:327)
	... many, many more
	at org.locationtech.jts.index.kdtree.KdTree.queryNodePoint(KdTree.java:330)
	at org.locationtech.jts.index.kdtree.KdTree.queryNodePoint(KdTree.java:327)
```

~I'm afraid there are other tests in the performance section that fail, too.~
Double checked, it is just this test.

dr-jts:
The performance tests aren't for automatic validation - they are just useful for manual testing of performance issues and comparisons.

That said, this is due to the `GeometryPrecisionReducer` being changed to use OverlayNG to do precision reduction on polygons.  This uses the `SnapRoundingNoder`, which in turn uses a KdTree.  It is interesting that querying the `KdTree` is so stack-intensive.  This indicates that perhaps it would be better to query the tree using an iterative approach rather than a recursive call.  It might also indicate that the KD-tree is not ideal for this use.

And perhaps another thing to ponder is whether it would be better to try a pointwise precision reduction first, and only snap-round if the result is invalid.  That's a tough call - it's a costly way to see if snap-rounding is needed.

It also indicates that using a precision reduction to generate test data for this test isn't ideal (since it takes far longer than the test itself!)

Are the other performance test failures similar?

dr-jts:
After some investigation, I think this is due to the `KdTree` being built in a very unbalanced way, thus causing a large depth of recursion.  But it's odd, because running a similar case in the TestBuilder doesn't behave this way - it has a much better balanced tree.  More investigation required.

The good news is that I haven't seen this problem occur in other testing (which has processed some pretty big geometry).

[...]

PR #605, opened 2020-09-30
Fix StackOverflowError in HotPixelIndex

FObermaier:
Add nested CoordinateShuffler class and wire it `add(Coordinate[] pts)` in order to get a more balanced KdTree. The insertion speed for geometries with many coordinates greatly improves!

relates to #602 

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/noding/snapround/HotPixelIndex.java
+++ b/modules/core/src/main/java/org/locationtech/jts/noding/snapround/HotPixelIndex.java
@@ -11,7 +11,9 @@
  */
 package org.locationtech.jts.noding.snapround;
 
+import java.util.Iterator;
 import java.util.List;
+import java.util.Random;
 
 import org.locationtech.jts.geom.Coordinate;
 import org.locationtech.jts.geom.Envelope;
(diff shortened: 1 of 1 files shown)


dr-jts:
Nice fix, @FObermaier !  I was starting to think about this myself, but this is a very elegant implementation.  And it provides a dramatic performance boost.

dr-jts:
I wonder if it would be even better to skip shuffling if there are only a small number of points (say <= 100) ?

dr-jts:
Or another idea:  instead of randomly shuffling the array indices, step through the coordinate array with a stride length which is coprime to the array size.  That will visit every array entry once, but in a jumbled order.

This may not make a big peformance difference, but will use less memory. 

The trick is finding a stride length which is relatively prime.   One way to do it is given [here](https://math.stackexchange.com/questions/2430742/finding-a-coprime-of-a-general-magnitude)

dr-jts:
Further testing indicates that:
* Gating the shuffling to only use it for larger arrays sizes doesn't make an appreciable difference in performance (this is often the case for indexes - there's not much advantage in not using them)
* Using a coprime stride to step through the array is slightly better, except for some cases when it is MUCH worse.  So that is too risky to use

Based on this, I'll leave this approach as it is.

PR #615, opened 2020-10-17
Control Geometry overlay algorithm via system property

dr-jts:
Control Geometry overlay algorithm via a system property `jts-overlay`.  This makes it possible to use OverlayNG in existing deployments with no code changes.

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/geom/Geometry.java
+++ b/modules/core/src/main/java/org/locationtech/jts/geom/Geometry.java
@@ -1325,31 +1325,7 @@ public abstract class Geometry
    */
   public Geometry intersection(Geometry other)
   {
-  	/**
-  	 * TODO: MD - add optimization for P-A case using Point-In-Polygon
-  	 */
-    // special case: if one input is empty ==> empty
-    if (this.isEmpty() || other.isEmpty())
-      return OverlayOp.createEmptyResult(OverlayOp.INTERSECTION, this, other, factory);
-
-    // compute for GCs
-    // (An inefficient algorithm, but will work)
-    // TODO: improve efficiency of computation for GCs
-    if (this.isGeometryCollection()) {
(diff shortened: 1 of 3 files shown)


FObermaier:
I have a different proposal: Why don't we tie geometry operations/predicates to the `GeometryFactory`. The geometry would invoke the operation by passing itself and arguments to a function on the factory.
That way a factory that calculates distances in meters or areas in square meters from lat-lon coordinates could be easyly added.

dr-jts:
> I have a different proposal: Why don't we tie geometry operations/predicates to the `GeometryFactory`. The geometry would invoke the operation by passing itself and arguments to a function on the factory.
> That way a factory that uses old Overlay code and one that uses OverlayNG code is easily doable,
> plus one that calculates distances in meters or areas in square meters from lat-lon coordinates could be easyly added.

A good idea.  We have had design discussions about JTS 2, and one of the ideas is to provide a geometry method registry concept, to support this kind of flexibility.  The current concept is to make it static, so that it's easy to access from Geometry implementations (in JTS 2 `Geometry` will be an interface).  But it could make sense to attach it to the `GeometryFactory`, since that is already injected into geometries.  

Perhaps this would be done as a `GeometryMethods` interface, with a set of methods matching the `Geometry` methods (but taking an initial `Geometry` argument).  It would be backed by a concrete class that could be overridden for individual methods.

That said, the goal of this small enhancement is to make it easy for current users of JTS to switch to using OverlayNG with no code changes.  Thus I think the PR is reasonable as it stands.  The implementation can be enhanced as and when there is clear demand for it.

dr-jts:
I realized that this code should use OverlayNG with Snap-Rounding for geometries with a fixed precision model. Will add an enhancement for that.  

And since this is starting to get complex, I will expose a method to set the overlayNG flag programmatically, to allow unit testing.

PR #618, opened 2020-10-23
Fix OverlayNGRobust self-snapping to prevent heterogeneous results

dr-jts:
Fixes `OverlayNGRobust` self-snapping to prevent heterogeneous results causing failure in subsequent overlay operations.

Issue first identified during integration testing of OverlayNG in R-sf (seen https://github.com/r-spatial/sf/issues/794).


strk:
GEOS port will be tracked by https://trac.osgeo.org/geos/ticket/1062

PR #633, opened 2020-11-16
Add the check of userData in equals(object o)

jiayuasu:
This PR proposes to add the equality check for userData in equsl(Object o). This will make sure two geometries are exactly same

Signed-off-by: Jia Yu <jiayu198910@gmail.com>


jiayuasu:
@jnh5y @dr-jts Hey Jim and Martin, could you please take a look at this PR (1/2)?

This PR is related to the JTS dependency in Sedona (GeoSpark). Jim has known this issue long time ago.

jnh5y:
@jiayuasu thanks for putting this PR up.  How is Sedona using equals?  Is part of needing that Geometry are Comparable?  

I'm wondering if it'd be possible for Sedona to provide alternative "evidence" that Geometry is Comparable.

Issue #635, opened 2020-11-17
Densifier convert 2d to 3d


halset:
Looks like `Densifier` on a 2d line convert it to a 3d line. It would be better if it would keep the dimension.

```java
      GeometryFactory gf = new GeometryFactory();
      LineString line = gf
              .createLineString(new Coordinate[] { new CoordinateXY(1, 2), new CoordinateXY(3, 4) });
      assertEquals(2, line.getCoordinateSequence().getDimension());
      
      line = (LineString) Densifier.densify(line, 0.1);
      // the following assert fails, but should not.
      assertEquals(2, line.getCoordinateSequence().getDimension());
```

This got me in to a problem with a GeoTools transformation failing.
```java
java.lang.IllegalArgumentException: Invalid ordinate index: 2
	at org.locationtech.jts.geom.CoordinateXY.getOrdinate(CoordinateXY.java:106)
	at org.locationtech.jts.geom.impl.CoordinateArraySequence.getOrdinate(CoordinateArraySequence.java:256)
	at org.geotools.geometry.jts.DefaultCoordinateSequenceTransformer.transform(DefaultCoordinateSequenceTransformer.java:123)
	at org.geotools.geometry.jts.GeometryCoordinateSequenceTransformer.projectCoordinateSequence(GeometryCoordinateSequenceTransformer.java:262)
	at org.geotools.geometry.jts.GeometryCoordinateSequenceTransformer.transformStraightLineString(GeometryCoordinateSequenceTransformer.java:214)
	at org.geotools.geometry.jts.GeometryCoordinateSequenceTransformer.transformLineString(GeometryCoordinateSequenceTransformer.java:193)
	at org.geotools.geometry.jts.GeometryCoordinateSequenceTransformer.transform(GeometryCoordinateSequenceTransformer.java:139)
	at org.geotools.geometry.jts.JTS.transform(JTS.java:437)
```

halset:
`LineSegment.pointAlong` (and other methods on `LineSegment`) create a new `Coordinate` even though both coordinates of the segment are of type `CoordinateXY`. Changing this fixes the problem for me, but perhaps there is a better way?

dr-jts:
Sigh...  yes, the handling of Coordinate subclasses is imperfect, and needs work.  Thanks for pointing this out.

I think maybe one part of the solution is to provide `create` methods on Coordinate and subclasses, which instantiate a Coordinate of appropriate type.

halset:
[...]

PR #636, opened 2020-11-17
fix Densifier on a 2d geometry

halset:
Make sure `Densifier` insert extra `Coordinate`s of the same type and preserve the dimension. https://github.com/locationtech/jts/issues/635

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/geom/Coordinate.java
+++ b/modules/core/src/main/java/org/locationtech/jts/geom/Coordinate.java
@@ -400,6 +400,15 @@ public class Coordinate implements Comparable<Coordinate>, Cloneable, Serializab
   public Coordinate copy() {
     return new Coordinate(this);
   }
+  
+  /**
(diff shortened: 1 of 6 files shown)


modules/core/src/main/java/org/locationtech/jts/geom/Coordinate.java
```diff
@@ -400,6 +400,15 @@ public Object clone() {
   public Coordinate copy() {
     return new Coordinate(this);
   }
+  
+  /**
+   * Create a new Coordinate of the same type of this Coordinate, but with no values.
+   * 
+   * @return a new Coordinate
+   */
+  public Coordinate createEmptyCopy() {
```
dr-jts, line 409: Would prefer this called `create`, for brevity

halset: Fixed

halset:
Replaced by #637

PR #644, opened 2020-11-26
Remove OverlapUnion optimization from CascadedPolygonUnion

dr-jts:
The OverlapUnion optimization in fact reduces performance, so is being removed.


jnh5y:
@dr-jts does this imply any change in client code?  E.g., should clients be doing something different to union polygons efficiently?

dr-jts:
@jnh5y only if they know an algorithm that will union polygons more efficiently than `CascadedPolygonUnion`. In which case I hope they will share it!

Issue #629, opened 2020-11-05
buffer(0) returns wrong geom


i23098:
I have an isochrone returned by [Valhalla](https://github.com/valhalla/valhalla):

[valhalla-isochrone-json.txt](https://github.com/locationtech/jts/files/5493851/valhalla-isochrone-json.txt)

Since sometimes the geometries are a bit invalid, I use JTS .buffer(0) to clean it.

Initially, it's a big poligon:
`POLYGON ((-91.483063 45.088623, -91.482628 45.085438, -91.479324 45.085125, -91.478821 45.082005, -91.47226 45.076057, -91.470383 45.072361, -91.468643 45.071594, -91.465019 45.072636, ...
`

but the clean version is not ok, nor in shape, nor in location itself

`POLYGON ((-91.48652270044593 45.088637411415355, -91.486664 45.088638, -91.486732 45.089169, -91.48652270044593 45.088637411415355))`

I'm using jts-core-1.17.1

dr-jts:
This is expected behaviour, unfortunately.  The `buffer(0)` trick is just a heuristic that mostly works - but not always. In particular, if the invalidity is caused by a "figure-8"/self-intersection at the highest point of the polygon, then the buffer result keeps the portion of the polygon at the high point - even if that is a very small part of the area.  That's whats happening here.

dr-jts:
Thinking about this has given me an idea for a further heuristic to counter this.  If the polygon is flipped along the X axis, then `buffer(0)` will return the larger part, which can then be flipped back.  To detect that the "wrong" part of the polygon is returned by the first buffer, a couple of heuristics come to mind, such as:
* if the area is less than half of the original area
* if the number of vertices in the result is less than half of the original vertex count

Of course, if the polygon _also_ happens to contain a self-intersection at the bottom, then this won't work. But there are two other transformations that can be tried: flipping along the XY axis, and along the -XY axis.  One of these should work except for highly pathological polygons.

I will code this up in a TestBuilder function and post a link to the code.

i23098:
Meanwhile, I found similar issue in stackoverflow: https://stackoverflow.com/questions/31473553/is-there-a-way-to-convert-a-self-intersecting-polygon-to-a-multipolygon-in-jts

dr-jts:
> Meanwhile, I found similar issue in stackoverflow: https://stackoverflow.com/questions/31473553/is-there-a-way-to-convert-a-self-intersecting-polygon-to-a-multipolygon-in-jts

[...]

Issue #498, opened 2019-11-24
SimplifyDP produces wrong result for polygon at specific distance


dr-jts:
As noted in [GEOS 1004](https://trac.osgeo.org/geos/ticket/1004), the following polygon produces a VERY incorrect result for simplify tolerance = 0.0036.

Interestingly, using tolerance = 0.00365 produces a reasonable result.

Input:
```
[...]

PR #655, opened 2020-12-21
Fix buffer to use largest enclosed area for invalid rings

dr-jts:
The buffer algorithm has a long-standing issue where a polygon with a small self-intersecting lobe at the top of the polygon gets reduced to just the small lobe, rather than the small lobe being removed.  See for example #629.  This issue can also occur in hole rings.  

image

This issue also affects JTS algorithms which use the `buffer(0)` trick to produce valid polygons (such as `DouglasPeuckerSimplifier`, `VWSimplifier` and `Densifier`).  For example, #498. 

The issue occurs because `buffer` has to determine the interior of a ring to know which side to build the buffer offset curve on (which in the case of a zero-width buffer is just the ring linework). Currently buffer uses the `Orientation.isCCW` test to do this.  For efficiency this algorithm determines ring orientation by checking only the line segments incident on the uppermost vertex of the ring.  For a valid ring (where the linework does not cross itself) this works fine.  However, in a self-crossing ring (aka a "bow-tie") which lobe is chosen as the interior is not well-defined.  The uppermost vertex approach can end up choosing a very small lobe to be the "interior" of the polygon.  

The fix is to use an orientation test which takes into account the entire ring.  This is provided by the **Signed-Area Orientation** test.  This effectively determines the orientation based on the largest area enclosed by the ring.  This corresponds much more closely to user expectation based on visual assessment.  It also minimizes the change in area and (usually) extent.
image

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/algorithm/Orientation.java
+++ b/modules/core/src/main/java/org/locationtech/jts/algorithm/Orientation.java
@@ -21,7 +21,8 @@ import org.locationtech.jts.geom.impl.CoordinateArraySequence;
  * Orientation is a fundamental property of planar geometries 
  * (and more generally geometry on two-dimensional manifolds).
  * <p>
- * Orientation is notoriously subject to numerical precision errors
+ * Determining triangle orientation 
+ * is notoriously subject to numerical precision errors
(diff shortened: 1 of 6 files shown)


dr-jts:
Fixes #629.
Fixes #498.

Issue #663, opened 2021-01-12
Question: Is there a way to use PostGIS WKB format in WKBWriter?


hisener:
The WKB representation I get from `WKBWriter` is different than what I have in PostGIS.

For example, if I run the queries below on PostGIS and use the output (`TEST_WKB_POLYGON` and `TEST_WKB_MULTIPOLYGON`) to read and then write again using JTS, the test for `TEST_WKB_MULTIPOLYGON` fails. IIUC the difference is because `WKBWriter` adds another `0020 E610` to the output for SRID of the polygon inside the multi polygon.

It's problematic because we use the output of `WKBWriter#write` [in Snowflake](https://docs.snowflake.com/en/sql-reference/functions/try_to_geography.html) and get an error. But they do support the PostGIS format so if there is a way to produce that format, that would solve my problem.

`
SELECT st_geomfromtext('SRID=4326;POLYGON((4.880504608154297 52.385868017082856, 4.865398406982422 52.358828590091186, 4.909172058105469 52.35065062061594, 4.927196502685547 52.375179989586506, 4.880504608154297 52.385868017082856))');
`

`
SELECT st_geomfromtext('SRID=4326;MULTIPOLYGON(((4.880504608154297 52.385868017082856, 4.865398406982422 52.358828590091186, 4.909172058105469 52.35065062061594, 4.927196502685547 52.375179989586506, 4.880504608154297 52.385868017082856)))');
`

```java
private static final GeometryFactory GEOMETRY_FACTORY =
        new GeometryFactory(new PrecisionModel(), 4326);

private static final String TEST_WKB_POLYGON =
        "0103000020E6100000010000000500000000000000A3851340BCF8881F64314A40000000002B761340DDA76118EE2D4A4000000000FEA213400CEF991EE22C4A400000000073B513409FB1DCE505304A4000000000A3851340BCF8881F64314A40";
private static final String TEST_WKB_MULTIPOLYGON =
        "0106000020E6100000010000000103000000010000000500000000000000A3851340BCF8881F64314A40000000002B761340DDA76118EE2D4A4000000000FEA213400CEF991EE22C4A400000000073B513409FB1DCE505304A4000000000A3851340BCF8881F64314A40";

@ParameterizedTest
@ValueSource(strings = {TEST_WKB_POLYGON, TEST_WKB_MULTIPOLYGON})
void readWrite(String wkb) throws ParseException {
[...]

PR #664, opened 2021-01-15
Switch to compact WKB format

hisener:
Fixes https://github.com/locationtech/jts/issues/663.

I have ported [GEOS WKBWriter](https://github.com/libgeos/geos/blob/master/src/io/WKBWriter.cpp#L213-L214) and changed `WKBReader` to use the parent's SRID in sub geometries.

Added tests to verify that
* `WKBWriter` writes compact WKB format
* `WKBReader` can read from both format types
* Parsed sub geometry types have the SRID of the parent in case the given WKB is compact.


modules/core/src/main/java/org/locationtech/jts/io/WKBReader.java
```diff
@@ -202,7 +202,6 @@ else if(isStrict)
 
     // determine if SRIDs are present
     hasSRID = (typeInt & 0x20000000) != 0;
-    int SRID = 0;
```
hisener, line 205: So that we use the class field. This might be problematic in case a geometry collection contains mixed SRIDs as @dr-jts pointed [here](https://github.com/locationtech/jts/issues/663#issuecomment-759626055).

dr-jts: Yes, definitely problematic.  This should just use the SRID of the top-most geometry (or 0 if not set).  JTS does not support mixed SRIDs in collections.

hisener: Shouldn't we throw a `ParseException` exception in that case instead of using the SRID of top-most geometry?

dr-jts: It's nice to allow reading as many WKB inputs as possible.   Even if they are invalid, they can only be fixed if they are able to be read.

To preserve the most information, the SRIDs should always be read, and then applied to any missing SRIDs in child geometries.  This will require passing the most recent SRID down the call stack as a parameter.

hisener: Added a commit, PTAL.

dr-jts: Looks good.

dr-jts (approved):
This looks good to me.

hisener:
Can we include this change in the next release, and is there a planned date for it?

dr-jts:
Yes, I will commit this.  

There should be a new release by the middle of 2021, I think.  Is that soon enough for you?

hisener:
Is it going to be part of the next major release? I was hoping for a minor in the next couple of weeks. 🙈

dr-jts:
We might be able to get out a minor release this quarter.  A minor release takes just as much work as a major release.

hisener:
Ack! This quarter would be nice. Thanks! :pray:

PR #672, opened 2021-01-21
PackedCoordinateSequence separated arrays variant

bjornharrtell:
This PR implements a third variant of the PackedCoordinateSequence implementations named Double2 for lack of a better name.

This variant explicitly uses three arrays (XY, Z and M respectively) instead of a stride based single array as the other two implementations do.

The motivation of adding this variant is to support a more direct and efficient way to interop with memory models of this layout. Popular examples of this memory model are GDAL and QGIS. It's also the memory model I chose for my own FlatGeobuf format and is the main reason for me wanting this (to be able to more efficiently use FlatGeobuf with JTS or ports of it).


bjornharrtell:
@dr-jts can I get your opinion on this?

jagill:
Aside: Have you had success with this memory format?  I was trying it out recently, in the hopes I could vectorize some operations, but I found significant overhead that swamped any SIMD benefits.

bjornharrtell:
@jagill when designing FlatGeobuf I was opting between a single array with dimensions as stride or the one I described here. I couldn't find any pro/cons with either and the choice was ultimately decided by that GDAL has that model and that 3D wasn't a primary goal at the time. I'm perhaps regretting it now but I still don't feel there is a clear winner.

I do not have any experience with SIMD. Which memory model would you say is ideal?

jagill:
I don't know which model is ideal.  A flat strided array `[x, y, z, x, y, z, ...]` so far has been fastest for me.  In Java, it is significantly faster than an array of points `[Point(x, y, z), Point(x, y, z), ...]` (this is not surprising).  In theory, if you have separate arrays for each dimension `[x, x, ...], [y, y, ...], [z, z, ...]` you can:
1. Be more agnostic as to whether you have `z` or `m` dimensions, and
2. use SIMD more easily, because all your `x`s are in one place.

But so far I have not succeeded in making separate arrays faster than a strided array.

bjornharrtell:
@jagill thanks for your thoughts, they give me some comfort that perhaps the memory model I've chosen isn't bad. :)

bjornharrtell:
I just remember another rationale for the paired XY layout and it's basically that X is more or less useless without Y but other dimensions and/or measures are more standalone.

PR #686, opened 2021-02-09
Parse null and empty coordinates as empty geometry

JamesRTaylor:


dr-jts:
Are you able to fill out an ECA?

JamesRTaylor:
I tried a couple of times to fill it out and submit it, but it seems that it's not recognizing my email address. I'll try closing this PR and reopening to see if it makes a difference.

Issue #684, opened 2021-02-08
GeoJSONReader should read empty coordinates array as an empty Geometry


JamesRTaylor:
For Point and Polygon, an empty coordinate array does not deserialize as an empty geometry, while it does for other types.  For example, 
   * {"type":"Point", "coordinates":[]} deserializes as POINT (0 0)
   * {"type":"Polygon", "coordinates"[]} causes an exception to be raised

Would be ideal if in both these cases, an empty point and an empty polygon are deserialized instead.

JamesRTaylor:
I'd be happy to contribute a fix for this if you're open to it.

dr-jts:
Is this for the `GeoJSONReader` ?

Would be happy to accept a PR for this.

JamesRTaylor:
Yes, exactly.

Would you be up the related fix for https://github.com/locationtech/jts/issues/411 so that empty geometries for atomic types roundtrip?

dr-jts:
Round-tripping would be nice, for sure.  I can put that on the todo list, or feel free to submit a PR.

dr-jts:
It would be nice to handle both an empty array and `null`.

JamesRTaylor:
FYI, I submitted a PR for this, @dr-jts.

dr-jts:
Great, thanks.  I just merged it.

JamesRTaylor:
Thanks for merging, @dr-jts. Do you know when the next release will occur yet?

dr-jts:
No schedule for a release yet.  But the improvements are piling up fast, so will be soon-ish.

PR #687, opened 2021-02-09
Parse null and empty coordinates as empty geometry

JamesRTaylor:
Fix for https://github.com/locationtech/jts/issues/684


JamesRTaylor:
I'm not sure what the problem is with the CLA. Maybe you need to do something on your end?

Issue #681, opened 2021-02-04
[Clarification] Support for GeoJson:2016 polygon ring orientation?


ruibritopt:
I'm looking for clarification on support for the latest GeoJSON spec - [RFC-7946](https://tools.ietf.org/html/rfc7946). It seems to me that the `GeoJsonReader` and `GeoJsonWriter` in the `jts.io` module only support the [previous spec.](https://geojson.org/geojson-spec). 

If it does in fact support it, I would appreciate if you could point me to any usage or documentation. If they do not support it, is there any plan to do so? Or to your knowledge another helper library that might be used instead?

Thank you and below there's an example of a GeoJSON polygon failing the right-hand rule, a newly introduced rule on the latest GeoJSON spec, and how I came to seek this clarification. https://geojsonlint.com/ can be used to prove my claim.
```
{
  "type": "Feature",
  "properties": {},
  "geometry": {
    "type": "Polygon",
    "coordinates": [
      [
        [
          9.92806,
          53.557383
        ],
        [
          9.929417,
          53.55745
        ],
        [
          9.92965,
          53.557322
        ],
        [
          9.929565,
          53.557063
        ],
        [
          9.927875,
          53.557012
        ],
        [
          9.92806,
          53.557383
        ]
      ]
    ]
  }
}
```

dr-jts:
Yes, it appears that the most recent GeoJSON spec is not quite supported, due to its requirement for CCW orientation of. polygon rings.  Although I notice that the spec does say:

< For backwards compatibility, parsers SHOULD NOT reject Polygons that do not follow the right-hand rule.

Unfortunately, JTS was original implemented before the OGC specification included a requirement for polygon ring orientation.  It chose to orient polygon shells CW.  Not sure it's easy to change that now.

A couple of possibilities are:

* provide a `orientCCW` operation, which can be used when emitting GeoJSON
* and/or have the `GeoJSONWriter` force polygons to use the CCW orientation

ruibritopt:
Hi @dr-jts, I've opened a PR to try and cover this aspect of the specification.
I would be open to any edit or comment from your or any of the other maintainers. Thank you.

PR #694, opened 2021-02-15
Allow GeoJSONWriter to force polygons to be CCW

ruibritopt:
A best-effort PR to enforce the right hand rule for polygon ring orientation in the latest GeoJson specification and discussed in issue #681

Signed-off-by: Rui Brito <ruibritopt@gmail.com>


dr-jts:
I would prefer to not bake this into the GeoJSONWriter.  Instead could create a ForceRHR function to process each geometry beforehand.

ruibritopt:
@dr-jts That is a valid point. If I understand your comment correctly, we would end up with a function like `public static Geometry forceHRH(Geometry geometry) { ... }`. 
Would this function have to be called by the library clients before invoking `write` or would we extend the `write` function as well like eg: `public String write(Geometry geometry, Boolean forceRHR) { ... }`?

dr-jts:
Actually it will be more efficient if the orientation is modified on a per-ring basis in the writer.  In this case, it would be best to not change the current behaviour.  So an option can be added to the `GeoJSONWriter` (called say `setForceRHR`) which will check and reorient polygon rings as required.  The option can be disabled by default.

ruibritopt:
@dr-jts I extended the GeoJsonWriter class to have an attribute `isRHREnforced`, very similar to how the `isEncodeCRS` is set up. If I understood you correctly this was your recommendation.
The reason why the transformation function are outside the GeoJsonWriter class is because I believe the same logic could be used on the GeoJsonReader side, for the clients that would like to store geometries under the specification. 

Please let me know what you think. I'm willing to make any changes necessary. Also, if the idea of implementing the same logic on the reader side is appealing, let me know and I will make another PR. Thank you

modules/io/common/src/main/java/org/locationtech/jts/io/geojson/TransformationUtils.java
```diff
@@ -0,0 +1,85 @@
...
+     * @param ring the LinearRing, a constraint specific to Polygons
+     * @param isExteriorRing true if the LinearRing is the exterior polygon ring, the one that defiens the boundary
+     * @return LinearRing under the Right Hand Rule specifications
+     */
+    public static LinearRing enforceRightHandRuleOrientation(LinearRing ring, boolean isExteriorRing) {
+        final boolean isRingClockWise = !Orientation.isCCW(ring.getCoordinateSequence());
+
+        final LinearRing rightHandRuleRing;
+        if (isExteriorRing) {
+            rightHandRuleRing = isRingClockWise? ring.reverse() : ring;
```
dr-jts, line 79: need to copy the ring if it is unchanged, to avoid aliasing errors

modules/io/common/src/main/java/org/locationtech/jts/io/geojson/TransformationUtils.java
```diff
@@ -0,0 +1,85 @@
...
+        } else {
+            rightHandRuleRing = isRingClockWise? ring : ring.reverse();
```
dr-jts, line 81: copy here too

modules/io/common/src/main/java/org/locationtech/jts/io/geojson/TransformationUtils.java
```diff
@@ -0,0 +1,85 @@
+package org.locationtech.jts.io.geojson;
+
+import org.locationtech.jts.algorithm.Orientation;
+import org.locationtech.jts.geom.*;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class TransformationUtils {
```
dr-jts, line 9: How about calling this `OrientationTransformer`?

modules/io/common/src/main/java/org/locationtech/jts/io/geojson/TransformationUtils.java
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * Transforms a geometry using the Right Hand Rule specifications defined
+     * in the latest GeoJSON specification.
+     * See <a href="https://tools.ietf.org/html/rfc7946#section-3.1.6">RFC-7946 Specification</a> for more context.
+     *
+     * @param geometry to be transformed
+     * @return Geometry under the Right Hand Rule specifications
+     */
+    public static Geometry enforceRightHandRuleOrientation(final Geometry geometry) {
```
dr-jts, line 19: just call this `transformRightHandRule`.

dr-jts:
> @dr-jts I extended the GeoJsonWriter class to have an attribute `isRHREnforced`, very similar to how the `isEncodeCRS` is set up. If I understood you correctly this was your recommendation.

Yes, that looks good.

> The reason why the transformation function are outside the GeoJsonWriter class is because I believe the same logic could be used on the GeoJsonReader side, for the clients that would like to store geometries under the specification.

It's good to have the RHR conversion factored out.  But I don't think it's necessary to bake it into `GeoJSONReader`.  Clients can simply transform geometries if they want to after reading them.

dbaston:
Maybe consider avoiding the term RHR, since it is used by other software such as PostGIS to mean the opposite of what it means here.

dr-jts:
> Maybe consider avoiding the term RHR, since it is used by other software such as PostGIS to mean the opposite of what it means here.

This is a good point.  To be fair, the GeoJSON spec uses the term "right-hand rule", and defines that to mean "shell is CCW".  That seems to be the most commonly accepted definition of RHR.  The PostGIS usage is opposite, and is deprecated (discussion [here](https://trac.osgeo.org/postgis/ticket/3689)). (SQL Server seems to have the [same problem](https://docs.microsoft.com/en-us/sql/relational-databases/spatial/spatial-data-types-overview?view=sql-server-ver15)).

But it's good to avoid stepping into that minefield.  Also, two important design criteria for an API are consistency and parsimony of terminology.  The term "CCW" is already used for this concept in JTS, and it is unambiguous.  So this PR should change to using that acronym.   So use `isEnforceCCW`, `setEnforceCCW`, `transformCCW`, etc.

The GeoJSONWriter Javadoc can point out that in the GeoJSON context "right-hand rule" = CCW.

modules/io/common/src/main/java/org/locationtech/jts/io/geojson/TransformationUtils.java
ruibritopt: I followed this recommendation 👍

modules/io/common/src/main/java/org/locationtech/jts/io/geojson/TransformationUtils.java
ruibritopt: I followed this recommendation 👍

modules/io/common/src/main/java/org/locationtech/jts/io/geojson/TransformationUtils.java
ruibritopt: I believe the change I made resolves this concern.

ruibritopt:
Thank you @dr-jts @dbaston. I've implemented your suggestions. Please let me know if it fits your views. Thank you.

modules/io/common/src/main/java/org/locationtech/jts/io/geojson/GeoJsonWriter.java
```diff
@@ -88,6 +89,16 @@ public void setEncodeCRS(boolean isEncodeCRS) {
     this.isEncodeCRS  = isEncodeCRS;
   }
   
+  /**
+   * Sets whether the GeoJSON should be output following counter-clocked orientation aka Right Hand Rule defined in RFC7946
```
dr-jts, line 93: should be "counter-clockwise".

modules/io/common/src/main/java/org/locationtech/jts/io/geojson/OrientationTransformer.java
```diff
@@ -0,0 +1,85 @@
+package org.locationtech.jts.io.geojson;
+
+import org.locationtech.jts.algorithm.Orientation;
+import org.locationtech.jts.geom.*;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class OrientationTransformer {
```
dr-jts, line 9: Add class Javadoc.  Can be something simple like

> Utilities to modify the ring orientation of polygonal geometries.

dr-jts:
Looking good. Just a couple of small things to fix, and it will be good to merge.

modules/io/common/src/main/java/org/locationtech/jts/io/geojson/GeoJsonWriter.java
ruibritopt: Silly mistake, thank you for spotting it.

modules/io/common/src/main/java/org/locationtech/jts/io/geojson/OrientationTransformer.java
ruibritopt: Took your recommendation verbatim

modules/io/common/src/main/java/org/locationtech/jts/io/geojson/GeoJsonWriter.java
```diff
@@ -88,6 +89,16 @@ public void setEncodeCRS(boolean isEncodeCRS) {
     this.isEncodeCRS  = isEncodeCRS;
   }
   
+  /**
+   * Sets whether the GeoJSON should be output following counter-clockwise orientation aka Right Hand Rule defined in RFC7946
+   * See <a href="https://tools.ietf.org/html/rfc7946#section-3.1.6">RFC 7946 Specification</a> for more context.
+   *
+   * @param enforceCCW true if the GeoJSON should be output following the RFC7946 counter-clockwise orientation aka Right Hand Rule
+   */
+  public void setEnforceCCW(boolean enforceCCW) {
```
dr-jts, line 98: One final request.  In the interest of brevity, can this be changed to `setForceCCW` and `isForceCCW`?

ruibritopt: Done and done. Don't hold back on the requests, they have all been on the money and I'm a big fan of things done right 😄👍👍

PR #699, opened 2021-03-08
Get GeometrySnapper tolerance from largest geometry element

gwalbran:
…lections

Currently the snap tolerance has been calculated from the geometry envelope.
But this can result in a large tolerance for geometry collections where the
envelope is large but that contain parts are much smaller.

In the geometry collection case we can calculate the tolerance that ought to
be applied to each part, and use the maximum of those values.

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/operation/overlay/snap/GeometrySnapper.java
+++ b/modules/core/src/main/java/org/locationtech/jts/operation/overlay/snap/GeometrySnapper.java
@@ -19,6 +19,7 @@ import org.locationtech.jts.geom.Coordinate;
 import org.locationtech.jts.geom.CoordinateSequence;
 import org.locationtech.jts.geom.Envelope;
 import org.locationtech.jts.geom.Geometry;
+import org.locationtech.jts.geom.GeometryCollection;
 import org.locationtech.jts.geom.LineString;
 import org.locationtech.jts.geom.Polygonal;
 import org.locationtech.jts.geom.PrecisionModel;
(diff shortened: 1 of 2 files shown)


bjornharrtell:
@gwalbran have you seen/tried the new overlay implementation at https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/operation/overlayng/OverlayNG.java ? I'm guessing it can handle your case.

dr-jts:
This fix makes sense to me.  

However, as noted above, as of JTS 1.18 the old overlay code (which uses GeometrySnapper) has been superseded by the OverlayNG implementation.  It would be worth trying the geometry failure cases against that code.

Is it possible for you to post some or all of the overlay failure cases you've encountered?

gwalbran:
I hadn't tried with OverlayNG, but that does fix the problem. Thanks for the pointer

The attached files have the geoms that caused a TopologyException when doing a difference.
e.g difference1lhs.difference(difference1.rhs)

Difference1 is actually a subset of difference2.

[difference1lhs.txt](https://github.com/locationtech/jts/files/6103615/difference1lhs.txt)
[difference1rhs.txt](https://github.com/locationtech/jts/files/6103616/difference1rhs.txt)
[difference2lhs.txt](https://github.com/locationtech/jts/files/6103617/difference2lhs.txt)
[difference2rhs.txt](https://github.com/locationtech/jts/files/6103618/difference2rhs.txt)

dr-jts:
> I hadn't tried with OverlayNG, but that does fix the problem. Thanks for the pointer

Given that, there's not really a need to apply this PR.

Thanks for the data.  As expected, it contains a lot of nearly-coincident lines, which is the usual cause of non-robustness.  The new OverlayNG snapping algorithm handles this better than the old `GeometrySnapper` (which was a very blunt instrument).

dr-jts:
I'm going to close this, since the old overlay code is basically obsolete.

Issue #652, opened 2020-12-13
RFC: Fixing invalid geometry


dr-jts:
It is desirable to have a way to convert an invalid geometry into a valid one that preserves the character of the input (to some extent).  This allows the fixed geometry to be used in JTS operations.  

See  [`GeometryFixer` prototype code](https://github.com/dr-jts/jts/blob/add-geometry-fixer/modules/core/src/main/java/org/locationtech/jts/geom/util/GeometryFixer.java).

## Goals

- Allow converting any geometry to one which is valid, preserving as much as possible of the original appearance

### Non-Goals

- Preserving all vertices in input - some vertices may be dropped to produce a "natural" repair
- Precision reduction - can be done subsequent to processing
- Normalization - can be done subsequent to processing
- Simplification - can be done before processing

## Motivating Cases

Descriptions use following terminology:  "touch" means touches in a point, "adjacent" means touches along a line.

See [XML file for example data test cases](https://github.com/locationtech/jts/blob/master/modules/tests/src/test/resources/testxml/misc/TestInvalidA.xml).

- Bow-tie polygon
- Bow-tie polygon with multiple crossings
- Checkerboard polygon
- Polygon with zero area and collapses to a line
- Polygon with zero area and collapses to a point
- Polygon with portion of the shell boundary which collapses to a line (internal or external)
- Polygon with portion of the shell boundary which collapses to a line and encloses area
- Polygon which overlaps itself
- Polygon with linework separating it into two adjacent regions (i.e cutting it in two)
- Polygon with collapsed linework that encloses area (i.e. with an inverted boundary that is adjacent)
- Polygon with overlapping holes (intersecting in points)
- Polygon with overlapping holes (intersecting along edges)
- Polygon with adjacent holes
- Polygon with holes forming a separated region
- Polygon with hole adjacent to shell
- Polygon with holes adjacent to shell and separating a region
- Polygon with hole overlapping shell 
- Polygon with hole surrounding shell
- Polygon with bowtie hole overlapping shell - intersection inside
- Polygon with bowtie hole overlapping shell - intersection outside
- Polygon with nested holes - disjoint
- Polygon with nested holes - touching
- Polygon with nested holes - adjacent
- MultiPolygon with overlapping elements
- MultiPolygon with adjacent elements
- MultiPolygon with polygon with hole overlapping shell and other polygon
- GeometryCollection with multiple invalid elements
- Line that collapses to a point
- Geometry with NaN ordinates

## Issues

[...]

PR #704, opened 2021-04-06
Add GeometryFixer class

dr-jts:
Adds the class `GeometryFixer` to support repairing invalid geometry into valid geometry.

See the requirements specification in #652.


hirenkp2000:
Using GeometryFixer, how can one achieve `method=linework` algorithm as supported by [ST_MakeValid](https://postgis.net/docs/ST_MakeValid.html)? It seems GeometryFixer defaults to `method=structure` algorithm.
Is there a way to specify which algorithm to be invoked? Or any other alternate way in JTS to achieve `method=linework`

Issue #472, opened 2019-09-07
Large negative buffer does not return empty polygon


dr-jts:
Computing a negative buffer on some polygons returns an "inverted" polygon rather than an empty one.

Example:
```
POLYGON ((666360.09 429614.71, 666344.4 429597.12, 666358.47 429584.52, 666374.5 429602.33, 666360.09 429614.71))
```
Buffer of -15 should be empty, but returns:
```
POLYGON ((666361.4831382894 429593.7375842848, 666353.0817850837 429600.9554019591, 666357.3305565341 429605.6759646248, 666365.55909509 429598.30712412694, 666361.4831382894 429593.7375842848))
```

Originally reported on [GIS StackExchange](https://gis.stackexchange.com/questions/333026/postgis-and-qgis-inner-buffer-issue). See also [GEOS 1101](https://trac.osgeo.org/geos/ticket/1101) and [QGIS 41696](https://github.com/qgis/QGIS/issues/41696).

dr-jts:
One heuristic for dealing with this is to compute the radius of the circle with the same area as the geometry.  This is an upper bound on the buffer distance which will erode the polygon completely (i.e. buffer distances less than the radius should produce an empty buffer).  

For example, the effective radius of the geometry above is 11.970122020294427.  So the distance of -15 should return an empty buffer result.  

However, this bound is not sufficient.  A buffer distance of -11.9 still produces an erroneous non-empty polygon. The radius of the MaximumInscribedCircle is 9.475544793980276.  Distance = -9.5 produces an empty result, as expected.

dr-jts:
Here's some more failure cases:

**Pentagon** (max inscribed circle radius = 8.2, buffer of -9 is non-empty)
```
POLYGON ((6 20, 16 20, 21 9, 9 0, 0 10, 6 20))
```
**Hexagon** (max inscribed circle radius = 9.599, buffer of -9.6 is non-empty)
```
POLYGON ((16 20, 6 20, 0 10, 6 0, 16 0, 23 10, 16 20))
```

dr-jts:
My hypothesis is that this only happens for relatively regular polygons with a small number of vertices (say <= 9).

Given this, it is not too expensive to compute the buffer curve and then check if every vertex of the curve is closer to the shell than the provided buffer distance.  If this is the case, the generated buffer is actually empty.

Note that each element of input MultiPolygons must be processed indivdually.

dr-jts:
Fixed by #706.

PR #706, opened 2021-04-12
Fix buffer to remove inverted ring curves

dr-jts:
This is a fix for the following buffer situation. It can happen that the sides of a small, convex polygon produce offset segments which all cross one another to form a curve with inverted orientation. This happens at buffer distances slightly greater than the distance at which the buffer should disappear. The inverted curve will produce an incorrect non-empty buffer (for a shell) or an incorrect hole (for a hole). It must be discarded from the set of offset curves used in the buffer.

Heuristics are used to reduce the number of cases which area checked, for efficiency and correctness.
  
This is a fix for the issue described in #472.


dr-jts:
Fixes #472

See also [GEOS 1101](https://trac.osgeo.org/geos/ticket/1101) and [QGIS 41696](https://github.com/qgis/QGIS/issues/41696).

Issue #711, opened 2021-04-21
JAXB serialization to/from XML fails for Coordinates


seanrl:
If you try to serialize a jts.Coordinate to/from XML, it will fail because it will find 2 options for the x, y, and z values (one public field, and one public getter).  An easy solution to this is to add `@XmlTransient` to the public fields (which really should just be private at some point, but I understand the desire for backwards compatibility).  Like so:

```
/**
   * The x-ordinate.
   */
  @XmlTransient
  public double x;

  /**
   * The y-ordinate.
   */
  @XmlTransient
  public double y;

  /**
   * The z-ordinate.
   * <p>
   * Direct access to this field is discouraged; use {@link #getZ()}.
   */
  @XmlTransient
  public double z;
  ```
  
  It would be great if we could get this in a future release.

jnh5y:
@jodygarnett do you know anything about this annotation?

bjornharrtell:
It seems like sound rationale to use this annotation and superficially it looks safe to use all the way back to at least Java 8, see https://docs.oracle.com/javase/8/docs/api/javax/xml/bind/annotation/XmlTransient.html.

pomadchin:
But it is deprecated in JDK 9 and removed in JDK 11.

bjornharrtell:
@pomadchin wow nice catch. Seems that it has since been removed and the recommended replacement is to depend on external implementation of it definitely making it less clear if this is a good idea to use here.

mprins:
it is not a good idea to add `@XmlTransient` as I noted on the PR yesterday: https://github.com/locationtech/jts/pull/713#pullrequestreview-641461947

seanrl:
@mprins Closing that PR, I agree with you.  Looking forward to the switch over to the JavaBean pattern eventually, and for now I'll just keep making internal releases for my needs.

PR #713, opened 2021-04-21
Resolve issue #711

seanrl:
This is a very minor change that will resolve issue #711 while maintaining backwards compatibility.

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/geom/Coordinate.java
+++ b/modules/core/src/main/java/org/locationtech/jts/geom/Coordinate.java
@@ -17,6 +17,8 @@ import java.util.Comparator;
 import org.locationtech.jts.util.Assert;
 import org.locationtech.jts.util.NumberUtil;
 
+import javax.xml.bind.annotation.XmlTransient;
+
 
 /**
  * A lightweight class used to store coordinates on the 2-dimensional Cartesian plane.
@@ -77,11 +79,13 @@ public class Coordinate implements Comparable<Coordinate>, Cloneable, Serializab
   /**
    * The x-ordinate.
    */
+  @XmlTransient
   public double x;
   
   /**
    * The y-ordinate.
    */
+  @XmlTransient
   public double y;
   
   /**
@@ -89,6 +93,7 @@ public class Coordinate implements Comparable<Coordinate>, Cloneable, Serializab
    * <p>
    * Direct access to this field is discouraged; use {@link #getZ()}.
    */
+  @XmlTransient
   public double z;
 
   /**


modules/core/src/main/java/org/locationtech/jts/geom/Coordinate.java
```diff
@@ -17,6 +17,8 @@
 import org.locationtech.jts.util.Assert;
 import org.locationtech.jts.util.NumberUtil;
 
+import javax.xml.bind.annotation.XmlTransient;
```
mprins, line 20: This makes  JTS depend on JAXB / JAX-RI which is a JEE module as of Java 11 (it is still part of the Java 8 API though). Not sure that is a good idea.

jnh5y: @mprins Good info.  Do you have any insight for alternatives to addressing the issue that @seanrl has raised here: https://github.com/locationtech/jts/issues/711 ?

mprins: The proper solution would be to change the API of Coordinate and use a JavaBean pattern only (having only getters/setters for x,y,z); or alternatively a class member pattern only (having public fields/members only for x,y,z), or perhaps changing the visibility of x,y,z to les than `public`, not sure if eg `protected` would suffice. (NB adding this annotation to these public fields strictly incurs an API change as well)

The problem is that JAXB gets confused because it finds 2 ways to serialize x,y,z to XML and has no way to choose; the annotation exists to work around poor (or in case of JTS historic) design. `Coordinate` (or anything else in JTS) was ever designed to be an `@XMLType`. The workaround outside JTS is probably to use an `@XmlJavaTypeAdapter` I think, or perhap setting up the `@XmlAccessorType` in the container class.

personally I'd go for the bean pattern, always.

seanrl: I agree, the correct long-term solution is to make the x, y, and z fields private.  Of course, the XmlTransient annotation is safe as long as this is Java 8 based, so perhaps the low-hanging fruit here is to use the XmlTransient for now and deprecate the public fields, opting to switch to the JavaBean pattern at such a time when we migrate off of Java 8.

mprins: There are many downstream projects relying on JTS that use Java 11 and do not have a JAXB dependency. 
IMO introducing the JAXB dependency + a public API change -which adding the annotation is- is worse than changing the API to follow the JavaBean pattern.

seanrl: @mprins That's a good point.  I'll close the PR, and just keep making my own internal releases until the JavaBean switch

Issue #714, opened 2021-05-03
Z-interpolation forced on overlays


bjornharrtell:
When OverlayNG was introduced in 1.18, missing Z-values became interpolated both in original Overlay and OverlayNG.

Past logic could rely on missing Z-values to be able to supply those values by other means as post processing so it can be seen as a breaking change for the original Overlay and potentially an undesirable feature of OverlayNG.

dr-jts:
> Past logic could rely on missing Z-values to be able to supply those values by other means 

Is this theoretical, or is there actually code broken because of this change?

bjornharrtell:
We have a large suite of custom / data specific geometry validation rules that breaks because of this.

FObermaier:
I wondered why `ElevationModel` doesn't implement some interface that defines the contract for interpolating/getting z-ordinate values. Then it would be possible to implement and use some DEM instead of `ElevationModel` or create an implementation that behaves like in v17.1.

bjornharrtell:
@FObermaier the `ElevationModel` is only part of the problem and only applies to OverlayNG. The interpolation that affects much more is in `RobustLineIntersector`. I agree it could make sense to make it possible to customise `ElevationModel` for OverlayNG but that's another issue.

FObermaier:
But `RobustLineIntersector` could rely on `ElevationModel` or whatever that interface were to be named.

bjornharrtell:
@FObermaier agree that could make sense but requires a more involved effort and is perhaps not the way to tackle the regression.

dr-jts:
> We have a large suite of custom / data specific geometry validation rules that breaks because of this.

Can you provide more detail about what semantics these tests are checking?  E.g. Do they use the absence of Z as a way of determining vertices which have been added? 

I am curious about whether your use case involves your own custom way of computing Z, or you just don't want Z interpolated at all.

dr-jts:
The solutions for setting Z-interpolation on/off in order of increasing complexity seem to be:

1. a `public static` flag on `RobustLineIntersector`
1. a System property used by `RobustLineIntersector`
1. a `public static` flag on `GeometryFactory`
1. a `JTSOptions` singleton containing the flag
1. a field on `GeometryFactory`
1. a field flag on RobustLineIntersector, with options down through the call chain from functions which expose it for setting

The next level of complexity is to replace the flag with a Strategy object to provide custom Z interpolation, with built-ins for linear interpolation and no interpolation.

[...]

PR #715, opened 2021-05-03
Opt in interpolation for RobustLineIntersector

bjornharrtell:
ref https://github.com/locationtech/jts/issues/714

Wanted to see what effect it would have to turn interpolation off by default and it causes a number of expected test failures (OverlayNGZTest, RobustLineIntersectorZTest and GeometryFixerTest).

If we agree that https://github.com/locationtech/jts/issues/714 is indeed a regression then I would have no better current idea than doing something like this, unfortunately.

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/algorithm/RobustLineIntersector.java
+++ b/modules/core/src/main/java/org/locationtech/jts/algorithm/RobustLineIntersector.java
@@ -27,6 +27,22 @@ public class RobustLineIntersector
     extends LineIntersector
 {
 
+  public static String INTERPOLATE_PROPERTY_NAME = "jts.interpolate";
+  public static String INTERPOLATE_PROPERTY_VALUE_TRUE = "true";
+  public static String INTERPOLATE_PROPERTY_VALUE_FALSE = "false";
+  public static boolean INTERPOLATE_DEFAULT = false;
+  private static boolean isInterpolate = INTERPOLATE_DEFAULT;
+  static {
+    setInterpolate(System.getProperty(INTERPOLATE_PROPERTY_NAME));
+  }
+  static void setInterpolate(String interpolateCode) {
+    if (interpolateCode == null) 
+      return;
+    isInterpolate = INTERPOLATE_DEFAULT;
+    if (INTERPOLATE_PROPERTY_VALUE_TRUE.equalsIgnoreCase(interpolateCode) )
+      isInterpolate = true;
+  }
+
   public RobustLineIntersector() {
   }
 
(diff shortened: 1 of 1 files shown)


dr-jts:
Using a global option set via System property seems like the simplest way to allow a choice of Z-interpolation behaviour (although it's the least flexible).

The unit test issues can be handled by either making Z-interpolation the default, or by changing the option setting in the unit test.  The latter approach is already used in the [`GeometryOverlayTest`](https://github.com/locationtech/jts/blob/master/modules/core/src/test/java/org/locationtech/jts/geom/GeometryOverlayTest.java).

bjornharrtell:
@dr-jts I've opted in the interpolation in the test cases where it is the assumption.

bjornharrtell:
I've now reworked this so that the option to Z-interpolate is standalone and affects both RobustLineIntersector and ElevationModel.

modules/core/src/main/java/org/locationtech/jts/geom/ZInterpolating.java
```diff
@@ -0,0 +1,40 @@
+package org.locationtech.jts.geom;
```
jodygarnett, line 1: lacks a header

modules/core/src/main/java/org/locationtech/jts/geom/ZInterpolating.java
```diff
@@ -0,0 +1,40 @@
...
+ * <p>
+ * <ul>
+ * <li><code>jts.zinterpolating=false</code> - (default) do not Z-interpolate
+ * <li><code>jts.zinterpolating=true</code> - Z-interpolate
+ * </ul>
+ * 
+ * @author bjornharrtell
+ *
+ */
+public class ZInterpolating {
```
jodygarnett, line 14: This seems a lot of boilerplate to hold a boolean value decision, can this class take on more responsibility and swap between interpolation approaches as a stratagy object?

modules/core/src/main/java/org/locationtech/jts/geom/ZInterpolating.java
bjornharrtell: fixed

modules/core/src/main/java/org/locationtech/jts/geom/ZInterpolating.java
bjornharrtell: I agree it feel like boilerplate. I have difficulties properly understanding/interpreting what you mean with "take on more responsibility and swap between interpolation approaches as a stratagy object". In either case it's intentionally done this way because it is modeled after the other existing system property option jts.overlay and broken out into a self contained class because it is affecting two very separate things (RobustLineIntersector and ElevationModel), and I wanted to do it in a way that is as non intrusive as possible on the existing source.

I suppose system wide settings could be done with smarter and nicer desig but I don't think this is the place to do such general improvement and restructuring.

bjornharrtell:
Since Z interpolation has now been enabled for so long time it's no longer sensible to regard it as a regression. So I've changed it to be default on in this PR. This means this PR now do not change any behaviour and only allows the interpolation to be opt out.

bjornharrtell:
@jodygarnett and @dr-jts please consider accepting this.

PR #722, opened 2021-05-20
Improve code quality

matteobaccan:
Hi

I have made some code improvement

- Correct boolean compare
- Use try/resource for stream
- Removed unused imports
- Typo error: check identical expression.
- Removed redundant boolean literals
- Public constants must be static final not only final
- toString used on String
- Null checks should not be used with "instanceof" 

all the best
matteo baccan


dr-jts:
These are worthwhile changes, but this is a huge PR that mixes code cleanup with some actual bug fixes.  Can you split it into smaller PRs?

* unused import removal
* redundant code removal
* try block addition
* the couple of things which are actual bugs

matteobaccan:
OK

I can try to do this

matteobaccan:
Hi

you can reject this PR. 
I have made 1 PR for each commit that solves a single problem

ciao
matteo

Issue #721, opened 2021-05-20
Anlge#interiorAngle produces incorrect result


grimsa:
Test case:
![image](https://user-images.githubusercontent.com/8446128/118983666-5737dd80-b985-11eb-8489-0721d6aede51.png)

```
    @Test
    void interiorAngle_rightTriangle() {
        // Triangle WKT: "POLYGON ((1 2, 3 2, 2 1, 1 2))"
        var p1 = new CoordinateXY(1, 2);
        var p2 = new CoordinateXY(3, 2);
        var p3 = new CoordinateXY(2, 1);

        assertEquals(45, Math.toDegrees(Angle.interiorAngle(p1, p2, p3)), 0.01);   // returns 315
        assertEquals(90, Math.toDegrees(Angle.interiorAngle(p2, p3, p1)), 0.01);
        assertEquals(45, Math.toDegrees(Angle.interiorAngle(p3, p1, p2)), 0.01);
    }
```

PR #723, opened 2021-05-20
Fix Angle#interiorAngle method

grimsa:
Fixes https://github.com/locationtech/jts/issues/721

### DIFF
--- a/modules/core/src/test/java/org/locationtech/jts/algorithm/AngleTest.java
+++ b/modules/core/src/test/java/org/locationtech/jts/algorithm/AngleTest.java
@@ -16,8 +16,15 @@ import org.locationtech.jts.geom.Coordinate;
 
 import junit.framework.TestCase;
 import junit.textui.TestRunner;
+import org.locationtech.jts.geom.CoordinateSequenceFactory;
+import org.locationtech.jts.geom.CoordinateSequences;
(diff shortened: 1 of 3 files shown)


modules/core/src/test/java/org/locationtech/jts/algorithm/AngleTest.java
```diff
@@ -97,9 +101,52 @@ public void testNormalize() throws Exception
...
+		// Tests interior angles greater than 180 degrees
+		assertEquals(315, Math.toDegrees(Angle.interiorAngle(p3, p2, p1)), 0.01);
+		assertEquals(270, Math.toDegrees(Angle.interiorAngle(p1, p3, p2)), 0.01);
+		assertEquals(315, Math.toDegrees(Angle.interiorAngle(p2, p1, p3)), 0.01);
+  }
 
+  /**
+   * Tests interior angle calculation using a large number of random triangles
+   */
+  public void testInteriorAngle_randomTriangles() {
```
dr-jts, line 124: I would rather not have lengthy tests like this in the main set of unit tests, to avoid impacting CI performance.

You could reduce the number of tests to say 100 - it can be increased manually if needed.

grimsa: Reduced to 100.

PR #724, opened 2021-05-21
Fix envelope of Voronoi Diagram

narcismadern:
When an envelope containing the entire envelope automatically generated from the original coordinates is set, it's never used and the (smaller) envelope automatically generated is always used instead when creating the QuadEdgeSubdivision

This is causing that large envelopes are never taken into account and results in Voronoi Diagrams smaller than they are supposed to when setting the envelope via 'setClipEnvelope'


dr-jts:
Looks like you need to fill out an Eclipse ECA.  Or I can make this change myself.

narcismadern:
I see that.
Honestly, I don know why I need to do that :-S

If you can make the change yourself, it's OK to me.

Thanks!

dr-jts:
Replaced by #740

PR #730, opened 2021-05-21
Typo error: check identical expression.

matteobaccan:


dr-jts:
This is a bug, but it actually has no effect on the semantics.  (Might be able to simplify the code because of that).

PR #746, opened 2021-06-21
Fix SortedPackedIntervalRtree to be thread-safe

### LINKED ISSUE #747: PreparedPolygon is not thread safe
JavaDoc of PreparedGeometry says "Subclasses are intended to be thread-safe, to allow <code>PreparedGeometry</code>  to be used in a multi-threaded context  (which allows extracting maximum benefit from the prepared state)."
 However, PreparedPolygon is not thread safe, and under concurrent load may give sometimes incorrect result of #contains(Geometry) method. This affect and suggested fix are shown in this PR:  https://github.com/locationtech/jts/pull/746 .

ivanveselovsky-tomtom:
`SortedPackedIntervalRtree` is not thread-safe, because the `init` method is not synchronized.

This also contains some cleanup for `IndexedPointInAreaLocator`.

Fixes #747.

Signed-off-by: veselovs <ivan.veselovsky@tomtom.com>


dr-jts:
Can you describe the core changes which allow concurrency?

modules/core/src/main/java/org/locationtech/jts/index/intervalrtree/SortedPackedIntervalRTree.java
```diff
@@ -132,12 +127,7 @@ private void buildLevel(List src, List dest)
 			}
 		}
 	}
-	
-  private void printNode(IntervalRTreeNode node)
-  {
-    System.out.println(WKTWriter.toLineString(new Coordinate(node.min, level), new Coordinate(node.max, level)));
-  }
-
```
dr-jts, line 140: why is this removed?

modules/core/src/main/java/org/locationtech/jts/algorithm/locate/IndexedPointInAreaLocator.java
```diff
@@ -102,15 +102,13 @@ public int locate(Coordinate p)
   private synchronized void createIndex() {
     if (index == null) {
       index = new IntervalIndexedGeometry(geom);
-      // no need to hold onto geom
-      geom = null;
```
dr-jts, line 106: Is there a reason the geometry can't be freed?

modules/core/src/main/java/org/locationtech/jts/index/intervalrtree/SortedPackedIntervalRTree.java
```diff
@@ -68,7 +66,7 @@ public void insert(double min, double max, Object item)
     leaves.add(new IntervalRTreeLeafNode(min, max, item));
 	}
 	
-  private void init()
+  private synchronized void init()
```
dr-jts, line 69: It looks like this is the key change for concurrency?

modules/core/src/main/java/org/locationtech/jts/algorithm/locate/IndexedPointInAreaLocator.java
ivanveselovsky-tomtom: The reason is that final ``` geom```  field looks clearer, and the performance effect of nulling it is likely negligible.
However, it can be safely nulled, fixed in this way.

modules/core/src/test/java/org/locationtech/jts/geom/prep/PreparedGeometryConcurrencyBugTest.java
```diff
@@ -0,0 +1,67 @@
+package org.locationtech.jts.geom.prep;
```
dr-jts, line 1: This should go into the `test.jts.perf.geom.prep` package, because:
* it requires custom VM settings
* it is not guaranteed to reveal the issue on every run
* it may require significant execution time

modules/core/src/main/java/org/locationtech/jts/index/intervalrtree/SortedPackedIntervalRTree.java
ivanveselovsky-tomtom: Because this is a dead code, method never invoked.

modules/core/src/main/java/org/locationtech/jts/index/intervalrtree/SortedPackedIntervalRTree.java
ivanveselovsky-tomtom: yes.

modules/core/src/test/java/org/locationtech/jts/geom/prep/PreparedGeometryConcurrencyBugTest.java
ivanveselovsky-tomtom: fixed.

ivanveselovsky-tomtom:
> Can you describe the core changes which allow concurrency?

Critical change is synchronizing method  org.locationtech.jts.index.intervalrtree.SortedPackedIntervalRTree#init .
Before the fix absence of this synchronization was causing incorrect results under high concurrent load.

Other changes make the code more correct and reliable from the concurrency viewpoint, e.g "double-checked locking" in method org.locationtech.jts.algorithm.locate.IndexedPointInAreaLocator#createIndex can only work correctly if ``` index ``` field is volatile.

modules/core/src/main/java/org/locationtech/jts/index/intervalrtree/SortedPackedIntervalRTree.java
dr-jts: Just comment it out then.  It may be used for debugging in the future.

ivanveselovsky-tomtom: fixed.

PR #764, opened 2021-08-04
Add DiscreteFrechetDistance

FObermaier:
Like `Hausdorff` distance the `Fréchet` distance is a measure of similarity between curves.
Its metric is better than `Hausdorff`'s because it takes the flow of the curves into account.
It is possible that two curves have a small `Hausdorff` but a large `Fréchet` distance.

This implementation attempts to compute only relevant coordinate distances for perfromance 
and uses a HashMap as sparse matrix to reduce memory consumption. 

Computation of coordinate distances can be delegated to objects implementing the 
`DistanceComputer` interface. A default implementation is provided for `CartesianDistance`.

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/algorithm/distance/DiscreteFrechetDistance.java
+++ b/modules/core/src/main/java/org/locationtech/jts/algorithm/distance/DiscreteFrechetDistance.java
@@ -1,0 +1,295 @@
+/*
+ * Copyright (c) 2021 Felix Obermaier.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * and Eclipse Distribution License v. 1.0 which accompanies this distribution.
+ * The Eclipse Public License is available at http://www.eclipse.org/legal/epl-v20.html
+ * and the Eclipse Distribution License is available at
+ *
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ */
+package org.locationtech.jts.algorithm.distance;
+
+import org.locationtech.jts.geom.Coordinate;
+import org.locationtech.jts.geom.Geometry;
+import java.util.HashMap;
+
+/**
+ * The Fréchet distance between two curves in a metric space is a measure of the
+ * similarity between the curves.
+ * <p/>
+ * An analogy for the Fréchet distance taken from
+ * <a href="http://www.kr.tuwien.ac.at/staff/eiter/et-archive/cdtr9464.pdf">Computing Discrete Fréchet Distance</a>
+ * <pre>
+ * A man is walking a dog on a leash: the man can move
+ * on one curve, the dog on the other; both may vary their
+ * speed, but backtracking is not allowed.
+ * </pre>
+ * @see <a href="http://www.kr.tuwien.ac.at/staff/eiter/et-archive/cdtr9464.pdf">Computing Discrete Fréchet Distance</a>
+ */
+public class DiscreteFrechetDistance {
+
+  /**
+   * Computes the Discrete Fréchet Distance between two {@link Geometry}s
+   * using a {@code cartesian} distance computation function.
+   *
+   * @param g0 the 1st geometry
+   * @param g1 the 2nd geometry
+   * @return the cartesian distance between {#g0} and {#g1}
+   */
+  public static double distance(Geometry g0, Geometry g1) {
+    return distance(g0, g1, CartesianDistance.getInstance());
+  }
+
+  /**
+   * Computes the discrete Fréchet distance between two {@link Geometry}s
+   * using the provided distance computation function.
+   *
+   * @param g0 the 1st geometry
+   * @param g1 the 2nd geometry
+   * @return the Discrete Fréchet Distance between the two geometries
+   */
+  public static double distance(Geometry g0, Geometry g1, DistanceFunction distanceFunction)
+  {
+    if (distanceFunction == null)
+      throw new NullPointerException("distanceFunction");
+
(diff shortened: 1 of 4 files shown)


modules/core/src/main/java/org/locationtech/jts/algorithm/distance/DistanceFunction.java
```diff
@@ -0,0 +1,29 @@
...
+package org.locationtech.jts.algorithm.distance;
+
+import org.locationtech.jts.geom.Coordinate;
+
+/**
+ * Base interface for classes that can compute a distance between two {@link Coordinate}s.
+ *
+ * @author Felix Obermaier
+ */
+public interface DistanceFunction {
```
dr-jts, line 21: The term "function" is fairly overloaded in the JTS codebase.  Can this be changed to `DistanceMetric`?

dr-jts:
Frechet Distance is a nice addition to JTS.

How do you see the `DistanceFunction` capability being used?   I see the example of a Scaled distance metric in the unit test - is this a real-world scenario?  Does it change the Frechet Distance result by more than the scale factor?

dr-jts:
It is very useful to have distance algorithms also be able to return the two points at which the calculated distance is obtained.  This is for confirmation and visualization purposes.  (For example see the implementation in [`DiscreteHausdorffDistance`](https://github.com/locationtech/jts/blob/176eeba6ccef4cc09ab366973cd17be4e8879e9d/modules/core/src/main/java/org/locationtech/jts/algorithm/distance/DiscreteHausdorffDistance.java#L116)) 

Is it possible to add this capability to the Frechet Distance code?

FObermaier:
> How do you see the DistanceFunction capability being used?

I thought more about an implementation using `Haversine` or `Vincenty`'s formula for geographic coordinates.

dr-jts:
> I thought about an implementation using `Haversine` or `Vincenty`'s formula for geographic coordinates.

In geodetic space the locus of boundary segments between vertices is a great circle arc, rather than a straight line. So usually it requires more than just using a geodetic distance metric to produce accurate results.  Of course, geometries can be densified to mitigate this somewhat.

i have had some thoughts about bundling up the various pieces of metadata about the geometry coordinate system (including geodetic information and precision model) into a single object (perhaps called `GeometryRealm` or maybe `CoordinateSystem`.  Still thinking about whether this should be provided on a per-geometry basis (perhaps as part of the `GeometryFactory`), or on a per-operation basis.  Or both...  The advantage to doing this is that it will avoid a proliferation of objects and function signatures in the API.  

Given this, I'd prefer to defer introducing the `DistanceMetric` API unless there is a clear current reason for doing so.

dr-jts:
@FObermaier is there a bug in this algorithm?  What should the Frechet Distance be between the following:

```
A: LINESTRING (1 1, 2 2)
B: LINESTRING (1 4, 2 3)
```
The implementation gives the distance as 2.23606797749979.  But since the distance between the two start points is 3, shouldn't that be the overall Frechet Distance?

FObermaier:
Yes, I'm working on it.

FObermaier:
@dr-jts, I fixed the algorithm, sorry for doing PR too early.

modules/core/src/main/java/org/locationtech/jts/algorithm/distance/DiscreteFrechetDistance.java
```diff
@@ -0,0 +1,596 @@
...
+   * @param rows the number of rows
+   * @param cols the number of columns
+   * @return a matrix storage
+   */
+  private MatrixStorage createMatrixStorage(int rows, int cols) {
+
+    int max = Math.max(rows, cols);
+    // NOTE: these constraints need to be verified
+    if (max < 64)
+      return new HashMapMatrix(rows, cols, Double.POSITIVE_INFINITY);
```
dr-jts, line 113: Is this faster?  If not, why not just use a `RectMatrix`, since storage isn't a problem for small dimensions?

modules/core/src/test/java/org/locationtech/jts/algorithm/distance/MatrixStorageImplTests.java
```diff
@@ -0,0 +1,288 @@
...
+    assertEquals(String.format("%1$d -> %2$d = %4$f /= %3$f", 2, 3, 60d, mat.get(2, 3)), 60d, mat.get(2, 3));
+    assertEquals(String.format("%1$d -> %2$d = %4$f /= %3$f", 2, 4, 70d, mat.get(2, 4)), 70d, mat.get(2, 4));
+    assertEquals(String.format("%1$d -> %2$d = %4$f /= %3$f", 3, 5, 80d, mat.get(3, 5)), 80d, mat.get(3, 5));
+
+  }
+
+  /**
+   * Abstract base class for storing 2d matrix data
+   */
+  private abstract static class MatrixStorage {
```
dr-jts, line 96: It's unusual to duplicate implementation code in unit tests, since this doesn't actually test the code in the library?

FObermaier: I did not see another way to test and verify that `MatrixStorage` implementations work correctly while keeping it private in jts-core.

dr-jts: Yeah, that's a bit of a limitation in the Java package model.  I usually make things like this package-private, so they can be unit tested.

dr-jts:
Computing the distance between the following geometries causes an `ArrayIndexOutOfBoundsException`.

```
A: LINESTRING (80 260, 170 180, 190 290, 310 350, 330 270, 360 280)
B: LINESTRING (120 90, 380 130)
```

PR #775, opened 2021-09-14
Add polygon triangulation algorithms and Tri data structure

dr-jts:
Adds two kinds of polygon triangulation:

* via ear-clipping (fast but low-quality)
* via Constrained Delaunay Triangulation (best quality)

Also adds the `Tri` class to represent triangulations in an efficient way.


modules/core/src/main/java/org/locationtech/jts/triangulate/polygon/PolygonTriangulator.java
```diff
@@ -0,0 +1,99 @@
...
+    geomFact = new GeometryFactory();
+    this.inputGeom = inputGeom;
+  }
+
+  private Geometry compute() {
+    @SuppressWarnings("unchecked")
+    List<Polygon> polys = PolygonExtracter.getPolygons(inputGeom);
+    List<Tri> triList = new ArrayList<Tri>();
+    for (Polygon poly : polys) {
+      List<Tri> polyTriList = triangulatePolygon(poly);
```
pramsey, line 73: Putting in an empty check here
```
if (poly.isEmpty()) continue;
```
made sense to me as I thought about degenerate inputs.

modules/core/src/main/java/org/locationtech/jts/triangulate/polygon/PolygonTriangulator.java
```diff
@@ -0,0 +1,99 @@
...
+   * 
+   * @param inputGeom the input geometry
+   */
+  public PolygonTriangulator(Geometry inputGeom) {
+    geomFact = new GeometryFactory();
+    this.inputGeom = inputGeom;
+  }
+
+  private Geometry compute() {
+    @SuppressWarnings("unchecked")
```
pramsey, line 69: An empty check on `inputGeom` made sense for me here too... probably extraneous in light of the check below. Just returns an empty geometry collection on empty input.

dr-jts: I think it's ok to just let this be handled by the subsequent empty short-circuit.  Less duplicated logic that way.

Issue #763, opened 2021-08-03
GeoJsonWriter produces invalid GeoJSON when serializing LinearRing (or other subclasses of LineString)


pascalriquier:
`
	LinearRing linearRing = new GeometryFactory().createLinearRing(new Coordinate[] { new Coordinate(0, 0), 
			new Coordinate(1, 0), new Coordinate(1, 1), new Coordinate(0, 0)});
	String json = new GeoJsonWriter().write(linearRing);
	new GeoJsonReader().read(json);
`

GeoJsonWriter uses the getGeometryType method of Geometry to specify the type used in the GeoJSON. But, for subclasses of LineString, this returns their actual class (e.g. LinearRing) and not 'LineString'.
This produces invalid GeoJSON, and so, in this case, the GeoJsonReader can not read the json produced by the GeoJsonWriter for these geometries.

mrpan:
i reproduce the error ,GeoJsonReader can not support type "LinearRing"

mrpan:
> i reproduce the error ,GeoJsonReader can not support type "LinearRing"

#796

dr-jts:
The fix seems to be to emit LinearRings as GeoJSON LineStrings.  @pascalriquier  Do you agree with this?

pascalriquier:
Seems like the logica thing to do indeed. Maybe make sure the fix does not only cover LinearRing but all subclasses of LineString?

PR #796, opened 2021-11-02
fix issue ##763 Add  GeoJsonReader support type LinearRing

mrpan:
Add  GeoJsonReader support type LinearRing


dr-jts:
GeoJSON does not have a LinearRing geometry type (cf the spec [section 3.1.6](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.6)).  So I don't think this change is appropriate.

mrpan:
yes,Is such that，if so,LinearRing type in GeoJSON should be use type "Polygon" to parse?，but i think if Geometry has type "LinearRing" ,GeoJSONWriter or GeoJSONReader should can solve it ,

dr-jts:
I don't think that JTS should output a GeJSON type of `LinearRing`, because it's not standard.  And so there's no point in parsing that either (since JTS is the only lib that would emit it).

What is your use case for thinking you need a GeoJSON `LinearRing`?

mrpan:
I see the issue #763 ,That case may be seen frequently。LinearRing is indeed not standard。

PR #803, opened 2021-11-22
difference throws org.locationtech.jts.geom.TopologyException

larsop:
I have added two tests that throws an error

1)  OverlayMiscTest#testFailDiff with standard difference 
2) OverlayMiscTest#testFailDiffOverlayNG where testet with OverlayNG with no snapping


larsop:
I have not testing with scale factor because I was hoping to avoid to use that since we work in degrees and that behaves differently depending where we are on the map .

dr-jts:
The fact that classic Overlay throws an error isn't too surprising, since the B geometry is very, very narrow:
image. image

The `OverlayNG` class itself doesn't actually improve much on classic overlay in terms of robustness, since it uses the same noding strategy by default.

Instead, you should always use `OverlayNGRobust`, which implements more robust noding strategies used if the regular fast one fails.  Using that class it looks to me like this case works?

larsop:
> 

Thanks tested it now and it works and even works with invalid polygons.

dr-jts:
> Thanks tested it now and it works and even works with invalid polygons.

Good to hear.  And yes, OverlayNG works even with "mildly" invalid polygons.  There's some description about this here:  http://lin-ear-th-inking.blogspot.com/2020/06/jts-overlayng-tolerant-topology.html

larsop:
Thanks a couple of more questions .
1) Will g1.difference(g2) be moved to  OverlayNGRobust.overlay(g1, g2, OverlayNG.DIFFERENCE) in Java or do we have to change in our  java code base ?
2) This seems work in latest Postgis is it there using OverlayNGRobust by default ?

dr-jts:
You can set the system property `Djts.overlay=ng` (described [here](https://github.com/locationtech/jts/blob/master/USING.md#jts-system-properties)) to force the standard `Geometry` overlay methods to use OverlayNG.  This uses `OverlayNGRobust`, so there should be no errors encountered.

PostGIS (and GEOS) use OverlayNG now (including the robustness heuristics).

Issue #798, opened 2021-11-03
OverlayNG robustness failure


dr-jts:
The following case produces incorrect results for all overlay operations using OverayNG (in particular, intersection):
```
A: POLYGON ((66697.40120137333 185279.95469107336, 66698.375 185273.625, 66697.375 185280.125, 66697.40120137333 185279.95469107336))
B: POLYGON ((66710 185280, 66710 185260, 66690 185260, 66690 185280, 66710 185280))
```
![image](https://user-images.githubusercontent.com/3529053/140175813-13c6af7b-80dd-47b4-bfc9-a8c82165a8d6.png)

Incorrect result of `intersection`:
![image](https://user-images.githubusercontent.com/3529053/140178360-ed8e71d6-179d-4a24-bf5d-68370667f683.png)

### Analysis
A is a very narrow triangle (shown below in Reveal Topology mode):
![image](https://user-images.githubusercontent.com/3529053/140176146-7c3d405c-a859-4704-a2b3-2ba6959d8629.png)

The cause of the problem is that after noding the long side of the triangle has a vertex introduced, which causes the lower section to shift left enough so that middle vertex ends up on the opposite side of the noded line segment:
![image](https://user-images.githubusercontent.com/3529053/140176246-fc0889b5-7971-47f1-b846-986e8e747585.png)

### Case 2

```
A: POLYGON ((61607.62679190189 190194.1555478014, 61620.5389918983 190197.4990478009, 61619.64380966499 190197.26724827816, 61607.62679190189 190194.1555478014))
B: POLYGON ((61620.04309999943 190199.41420000046, 61620.41290000081 190197.98570000008, 61620.53899999708 190197.4989, 61607.62680000067 190194.15549999848, 61607.07620000094 190196.27259999886, 61620.04309999943 190199.41420000046))
```
![image](https://user-images.githubusercontent.com/3529053/140262951-d5bd779e-08f8-4092-9061-c2c5089a69e6.png)

### Case 3

```
A: POLYGON ((181093.43788657014 172099.78376270586, 181093.375 172099.375, 181093.57688359552 172100.6872433709, 181093.43788657014 172099.78376270586))
B: POLYGON ((181118.78476615425 172110.76716212876, 181097.49019999802 172097.5095999986, 181081.35400000215 172105.30889999866, 181118.78476615425 172110.76716212876))
```

[...]

PR #808, opened 2021-11-27
A intersection between a very narrow polygon and bigger one seems to give wrong result with OverlayNGRobust

larsop:
We startet to test more on OverlayNGRobust, but then we got this case .

Input is two geometries where one is very narrow at the border of the big polygon

Screenshot 2021-11-27 at 09 57 32

The other is big a polygon

Screenshot 2021-11-27 at 09 59 44

The intersection between this should have been close to the narrow polygon like this

Screenshot 2021-11-27 at 10 01 55

But with latest master i get this which is way to big.

Screenshot 2021-11-27 at 10 03 15

Before testing  on OverlayNGRobust used our own code together with snapcode that I got you some years a go and home made code handled this case correctly. 

The difference between OverlayNGRobust master and my old code is that I try resolve MultiPolygon before I start to do the intersection, then I don't get any TopoError and do not need to do any snaping.

So I added a test OverlayMiscTest#test_7_Intersection_loop where used code like this and that seems to work

```
     for (int i = 0; i < g2.getNumGeometries(); i++) {
          Geometry g2GeoN = g2.getGeometryN(i);
          for (int y = 0; y < g1.getNumGeometries(); y++) {
              Geometry g1GeoN = g1.getGeometryN(y);
              Geometry cleanIntersection = g1GeoN.intersection(g2GeoN);
              cleanIntersection.getGeometryType();
              newGeoList.add(cleanIntersection);
          }
      }
```

But the OverlayMiscTest#test_7_Intersection_OverlayNGRobust failed 

With this [commit](https://github.com/larsop/jts/commit/585d17a1232facdcbc8aee120a1c69343cc0f5bbl) the OverlayMiscTest#test_7_Intersection_OverlayNGRobust runs through.


dr-jts:
I suspect this is a similar issue to #798.  I have a fix to that issue, which consists of an additional heuristic check on the result area compared to the inputs.  That will be implemented very soon.

However, in your second case this will return a result of `POLYGON EMPTY`, due to the triggering of more aggressive snapping. As you point out, this can be improved by intersecting each polygon element separately.  But at the moment I don't want to build that in, due to complexity and performance reasons.  

How important is having a non-empty result for you?  The narrow input polygon is very small, so it seems that perhaps it's just an artifact, and doesn't matter if it's essentially discarded?

dr-jts:
@larsop is this fixed by #812?

dr-jts:
Should be fixed.  Closing.

larsop:
Thanks , yes I tested with master now and it works perfect, sorry for the late replay.

PR #835, opened 2022-02-08
Interpolate Z values in Densifier class

nicolas-f:
Manage the interpolation of Z values of Coordinates in the Densifier class

About https://github.com/orbisgis/h2gis/issues/1272

Without this patch the Densifier would output a mix of Z and NaN values

(retry commit and PR eclipse is blocking for email issue I think ?)


nicolas-f:
Maven does not like checkstyle.xml . Does not know how to define checkstyle.header.file ?

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.1.2:check (validate) on project jts: Failed during checkstyle execution: Failed during checkstyle configuration: unable to parse configuration stream: com.puppycrawl.tools.checkstyle.api.CheckstyleException: Property ${checkstyle.header.file} has not been set -> [Help 1]
```

PR #855, opened 2022-03-26
Exclude junit compile dependency from json-simple

murdos:
JUnit should be a test dependency only. Projects using `jts-io-common` also inherits from this (outdated) dependency.
See https://github.com/fangyidong/json-simple/issues/91


mprins (approved):

jodygarnett (approved):
We do not want a transitive dependency on junit, json-simple library probably just messed up.

PR #859, opened 2022-04-07
Fix LengthLocationMap.getLength for component end points

FObermaier:
Issue reported and fixed by @inforithmics.
https://github.com/NetTopologySuite/NetTopologySuite/pull/603

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/linearref/LengthLocationMap.java
+++ b/modules/core/src/main/java/org/locationtech/jts/linearref/LengthLocationMap.java
@@ -20,6 +20,10 @@ import org.locationtech.jts.geom.Geometry;
  * along a linear {@link Geometry}.
  * Negative lengths are measured in reverse from end of the linear geometry.
  * Out-of-range values are clamped.
+ * <p>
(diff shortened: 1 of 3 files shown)


modules/core/src/main/java/org/locationtech/jts/linearref/LengthLocationMap.java
```diff
@@ -197,7 +197,13 @@ public double getLength(LinearLocation loc)
...
         totalLength += segLen;
       }
-      it.next();
+      else
+      {
+        // At the end of the component
+        if (loc.getComponentIndex() == it.getComponentIndex()) {
+          return totalLength;
+        }
+      }      it.next();
```
dr-jts, line 206: `it.next()` should be on new line

dr-jts:
Here's a test that can be added to `LengthIndexedLineTest` to expose this bug:

```
  public void testExtractLineIndexAtEndpointOfTouchingLines()
  {
    runIndicesOfThenExtract("MULTILINESTRING((0 0, 0 50), (0 50, 0 100))",
            "LINESTRING (0 50, 0 60)");
  }
```

PR #882, opened 2022-06-03
Fix STRtree.nearestNeighbour to handle empty trees

### LINKED ISSUE #881: STRtree.nearestNeighbour with K throws NPE for empty tree
```java
java.lang.NullPointerException
	at org.locationtech.jts.index.strtree.BoundablePair.distance(BoundablePair.java:94)
	at org.locationtech.jts.index.strtree.BoundablePair.(BoundablePair.java:48)
	at org.locationtech.jts.index.strtree.STRtree.nearestNeighbour(STRtree.java:523)
```
STRtree's internal structure and logic doesn't protect itself when the tree is empty.
(Version 1.18.2)

ashr123:
fixes #881

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/index/strtree/STRtree.java
+++ b/modules/core/src/main/java/org/locationtech/jts/index/strtree/STRtree.java
@@ -295,8 +295,9 @@ implements SpatialIndex, Serializable
    */
   public Object[] nearestNeighbour(ItemDistance itemDist)
   {
-    if (isEmpty()) return null;
-    
+    if (isEmpty())
+      return new Object[0];
+
     // if tree has only one item this will return null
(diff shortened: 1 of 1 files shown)


dr-jts:
This needs unit tests - are you able to provide some?

Also, we should split the empty tree fix and the change in return value into two separate PRs, for tracking purposes.

ashr123:
@dr-jts can you provide the unit test? I've created the fix in the Github's editor and I didn't cloned it to my computer and I don't fully familiar with the code (as probably you are)
and I agree with your proposal ablut splitting the PR

dr-jts:
I'd rather make this change in my dev environment, with unit tests at the same time.  I will make a new PR for that, and then close this one.

dr-jts:
Replaced by #886

PR #892, opened 2022-06-10
Fixes regression in WKTReader.

jnh5y:
https://github.com/locationtech/jts/commit/96e76995 introduced a small change in how geometries like "POINT M (2 0 20))" were read.

This adds a test and (hopefully) fixes the change.


dr-jts (approved):
Looks good!

modules/core/src/main/java/org/locationtech/jts/io/WKTReader.java
```diff
@@ -337,10 +337,10 @@ private Coordinate createCoordinate(EnumSet<Ordinate> ordinateFlags) {
     boolean hasM = ordinateFlags.contains(Ordinate.M);
     if (hasZ && hasM) 
       return new CoordinateXYZM();
+    if (hasM)
```
jodygarnett, line 340: nice!

PR #896, opened 2022-06-22
Add missing TWKB feature on 1.19.0 changelog

murdos:
And also fix of transitive compile dependency on `junit`


murdos:
@dr-jts @jodygarnett if you could also update [the description of the github release](https://github.com/locationtech/jts/releases/tag/1.19.0) that would be nice! :pray:

dr-jts:
Thanks!

PR #923, opened 2022-11-01
Replace Math.sqrt(x^2+y^2) with Math.hypot(x,y)

RichMacDonald:
Replaced all occurrences of Math.sqrt(x^2, y^2) with Math.hypot(x, y)


micycle1:
`hypot()` is a much more costly function than `sqrt()`. I don't think it should be included in hot areas (or areas that are likely to be hot from users' calling code).

Issue #939, opened 2022-12-08
WKTReader silently adds dimensions when setIsOldJtsCoordinateSyntaxAllowed is true


chenhh021:
As the title says, WKTReader silently adds dimension. 

For example, it will create coordinate sequence `10 10 NaN` after load `POINT (10, 10)`. The following test can reproduce this behavior.

``` java
public void testDimensionConsistence() throws Exception {
    //test 2-d point with XY coordinate
    CoordinateSequence seq1 = createSequence(Ordinate.createXY(), new double[] {10, 10});
    Point pt1 = (Point)readerXYOld.read("POINT (10 10)");
    assertTrue(isEqual(seq1, pt1.getCoordinateSequence()));
}
```

Is it designed to behave this way?

I could fix the dimension change alone. But a thorough fix may need some more time, as there may need some conversion for mixed use of 2-d and 3-d coordinates.  At present such conversion is useless as WKTReader always extend 2-d coordinates to 3-d.

PR #940, opened 2022-12-08
Fix silent dimension addition of WKTReader when allowing old syntax

chenhh021:
A quick fix for #939. May need further work for thorough fix.


chenhh021:
Wrongly include old commits. Will recreate later.

PR #949, opened 2023-01-12
Dedup `Polygonizer` invalid ring lines result

### LINKED ISSUE #947: Polygonizer::getInvalidRingLines returns duplicate lines
As reported in https://github.com/libgeos/geos/issues/782

After polygonizing the following geometry, `Polygonizer::getInvalidRingLines` returns duplicates of the invalid ring.

`MULTILINESTRING ((0 0, 1 0, 1 1, 0 1, 0 0),  (0 0, 0.5 0.5),  (1 1, 2 2, 1 2, 2 1, 1 1))`

dr-jts:
An improvement to ensure that the result of `Polgyonizer.getInvalidRingLines` does not contain duplicate linework for rings.

Fixes #947 

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/operation/polygonize/EdgeRing.java
+++ b/modules/core/src/main/java/org/locationtech/jts/operation/polygonize/EdgeRing.java
@@ -104,9 +104,11 @@ class EdgeRing {
   private List<LinearRing> holes;
   private EdgeRing shell;
   private boolean isHole;
+  private boolean isValid = false;
   private boolean isProcessed = false;
   private boolean isIncludedSet = false;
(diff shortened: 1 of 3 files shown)


dr-jts:
### Note on Alternative Implementation

This fix adheres closely to the current behaviour of the `getInvalidRingLines` method, which returns closed LineStrings representing the invalid rings found during polygonization.  An alternative which has some appeal is to instead return the LineStrings of the input edges which form invalid rings (which may or may not form a complete ring).  The advantage is that there is then a direct relationship with the input data.  The disadvantage is that it may make it harder to determine the exact locations of the invalidities (which are typically ring self-intersections).

PR #961, opened 2023-03-01
Fix CoverageSimplifier to handle holes correctly

### LINKED ISSUE #953: CoverageSimplifier creates invalid coverage for dataset with holes
This example creates an invalid coverage output when simplified with tolerance  100:
```
GEOMETRYCOLLECTION (POLYGON ((200 300, 300 300, 300 100, 100 100, 100 300, 200 300), (170 220, 170 160, 200 140, 200 250, 170 220), (170 250, 200 250, 200 300, 170 250)), POLYGON ((170 220, 200 250, 200 140, 170 160, 170 220)), POLYGON ((170 250, 200 300, 200 250, 170 250)))
```
image

The cause is likely something to do with not respecting ring end-nodes in the same way on both inside and outside ring.

mukoki:
Fixes #953


dr-jts:
Nice fix - thanks.

It would be best to include the two test cases as unit tests.

mukoki:
Unfortunately, writing the tests revealed the need to go deeper in the change. Currently, the simplifier removes corners with a distinction between 
- normal edge case (end points are preserved) and 
- ring case (any corner may be removed as far as three distinct points are left).
In the case of a hole touching another hole or touching the shell, we would like to preserve the starting node (which is the node touching the other ring), but edges are passed to the simplifier as a MultiLineString and information that the end node must be preserved is lost. We could preserve initial node everywhen, but for a free ring, it is not required.
Do you agree to pass a bit more information than a MultiLineString to the simplifier ?

dr-jts:
> Unfortunately, writing the tests revealed the need to go deeper in the change.

Yay for unit tests!

>   Do you agree to pass a bit more information than a MultiLineString to the simplifier ?

Yes, that seems to be necessary.  What are your ideas for doing this?  Perhaps an array of booleans indicating whether a ring can be fully simplified, or should have its endpoint preserved?

dr-jts:
See also #962, which I think requires a different fix.

mukoki:
I pushed a new commit adding an attribute to CoverageEdge indicating how many nodes need to be preserved (0, 1 or 2).
It is a much deeper change. Let me know what you think. The main drawback is that I removed public static method from TPVWSimplifier as MultiLineString is not sufficiant to preserve topology :-(
Also added two tests from #953 report

wrt TPVWSimplifier, it could already remove endpoint corner of rings and that was the point ! The LinearRing to CoverageEdge transformation was setting the endpoint on the node touching another ring if any, and but it could be removed as any vertex of a ring. Now it should be preserved (processed as if it was a linestring, not a ring).

See also my comments on #962

modules/core/src/main/java/org/locationtech/jts/coverage/CoverageEdge.java
```diff
@@ -26,12 +26,16 @@ class CoverageEdge {
...
+    CoverageEdge edge = new CoverageEdge(pts);
+    edge.nodeCount = 0;
+    return edge;
   }
 
   public static CoverageEdge createEdge(LinearRing ring, int start, int end) {
     Coordinate[] pts = extractEdgePoints(ring, start, end);
-    return new CoverageEdge(pts);
+    CoverageEdge edge = new CoverageEdge(pts);
+    edge.nodeCount = start == end ? 1 : 2;
```
dr-jts, line 37: This is only used to check if it is zero, so can it just be a boolean (say `isNodePreserved`) ?

modules/core/src/main/java/org/locationtech/jts/coverage/TPVWSimplifier.java
```diff
@@ -107,13 +115,14 @@ private Geometry simplify() {
...
     List<Edge> edges = new ArrayList<Edge>();
-    if (lines == null)
+    if (covEdges == null)
       return edges;
-    for (int i = 0 ; i < lines.getNumGeometries(); i++) {
-      LineString line = (LineString) lines.getGeometryN(i);
-      edges.add(new Edge(line, areaTolerance));
+    for (CoverageEdge covEdge : covEdges) {
+      Edge edge = new Edge(covEdge.getCoordinates(), areaTolerance);
+      edge.nodeCount = covEdge.getNodeCount();
```
dr-jts, line 124: Should be passed to constructor?

dr-jts:
Another alternative is to provide `TPVWSimplifier` with a `Set` of coordinates which are ring nodes that must be preserved.

I prefer this approach, since it reduces the amount of shared classes.  This makes `TPVWSimplifier` more generic.  At some point it may be promoted to being a public class (and moved to the 'simplify` package).

modules/core/src/main/java/org/locationtech/jts/coverage/TPVWSimplifier.java
mukoki: Yes, you're right.

mukoki:
I see. In this case, wouln't it be preferable to pass constrained rings (containing one node) as a third MultiLineString beside normal edges (0/2 nodes) and constraints to avoid computation of edges from nodes in TPVWSimplifier.

dr-jts:
> Wouln't it be preferable to pass constrained rings (containing one node) as a third MultiLineString beside normal edges (0/2 nodes) and constraints to avoid computation of edges from nodes in TPVWSimplifier.

That is another option.  I think I'd prefer passing the edges which have their endpoints to be preserved (line or ring), and the ones (rings only) which can have all vertices simplified.   Does this make sense?

It's the same amount of work in either case, since the "rings with nodes" have to be determined by checking if their endpoint is in the set of identified nodes.  But maybe it is best to keep this processing outside of `TPVWSimplifier`.

dr-jts:
Superseded by #963

Issue #962, opened 2023-03-02
CoverageSimplifier does not preserve degree-4 boundary nodes


dr-jts:
Degree-4 Boundary Nodes are vertices which occur in 2 rings which touch only at the vertex.  This situation only occurs when the vertex and its adjacent segments are all on the boundary of the coverage.

Degree-4 Boundary Nodes occur where:
* two polygons touch at a single point
* an empty hole touches its shell 

(Note that in the case of a hole touching its shell and containing another polygon, the vertex occurs in 3 rings and so **is** identified as a node.)

These nodes should (probably) be preserved during coverage simplification.  Currently the `CoverageSimplifier` does not recognize that the common vertex is a node, and thus may remove it during simplification.

### Example - Touching shells
Simplification tolerance = 4
```
POLYGON ((0 0, 0 5, 5 6, 10 5, 10 0, 0 0))
POLYGON ((0 10, 5 6, 10 10, 0 10))
```
![image](https://user-images.githubusercontent.com/3529053/222538509-e013a7a3-7812-41b1-b70c-72844dade6f3.png)

### Example - Shell touching empty hole
Simplification tolerance = 4
```
POLYGON ((0 0, 0 20, 10 19, 20 20, 20 0, 0 0), (5 18, 10 19, 15 18, 10 10, 5 18))
```
![image](https://user-images.githubusercontent.com/3529053/222600107-4f92015b-c81d-443a-80a1-b7a320e69439.png)

mukoki:
Hey, this case is not exactly the same as the #961
The intersection point between two shells is not identified as a "node" as it is shared by two geometries only !
Not easy to identify this particular case. We can avoid the separation of the two polygons using simplifyInner (nice option indeed!), but this is not exactly the same thing.
And unfortunately there is also a case with a hole touching the shell. In the test case of #961, there is such a thing and the test is OK, but I now think this is a strike of luck. 
To solve it, we may have to change the way unmodifiable nodes are identified. Not sure how to do though.

dr-jts:
> Hey, this case is not exactly the same as the #961 

Yes, it's a different issue.

> And unfortunately there is also a case with a hole touching the shell. 

RIght.  I"ve added that as an example.

> To solve it, we may have to change the way unmodifiable nodes are identified.

Agreed, that is the problem.  The current way of identifying nodes is too simplistic to handle this situation.  But it is fast!  The challenge is to allow these kinds of degree-2 nodes to be identified, with decent performance.

[...]

PR #963, opened 2023-03-05
Fix coverage simplifier issues 953 962

mukoki:
Replace the previous pull request #961 : fix both issues #953 and #962 according to the discussion


mukoki:
I tried to follow your suggestions about identification of nodes on shells sharing only one point.
Finally, this new pull request fix several bugs : 
- fix computation of CoverageEdge key in the case of ring including a single node (= first commit of my first PR)
- identifiying single node when two shells touch each other (propagate the information to CoverageEdge and Edge as a boolean) and pass it through MultiLineString via a BitSet included in the Geometry
- also fix PriorityQueue of corners which did not include all possible corners (I had to change the test testFilledHole() as the expected polygon was not simplified to the maximum before this change). Also clean the Queue from deprecated corners)
- add some tests and fix GeometryTestCase which was displaying the wrong message
- there is also an unfinished change in LinkedList (not sure it is necessary)

mukoki:
On a significative dataset (117 Mb), the processing time goes from 41.5 s to 46 s after applying the PR.

modules/core/src/main/java/org/locationtech/jts/simplify/LinkedLine.java
```diff
@@ -51,6 +51,7 @@ private int[] createNextLinks(int size) {
       next[i] = i + 1;
     }
     next[size - 1] = isRing ? 0 : size;
+    next[size - 1] = isRing ? 0 : NO_COORD_INDEX;
```
dr-jts, line 54: I think this is good

modules/core/src/main/java/org/locationtech/jts/coverage/TPVWSimplifier.java
```diff
@@ -111,9 +108,11 @@ private List<Edge> createEdges(MultiLineString lines) {
     List<Edge> edges = new ArrayList<Edge>();
     if (lines == null)
       return edges;
+    BitSet freeRings = (lines.getUserData() != null && lines.getUserData() instanceof BitSet) ?
```
dr-jts, line 111: Not crazy about passing this via userData.  Can't it be a parameter?

And I actually prefer your original idea of passing two sets of lines, one with endpoint constraints and one of the unconstrained rings.

modules/core/src/main/java/org/locationtech/jts/coverage/CoverageRingEdges.java
```diff
@@ -216,6 +217,22 @@ private Set<Coordinate> findNodes(Geometry[] coverage) {
     return nodes;
   }
 
+  private Set<Coordinate> findBoundaryNodes(Set<LineSegment> lineSegments) {
+    Map<Coordinate, Integer> boundaryVertexCounter = new HashMap<>();
+    Set<Coordinate> nodes = new HashSet<Coordinate>();
+    for (LineSegment line : lineSegments) {
+      int count = boundaryVertexCounter.getOrDefault(line.p0, 0);
+      boundaryVertexCounter.put(line.p0, count+1);
+      count = boundaryVertexCounter.getOrDefault(line.p1, 0);
```
dr-jts, line 226: Why set this twice?  Seems awkward.

dr-jts, line 226: Definining a new local var would make this logic clearer

mukoki: Do you mean a new local variable for the second count ? I can do it for sure. Not sure it would make the logic clearer though as the variable count all points with the same coordinate, either p0 or p1.
Indeed, I try to follow the JTS style which is mostly based on java5 I think ,but from my point of view, a more readable version would be : 
  private Set<Coordinate> findBoundaryNodes(Set<LineSegment> lineSegments) {
    Map<Coordinate, Integer> counter = new HashMap<>();
    for (LineSegment line : lineSegments) {
      counter.put(line.p0, counter.getOrDefault(line.p0, 0) + 1);
      counter.put(line.p1, counter.getOrDefault(line.p1, 0) + 1);
    }
    return counter.entrySet().stream()
        .filter(e->e.getValue()>2)
        .map(Map.Entry::getKey).collect(Collectors.toSet());
  }

dr-jts: That looks even better!

dr-jts:
You'll need to rebase, since I pushed a fix to `GeometryTestCase`.

modules/core/src/main/java/org/locationtech/jts/coverage/TPVWSimplifier.java
mukoki: Can probably do that, but this separation into two sets of lines is probably of no use for the end user. That's why I thought it would be better to consider it as an implementation detail. I understand the argument to keep things accessible and to avoid hidden tricks though.

dr-jts: Ok, either way is fine for now.  If a different calling signature becomes useful it can always be added later.

mukoki:
Sorry, with the different commits, sync and rebase, I'm not sure the PR is still usable for you. Let me know. I've also made a separate PR where I replaced the BitSet trick by a second MultiLineString to pass the unconstrained rings. Use the one you prefer or another way if you have a precise idea in mind.

modules/core/src/main/java/org/locationtech/jts/coverage/TPVWSimplifier.java
dr-jts: Actually I should have said - the BitSet data is fine, but would be better to pass it as an actual parameter to `TPVWSimplifier`.  I prefer not to hide key data in the untyped `userData` field.

dr-jts:
This PR looks good, with one small change request as noted in a comment (pass the BItSet as an explicit parameter to `TPVWSImplifier`).

PR #970, opened 2023-03-15
Improve use of isFreeRing in createEdges

FObermaier:
The instance field `isFreeRing` refers to the geometry in the instance field `inputLines`. It has nothing to do with the geometry in the instance field `constraintLines` and thus should not be applied to it.


dr-jts:
Good point, thanks.

Even better I think is to not create an empty BitSet if `isFreeRing` is null, but to do this instead:
```
      boolean isFree = isFreeRing == null ? false : isFreeRing.get(i);
      edges.add(new Edge(line, isFree, areaTolerance));
```
If you agree, do you want to update this PR?

Issue #907, opened 2022-09-15
MinimumDiameter.minimumRectangle result is incorrect


dr-jts:
As reported in https://github.com/libgeos/geos/issues/679, the result computed by the `MinimumDiameter.minimumRectangle` method is not always correct.  This is because the algorithm used is too simplistic.

A reproducer is:
```
POLYGON ((150 300, 200 300, 300 300, 300 250, 280 120, 210 100, 100 100, 100 240, 150 300))
```
![image](https://user-images.githubusercontent.com/3529053/190499155-f98574a9-d143-4040-b5ff-b621d694bdf5.png)

The fix is to create a new MinimumRectangle class using the classic [Rotating Calipers](https://www.geometrictools.com/Documentation/MinimumAreaRectangle.pdf) algorithm.

PR #977, opened 2023-04-25
Add MinimumAreaRectangle

dr-jts:
To address https://github.com/libgeos/geos/issues/679 (see #907), this adds a `MinimumAreaRectangle` class implementing the standard "rotating-calipers` algorithm for computing a **Minimum-Area Rectangle**.

`MinimumDiameter.getMinimumRectangle` was previously used for this, but it does not always compute the Minimum-Area Rectangle.  It is kept available, for backwards compatibility.  Also, it computes the **Minimum-Width Rectangle**, which may be useful.

image


willcohen:
@pramsey and @dr-jts 
Just want to say thank you for this -- this has been quite useful as I work on determining a first pass for regularizing city parcel boundaries to determine depth and width when they're rectangular-ish.

PR #1005, opened 2023-10-11
Fix OverlayNG Area Check heuristic for difference

### LINKED ISSUE #951: OverlayNG area check heuristic causing issues when snapping noder is used
After upgrading JTS from 1.18.2 to 1.19.0 ran into a number of tests that use difference operation failing with `TopologyException: Result area inconsistent with overlay operation`.

I tracked it down to `OverlayUtil#isResultAreaConsistent` (which was added in https://github.com/locationtech/jts/pull/812):
```
case OverlayNG.DIFFERENCE:
  isConsistent = isLess(areaResult, areaA, AREA_HEURISTIC_TOLERANCE)   // true; args are: 0, 6.512
              && isGreater(areaResult, areaA - areaB, AREA_HEURISTIC_TOLERANCE); // false; args are 0; 8.8818E-16
  break;
```
In our case we are performing a difference operation on two almost-exactly-the-same geometries and expect an empty result.

Note: we are using snapping noder like in [OverlayNGSnappingFunctions code](https://github.com/locationtech/jts/blob/master/modules/app/src/main/java/org/locationtech/jtstest/function/OverlayNGSnappingFunctions.java).

### Test case 1
This case fails with `TopologyException`:
```
@Test
void test2() throws ParseException {
    Geometry g1 = new WKTReader().read("MULTIPOLYGON (((0.37676311 2.57570853, 7.28652472 0.00028375, 7.60034931 0.81686059, 0.50229292 3.4551325, 0.37676311 2.57570853)))");
    Geometry g2 = new WKTReader().read("MULTIPOLYGON (((0.50229292 3.4551325, 7.60034931 0.81686059, 7.28652472 0.00028375, 0.37676311 2.57570853, 0.50229292 3.4551325)))");

    // Note: in our code we have an utility function that wraps this logic
    Geometry overlayResult = OverlayNG.overlay( 
            g1,
            g2,
            OverlayNG.DIFFERENCE,
            g1.getPrecisionModel(),
            new ValidatingNoder(new SnappingNoder(1e-4))
    );

    assertTrue(overlayResult.isEmpty());
}
```
In this case `OverlayUtil#isResultAreaConsistent` is reached once, and the expression in `isGreater` check works out to be:
- `v1 >= v2 * (1 - tol);`
- `0 >= 8.8819E-16 * (1 - 0.1)` ==> `false`

### Test case 2
This case passes with the same data:
```
@Test       // Passes
void test() throws ParseException {
    Geometry p1 = new WKTReader().read("MULTIPOLYGON (((0.37676311 2.57570853, 7.28652472 0.00028375, 7.60034931 0.81686059, 0.50229292 3.4551325, 0.37676311 2.57570853)))");
    Geometry p2 = new WKTReader().read("MULTIPOLYGON (((0.50229292 3.4551325, 7.60034931 0.81686059, 7.28652472 0.00028375, 0.37676311 2.57570853, 0.50229292 3.4551325)))");

    Geometry result = p1.difference(p2);

    assertEquals(result.getArea(), 0);
}
```
In this case it seems `OverlayNGRobust` is used instead, which results in multiple calls to `OverlayUtil#isResultAreaConsistent` (with slightly different values) and eventually it succeeds in passing the heuristic check.

### Question
To me this feels like a bug in the heuristic check, but maybe our utility function for the difference operation is implemented in a suboptimal way?

If it's indeed a bug, and the fix is relatively simple (for someone who does not know JTS internals well), we'd be happy to contribute a PR with the fix.

dr-jts:
This change fixes the OverlayNG area check heuristic for the difference operation, to handle cases where the difference result has a very small area.

Fixes #951

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/operation/overlayng/OverlayUtil.java
+++ b/modules/core/src/main/java/org/locationtech/jts/operation/overlayng/OverlayUtil.java
@@ -392,6 +392,8 @@ class OverlayUtil {
     if (geom0 == null || geom1 == null) 
       return true;
     
+    if (result.getDimension() < 2) return true;
+    
     double areaResult = result.getArea();
     double areaA = geom0.getArea();
     double areaB = geom1.getArea();
@@ -403,8 +405,7 @@ class OverlayUtil {
                   && isLess(areaResult, areaB, AREA_HEURISTIC_TOLERANCE);
       break;
     case OverlayNG.DIFFERENCE:
-      isConsistent = isLess(areaResult, areaA, AREA_HEURISTIC_TOLERANCE)
-                  && isGreater(areaResult, areaA - areaB, AREA_HEURISTIC_TOLERANCE);
+      isConsistent = isDifferenceAreaConsistent(areaA, areaB, areaResult, AREA_HEURISTIC_TOLERANCE);
       break;
     case OverlayNG.SYMDIFFERENCE:
       isConsistent = isLess(areaResult, areaA + areaB, AREA_HEURISTIC_TOLERANCE);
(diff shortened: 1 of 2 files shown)


grimsa:
@dr-jts I will test the fix by Monday and report here.

grimsa:
@dr-jts I can confirm that the fix works - we are no longer seeing the errors related to area check heuristic that we have seen before.

PR #1012, opened 2023-11-04
Improve performance of HPRtree

msbarry:
While profiling performance of [planetiler](https://github.com/onthegomap/planetiler) I noticed that STRtree in MCIndexNoder was one of the bottlenecks. I looked into HPRtree and saw there were a few opportunities to improve performance using some ideas from the [flatbush](https://github.com/mourner/flatbush) library:

- compute hilbert indexes only once for each item while sorting
- limit the sort so it just sorts items into the correct node, but doesn't bother sorting within the node
- extract item bounds and values into arrays for faster lookups during query

This version performs faster in some of the microbenchmarks (especially when result sets are large) but when I swap out STRtree with this improved HPRtree planetiler spends about 30% less time in MCIndexNoder.computeNodes, isValid checks, and about 10% less time in the most expensive operations planetiler uses: GeometryPrecisionReducer.reduce and bufferUnionUnbuffer.

### DIFF
--- a/doc/JTS_Version_History.md
+++ b/doc/JTS_Version_History.md
@@ -58,9 +58,10 @@ Distributions for older JTS versions can be obtained at the
 * Fix `LargestEmptyCircle` to handle polygonal obstacles (#988)
 * Make intersection computation more robust (#989)
 * Fix `VariableBuffer` to handle zero vertex buffer distances correctly (#997)
-* Fix IncrementalDelaunayTriangulator to ensure triangulation boundary is convex (#1004)
+* Fix `IncrementalDelaunayTriangulator` to ensure triangulation boundary is convex (#1004)
 * Fix OverlayNG Area Check heuristic for difference (#1005)
-* Fix InteriorPointPoint to handle empty elements
+* Fix `InteriorPointPoint` to handle empty elements
+* Fix `DistanceOp` for empty elements (#1010)
 
 ### Performance Improvements
 
--- a/modules/core/src/main/java/org/locationtech/jts/index/ArrayListVisitor.java
+++ b/modules/core/src/main/java/org/locationtech/jts/index/ArrayListVisitor.java
@@ -19,11 +19,11 @@ import java.util.ArrayList;
  * 
  * @version 1.7
  */
-public class ArrayListVisitor
-    implements ItemVisitor
+public class ArrayListVisitor<T>
+    implements ItemVisitor<T>
 {
 
-  private ArrayList items = new ArrayList();
+  private final ArrayList<T> items = new ArrayList<>();
   
   /**
    * Creates a new instance.
@@ -36,7 +36,7 @@ public class ArrayListVisitor
    * 
    * @param item the item to visit
    */
-  public void visitItem(Object item)
+  public void visitItem(T item)
   {
     items.add(item);
   }
@@ -46,6 +46,6 @@ public class ArrayListVisitor
    * 
    * @return the array of items
    */
-  public ArrayList getItems() { return items; }
+  public ArrayList<T> getItems() { return items; }
 
 }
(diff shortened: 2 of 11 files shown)


dr-jts:
The code is certainly nicer with generics.  Do you think the compile warnings are going to be obnoxious for downstream users?  

Generics have been asked for for a while for index containers, so maybe now is the time to add them.

msbarry:
I don't think they would be too bad for consumers if it means they can get rid of instanceof checks and casting and don't cause any compile failures. I could go and update the other spatial index classes too. Think I should do that in this PR or a separate one? It might touch a lot of files.

dr-jts:
> I could go and update the other spatial index classes too. Think I should do that in this PR or a separate one? It might touch a lot of files.

No, that should be done in a separate PR.  Actually I'm still wondering if SpatialIndex generics are too much for this PR.   I hate to lose all the work you've done - but perhaps you can move it to a different PR?

msbarry:
No worries. Sounds good to keep this one focused on the performance improvements. I'll move generics out into another pr

msbarry:
For the generics PR, there are a bunch of things that it _could_ include but it starts to get big pretty fast:

parameterizing the external API of index classes
- `SpatialIndex` and its subclasses (small)
- `ItemVisitor` and its subclasses (small)
- every other index in `jts.index` package (medium)

parameterizing the internal implementation of index classes
- `STRtree`/`AbstractSTRtree`/`SIRtree` (medium)
- `KDTree` (medium)
- each other is probably small-medium

updating places in the code that use these indexes to benefit from the parameterized versions:
- `STRtree` (medium/large)
- `KDTree` (medium/large)
- `Quadtree` (medium)
- each other one is probably small

What do you think the scope of an initial PR should be?

dr-jts:
That's a great list, @msbarry.

> What do you think the scope of an initial PR should be?

Ideally - all of it!  Or, at least `SpatialIndex` implementors and usage.

dr-jts:
Sorry, @msbarry , I've been delayed on getting on to merging this PR.

Why does switching to using two arrays instead of `ArrayList<Item>` improve performance?

msbarry:
@dr-jts thanks for getting back, I think it's because `ArrayList<Item>` does more memory reads scattered across the heap when processing the items in one node (item -> envelope then all of the doubles in the envelope are nearby, and item -> data).

The biggest gain comes from `double[] itemBounds` because it can do one hop to find the doubles to check against, and they are nearby in ram for a given node.

Moving to `Object[] itemValues` is a smaller improvement since it only saves one memory hop - it makes HPRTree run roughly 10x faster than flatbush js library on their [posted benchmarks](https://github.com/mourner/flatbush#performance) but that's probably because the benchmarks match a lot of items and don't do anything with them so the JVM doesn't even need to dereference the item value. In a more real-world scenario where you're looking for a needle in a haystack and then doing something with I'd imagine the difference would be smaller.

That being said JVM performance often defies my intuition and this represents the fastest I could get the benchmarks to run through mostly trial and error 😄

dr-jts:
> Moving to `Object[] itemValues` is a smaller improvement since it only saves one memory hop. It makes HPRTree run roughly 10x faster than flatbush js library on their [posted benchmarks](https://github.com/mourner/flatbush#performance) 

@msbarry did you port the Flatbush benchmark code?  If so it would be nice to include that in this PR.   Performance tests go in [this package](https://github.com/locationtech/jts/tree/master/modules/core/src/test/java/test/jts/perf), and can use the performance test harness framework via [PerformanceTestCase](https://github.com/locationtech/jts/blob/master/modules/core/src/test/java/test/jts/perf/PerformanceTestCase.java).

msbarry:
Hello, just checking in here, are there any other tests to run or changes to make before merging this?

dr-jts:
> Hello, just checking in here, are there any other tests to run or changes to make before merging this?

No, I think this looks good.  I'll merge soon.

PR #1033, opened 2024-01-25
Enable ring-endpoint-handling for LineString with isRing() == true

FObermaier:
With the current state of `TopologyPreservingSimplifier` and related classes, the `LINEARRING (220 180, 261 175, 380 220, 300 40, 140 30, 30 220, 176 176, 220 180)` simplifies to `LINEARRING (30 220, 380 220, 300 40, 140 30, 30 220)` using a tolerance value of `40`. (see [here](https://github.com/locationtech/jts/blob/7b1395b552a1cc1bede96ed23cc2c3ef390fd064/modules/core/src/test/java/org/locationtech/jts/simplify/TopologyPreservingSimplifierTest.java#L153-L159))

Performing the same operation on `LINESTRING (220 180, 261 175, 380 220, 300 40, 140 30, 30 220, 176 176, 220 180)` gives `LINESTRING (220 180, 380 220, 300 40, 140 30, 30 220, 220 180)`.

### DIFF
--- a/modules/core/src/test/java/org/locationtech/jts/simplify/TopologyPreservingSimplifierTest.java
+++ b/modules/core/src/test/java/org/locationtech/jts/simplify/TopologyPreservingSimplifierTest.java
@@ -156,6 +156,11 @@ public class TopologyPreservingSimplifierTest
         40,
         "LINEARRING (30 220, 380 220, 300 40, 140 30, 30 220)"
         );
+    checkTPS(
+      "LINESTRING (220 180, 261 175, 380 220, 300 40, 140 30, 30 220, 176 176, 220 180)",
+      40,
+      "LINESTRING (30 220, 380 220, 300 40, 140 30, 30 220)"
+    );
   }
   
   public void testPolygonKeepFlatEndpointWithTouch() throws Exception {
--- a/modules/core/src/test/java/org/locationtech/jts/simplify/TopologyPreservingSimplifierTest.java
+++ b/modules/core/src/test/java/org/locationtech/jts/simplify/TopologyPreservingSimplifierTest.java
@@ -34,90 +34,90 @@ public class TopologyPreservingSimplifierTest
   public void testPoint() {
     checkTPSNoChange("POINT (10 10)", 1);
   }
-  
+
   public void testPolygonEmpty() {
     checkTPSNoChange("POLYGON(EMPTY)", 1);
   }
-  
+
   public void testPolygonFlatVertices() throws Exception {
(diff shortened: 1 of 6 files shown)


dr-jts:
It's not clear from the description what this PR does?

Are you suggesting changing the `TopologyPreservingSimplifier` behaviour so that it simplifies the endpoint of closed `LineString`s?  If so, I disagree - the contract is to preserve the endpoints of LineStrings.

modules/core/src/test/java/org/locationtech/jts/simplify/TopologyPreservingSimplifierTest.java
```diff
@@ -141,36 +141,41 @@ public void testNoCollapseSmallSquare() throws Exception {
...
-  
+
   public void testLinearRingRemoveEndpoint() throws Exception {
     checkTPS(
       "LINEARRING (220 180, 261 175, 380 220, 300 40, 140 30, 30 220, 176 176, 220 180)",
         40,
         "LINEARRING (30 220, 380 220, 300 40, 140 30, 30 220)"
         );
+    checkTPS(
+      "LINESTRING (220 180, 261 175, 380 220, 300 40, 140 30, 30 220, 176 176, 220 180)",
```
dr-jts, line 160: This test is intended to test behaviour on `LINEARRING` geometries.   A `LINESTRING` geometry should be in a separate test.

FObermaier:
> Are you suggesting changing the `TopologyPreservingSimplifier` behaviour so that it simplifies the endpoint of closed `LineString`s?

Yes. 

> If so, I disagree - the contract is to preserve the endpoints of LineStrings.

Take a look at the following snipplet:
```
Geometry ls = read("LINESTRING (220 180, 261 175, 380 220, 300 40, 140 30, 30 220, 176 176, 220 180)");
Geometry lr = read("LINEARRING (220 180, 261 175, 380 220, 300 40, 140 30, 30 220, 176 176, 220 180)");
assertTrue(ls.equals(lr));
assertTrue(ls.equalsExact(lr));
```
If this is true, then any operation on either of these geometries should give the same result.

dr-jts:
> If this is true, then any operation on either of these geometries should give the same result.

Not necessarily.  The OGC topological model is somewhat limited in terms of real-world requirements, IMHO.  That's why the concept of `BoundaryNodeRule` was introduced - to give more control over how endpoints are treated.  

For an example, consider simplifying a road network that contains "loops" - e.g. turn-arounds modelled as closed LineStrings.  In order to preserve network integrity, the endpoints of **all** lines must be preserved, including those of closed LineStrings.

What could be done is allow the simplify behaviour to be controlled by an option flag.  This takes effort and complicates the API, so is there a clear practical use case for this?

FObermaier:
You have a point regarding `LineString`s in network models.

PR #1037, opened 2024-03-10
Enhance coverage simplification to handle one tolerance per geometry

nstrahl:
An enhancement has been made to the coverage simplification algorithm (simplify and simplifyInner). Currently, coverage simplification only allows one global tolerance value to be specified for the entire coverage. The proposed enhancement allows multiple tolerances to be passed in as a list, where each entry corresponds to a geometry. The geometries can thus be simplified with different tolerances while still ensuring topological correctness. At coinciding coverage edges the lowest available value is utilized.
 
This will be very useful for solving problems related to the boundary discretization of tessellated geometries.

The following figure shows an example where the boundary polygon is simplified with a tolerance of 1, and the other hole polygons with 1 and 0.5.

coverage1

Here is another one with tolerances of 50 and 10.

coverage2

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/coverage/CoverageRingEdges.java
+++ b/modules/core/src/main/java/org/locationtech/jts/coverage/CoverageRingEdges.java
@@ -54,13 +54,13 @@ class CoverageRingEdges {
    * @param tolerances the simplification tolerances for each geometry
    * @return the edges of the coverage
    */
-  public static CoverageRingEdges create(Geometry[] coverage, List<Double> tolerances) {
+  public static CoverageRingEdges create(Geometry[] coverage, Double[] tolerances) {
     CoverageRingEdges edges = new CoverageRingEdges(coverage, tolerances);
     return edges;
   }
 
   public static CoverageRingEdges create(Geometry[] coverage) {
-    CoverageRingEdges edges = new CoverageRingEdges(coverage, new ArrayList<Double>(0));
+    CoverageRingEdges edges = new CoverageRingEdges(coverage, null);
     return edges;
   }
   
@@ -68,9 +68,9 @@ class CoverageRingEdges {
   private Map<LinearRing, List<CoverageEdge>> ringEdgesMap;
   private List<CoverageEdge> edges;
 
-  private List<Double> coverageTolerances;
+  private Double[] coverageTolerances;
 
-  public CoverageRingEdges(Geometry[] coverage, List<Double> tolerances) {
+  public CoverageRingEdges(Geometry[] coverage, Double[] tolerances) {
     this.coverage = coverage;
     ringEdgesMap = new HashMap<LinearRing, List<CoverageEdge>>();
     edges = new ArrayList<CoverageEdge>();
@@ -192,16 +192,17 @@ class CoverageRingEdges {
   }
 
   private CoverageEdge createEdge(int coverageId, Coordinate[] ring, HashMap<LineSegment, CoverageEdge> uniqueEdgeMap) {
+    boolean hasTolerance = coverageTolerances != null;
     CoverageEdge edge;
     LineSegment edgeKey = CoverageEdge.key(ring);
     if (uniqueEdgeMap.containsKey(edgeKey)) {
       edge = uniqueEdgeMap.get(edgeKey);
-      if (!coverageTolerances.isEmpty()){
-        edge.setTolerance((edge.getTolerance() < coverageTolerances.get(coverageId)) ? edge.getTolerance() : coverageTolerances.get(coverageId));
+      if (hasTolerance){
+        edge.setTolerance((edge.getTolerance() < coverageTolerances[coverageId]) ? edge.getTolerance() : coverageTolerances[coverageId]);
       }
     }
     else {
-      double tolerance = coverageTolerances.isEmpty() ? -1 : coverageTolerances.get(coverageId);
+      double tolerance = hasTolerance ? coverageTolerances[coverageId] : -1;
       edge = CoverageEdge.createEdge(ring, tolerance);
       uniqueEdgeMap.put(edgeKey, edge);
       edges.add(edge);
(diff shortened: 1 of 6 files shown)


modules/core/src/main/java/org/locationtech/jts/coverage/CoverageSimplifier.java
```diff
@@ -64,7 +67,37 @@ public static Geometry[] simplify(Geometry[] coverage, double tolerance) {
...
+
+  /**
+   * Simplifies the boundaries of a set of polygonal geometries forming a coverage,
+   * preserving the coverage topology.
+   *
+   * @param coverage a set of polygonal geometries forming a coverage
+   * @param tolerances comma-separated string of simplification tolerances for each coverage
+   * @return the simplified polygons
+   */
+  public static Geometry[] simplify(Geometry[] coverage, String tolerances) {
```
dr-jts, line 92: Conversion from a string to an array of tolerances should be left up to client code, since it's highly usage-specific.  It can be done in `CoverageFunctions` as well, to allow for easier testing and experimentation.

nstrahl: Ok, then I will be moving that method into the JTS test builder code

dr-jts:
The class Javadoc needs some mention of the semantics when using tolerance-per-geometry.  Specifically, there should be a comment like:
```
The simplifier allows specifying the simplification tolerance separately for each input geometry.  Shared edges are simplified using the lowest of the tolerances for the adjacent geometries (i.e. the least amount of simplification is performed). This allows specific geometries in a coverage to be simplified less than other geometries (or to force their boundary to be preserved without simplification).
```

nstrahl:
@dr-jts Any other suggestions for changes in this PR?

modules/core/src/main/java/org/locationtech/jts/coverage/CoverageSimplifier.java
```diff
@@ -64,7 +72,20 @@ public static Geometry[] simplify(Geometry[] coverage, double tolerance) {
...
+
+  /**
+   * Simplifies the boundaries of a set of polygonal geometries forming a coverage,
+   * preserving the coverage topology.
+   *
+   * @param coverage a set of polygonal geometries forming a coverage
+   * @param tolerances the simplification tolerances for each coverage
+   * @return the simplified polygons
+   */
+  public static Geometry[] simplify(Geometry[] coverage, List<Double> tolerances) {
```
dr-jts, line 84: Why is tolerances a List and not an array?  It seems more uniform for it to have the same structure as the geometries.

modules/core/src/main/java/org/locationtech/jts/coverage/TPVWSimplifier.java
```diff
@@ -123,9 +166,11 @@ private List<Edge> createEdges(MultiLineString lines, BitSet isFreeRing) {
     List<Edge> edges = new ArrayList<Edge>();
     if (lines == null)
       return edges;
+
     for (int i = 0 ; i < lines.getNumGeometries(); i++) {
       LineString line = (LineString) lines.getGeometryN(i);
       boolean isFree = isFreeRing == null ? false : isFreeRing.get(i);
+      double areaTolerance = (areaTolerances.isEmpty()) ? -1 : ((areaTolerances.size() == 1) ? areaTolerances.get(0) : areaTolerances.get(i));
```
dr-jts, line 173: Better to put this complex logic into a function

modules/core/src/main/java/org/locationtech/jts/coverage/CoverageRingEdges.java
```diff
@@ -181,29 +191,37 @@ private List<CoverageEdge> extractRingEdges(LinearRing ring,
...
   }
   
-  private CoverageEdge createEdge(Coordinate[] ring, int start, int end, HashMap<LineSegment, CoverageEdge> uniqueEdgeMap) {
+  private CoverageEdge createEdge(int coverageId, Coordinate[] ring, int start, int end, HashMap<LineSegment, CoverageEdge> uniqueEdgeMap) {
     CoverageEdge edge;
     LineSegment edgeKey = (end == start) ? CoverageEdge.key(ring) : CoverageEdge.key(ring, start, end);
     if (uniqueEdgeMap.containsKey(edgeKey)) {
       edge = uniqueEdgeMap.get(edgeKey);
+      if (!coverageTolerances.isEmpty()){
+        edge.setTolerance((edge.getTolerance() < coverageTolerances.get(coverageId)) ? edge.getTolerance() : coverageTolerances.get(coverageId));
```
dr-jts, line 219: can this logic be factored out into a function?

dr-jts:
I'm about to land a relatively complex extension to `CoverageSimplifier`.  But I'll get this one in first.  I might need to refactor it however.

dr-jts:
One of my changes involves adding another tolerance value.  Passing all this information down the call chain is getting ugly. I might factor the action of getting a tolerance value out into a strategy class.  That can also include the array of per-geom tolerances, which will make this transparent to the rest of the codebase.

dr-jts:
Actually my changes impact quite a few method signatures.  So I'll refactor mine to support a tolerance strategy class, and then that will make refactoring your code on top of that much easier.

dr-jts:
One thing I'm adding is removing small rings below the tolerance area (of MultiPolygons, or holes).  I suppose this can just use the area tolerance of the parent geometry.  Does this make sense?

nstrahl:
@dr-jts I hope these few changes help you out.

nstrahl:
> One thing I'm adding is removing small rings below the tolerance area (of MultiPolygons, or holes). I suppose this can just use the area tolerance of the parent geometry. Does this make sense?

I do think removing tiny polygons or holes is a good idea. This feature could even be activated via a user flag. This reminds me a lot of the "qgis:eliminateselectedpolygons" processing algorithm in QGIS, which actually works by forwarding the geometry checks to GEOS anyway. So it might be helpful to draw some inspiration from that algorithm.

It's nice to see this exciting new feature being improved. Is there maybe a publication describing the underlying algorithm? If so, could you forward me the DOI?

dr-jts:
> I do think removing tiny polygons or holes is a good idea. This feature could even be activated via a user flag. 

I was not planning to make this optional, but could do so.  Do you think it's useful to have both options?
 
> It's nice to see this exciting new feature being improved. Is there maybe a publication describing the underlying algorithm? If so, could you forward me the DOI?

I'll be writing a blog post soon.  It's a fairly simple extension to the current algorithm.  When I have a PR ready I'll link to this one.

dr-jts:
@nstrahl my changes for small ring removal are larger than this PR.  So I'll commit mine, and then you or I can rework this.

dr-jts:
@nstrahl I've made substantial changes to `CoverageSimplifier` - see PR #1060.

Since it was easy to add with the new code structure, I added in the functionality of this PR (individual tolerances).  Perhaps you can test out #1060?

dr-jts:
Superseded by #1060

PR #1040, opened 2024-03-14
Fix vertex-vertex intersection in OverlayArea

tanelk:
Fixes a known bug in  OverlayArea, where COLLINEAR intersections did not cancel out.


dr-jts:
Thanks for looking into this!

tanelk:
I marked it as draft - there are quite many edge cases I did not see at first. It fixes some, but breaks others.

dr-jts:
Ah, too bad.  Hopefully there's a way to resolve all the problems.

dr-jts:
So this ultimately didn't solve enough of the problems?

tanelk:
Had to go with a more involved approach in #1043

PR #1043, opened 2024-03-17
Fix vertex intersection in `OverlayArea`

tanelk:
Second try at fixing `OverlayArea`.
It seems to me, that vertex-vertex intersection could not be done, when only looking at pairs of edges. Instead all four have to be looked at together.  

For this I replaced `STRtree` with `SegmentSetMutualIntersector`, because it has access to previous and next segments on both lines.

Added several new test cases and moved common ones into a shared base for simple and regular overlay area tests.

Covers #1040 and #1042


tanelk:
Did not change `SimpleOverlayArea`, will get to it after initial review.  Perhaps this approach is not OK.
`SimpleOverlayAreaTest` will fail until so.

tanelk:
I went ahead and fixed the `SimpleOverlayArea` also. The tests should now all pass

tanelk:
Pinging @dr-jts for a review on this.

PR #1045, opened 2024-04-05
Fix Douglas Peucker simplifier for XYZM coordinates

### LINKED ISSUE #1044: Missing dimensions after simplifying with Douglas Peucker
Currently the `DouglasPeuckerSimplifier.simplify(geometry, simplificationTolerance)` with 4 dimensions it returns a new geometry simplified but missing the `m` dimension.

It looks like the issue is happening around this [line](https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/simplify/DouglasPeuckerLineSimplifier.java#L71), when creating a new regular coordinate instead of just copying the original one. Opening a PR soon trying to address it.

dsantosmerino-wkl:
Resolves #1044 

As explained in the issue, it looks like the current implementation removes the `m` dimension. To resolve this issue, we have modified the implementation to ensure that the original coordinate, including its dimension, is preserved.


dr-jts:
Looks good, thanks.

PR #1052, opened 2024-05-13
RelateNG algorithm for DE-9IM relationship evaluation

dr-jts:
This PR contains the new RelateNG algorithm for evaluating the DE-9IM matrix and topological predicates for geometry pairs.

The RelateNG algorithm has the following capabilities:
 * **Efficient short-circuited evaluation** of topological predicates (including matching custom DE-9IM matrix patterns)
* **Optimized repeated evaluation** of predicates against a single geometry via cached spatial indexes (AKA "prepared mode")
* **Robust computation** (only point-local topology is computed, so invalid geometry topology does not cause failures)
* `GeometryCollection` inputs containing **mixed types** and **overlapping polygons** are supported, using _union semantics_.
* **Zero-length LineStrings** are treated as being topologically identical to Points.
 * Support for `BoundaryNodeRule`s.

Initially this is provided as a user-callable option.  A subsequent version will use this algorithm for `Geometry` methods. It will also replace the `PreparedGeometry` implementation (with the API preserved for some number of versions).

This algorithm fixes the issues summarized in https://github.com/libgeos/geos/issues/1060.


modules/core/src/test/java/org/locationtech/jts/operation/relateng/RelatePointLocatorTest.java
```diff
@@ -0,0 +1,87 @@
...
+  public RelatePointLocatorTest(String name) {
+    super(name);
+  }
+  
+  String gcPLA = "GEOMETRYCOLLECTION (POINT (1 1), POINT (2 1), LINESTRING (3 1, 3 9), LINESTRING (4 1, 5 4, 7 1, 4 1), LINESTRING (12 12, 14 14), POLYGON ((6 5, 6 9, 9 9, 9 5, 6 5)), POLYGON ((10 10, 10 16, 16 16, 16 10, 10 10)), POLYGON ((11 11, 11 17, 17 17, 17 11, 11 11)), POLYGON ((12 12, 12 16, 16 16, 16 12, 12 12)))";
+  
+  public void testPoint() {
+    //String wkt = "GEOMETRYCOLLECTION (POINT(0 0), POINT(1 1))";
+    checkLocation(gcPLA, 1, 1, DimensionLocation.POINT_INTERIOR);
+    checkLocation(gcPLA, 0, 1, Location.EXTERIOR);
```
pramsey, line 36: I don't thing this test comparing a Location to a DimensionLocation makes sense

dr-jts: In fact `DimensionLocation.EXTERIOR = Location.EXTERIOR` (see [here](https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/operation/relateng/DimensionLocation.java#L26)).  So this is techically correct - but it would be better to use the `DimensionLocation` value.   I'll fix that.

Also, would be better if the check method was called `checkDimLoc` to be clear - I'll change that too.

dr-jts: Fixed: https://github.com/locationtech/jts/commit/8c71ae7b56f600ed05b87c9584ae04fe33d154dd

PR #1060, opened 2024-06-19
Improve CoverageSimplifier

dr-jts:
Improves `CoverageSimplier`:

* Adds small ring removal, with configurable tolerance factor
* Adds smoothness weighting, to provide smoother results
* Adds ability to specify separate tolerances for inner and outer edges
* Adds support for a separate tolerance for each input coverage element

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/coverage/CoverageSimplifier.java
+++ b/modules/core/src/main/java/org/locationtech/jts/coverage/CoverageSimplifier.java
@@ -224,10 +224,10 @@ public class CoverageSimplifier {
     // assert: index0 >= 0
     double tolerance = tolerances[index0];
     
-    int index1 = covEdge.getAdjacentIndex(0);
-    if (index1 >= 0) {
+    if (covEdge.hasAdjacentIndex(1)) {
+      int index1 = covEdge.getAdjacentIndex(1);
       double tol1 = tolerances[index1];
-      //-- minimum tolerance is used
+      //-- use lowest tolerance for edge
       if (tol1 < tolerance)
         tolerance = tol1;
     }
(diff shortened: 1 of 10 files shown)


modules/core/src/main/java/org/locationtech/jts/coverage/CoverageSimplifier.java
```diff
@@ -58,93 +72,209 @@ public class CoverageSimplifier {
...
-    BitSet freeRings = getFreeRings(edges);
-    MultiLineString linesSimp = TPVWSimplifier.simplify(lines, freeRings, constraints, tolerance);
-    //Assert: mlsSimp.getNumGeometries = edges.length
+  private double computeTolerance(CoverageEdge covEdge, double[] tolerances) {
+    int index0 = covEdge.getAdjacentIndex(0);
+    // assert: index0 >= 0
+    double tolerance = tolerances[index0];
     
-    setCoordinates(edges, linesSimp);
+    int index1 = covEdge.getAdjacentIndex(0);
```
nstrahl, line 227: int index1 = covEdge.getAdjacentIndex(1);

nstrahl (changes requested):
There was an error where only one tolerance was considered for each coverage edge

nstrahl:
I have a suggestion here.

Take the geometry from example testInnerHoleTouchingShell.
 
image

If the tolerances 0,500,500 are provided then the holes are not simplified. I think it should be possible to coarsen hole or island polygons since they do not have any significant overlap with the exterior ring of the geometry enclosing them. Therefore we should probably skip the tolerance comparisson in such a case. A flag could be passed (e.g. forceHoleTolerances) to make this optional.

modules/core/src/main/java/org/locationtech/jts/coverage/CoverageSimplifier.java
dr-jts: Good catch.  Need a unit test to verify that.

nstrahl (approved):

PR #1062, opened 2024-07-04
Fix KMLReader to read coordinates with whitespace

tohidemyname:
KMLReader will raise "Invalid coordinate format" error for coordinates with extra whitespace.
Example file: https://developers.google.com/kml/documentation/KML_Samples.kml

Added line to trim and replace whitespace in coordinates string before processing.


dr-jts:
Thanks!

Issue #375, opened 2019-02-18
MaximalEdgeRing computeRing producing inconsistent CoordinateSequence


jodygarnett:
Follow on to #374 we have found GeoTools test cases producing inconsistent coordinate sequence through a sequence of JTS calls. We suspect one of the methods is calling new Coordinate() rather than relying on CoordianteSequence.createCoordinate().

The net effect is closing a Linear Ring with a Coordinate when all the other coordinates are CoordinateXY.

dr-jts:
There is probably no test case currently which creates this combination of coordinate arrays.  One should be added, at least to test `intersection`.

There may well be other places where JTS is creating Coordinates directly - these should be tracked down as well.

dr-jts:
See also this comment: https://github.com/locationtech/jts/issues/370#issuecomment-468926825

PR #1071, opened 2024-08-23
RobustLineIntersector defensive copyWithZ check to avoid exception with CoordinateXY or CoordinateXYM

jodygarnett:
Defensive check prevents copyWithZ, copyWithZInterpolate methods from throwing an exception when the target coordinate is CoordinateXY or CoordinateXYM.

This is in addition to the existing defensive check to avoid doing anything if the z vale was Double.NaN.

This mitigates but does not solve https://github.com/locationtech/jts/issues/375 which is due to the intermixing of Coordinate and CoordinateXY values in GeoTools clipping code.


jodygarnett:
This is minimal change to avoid exception. 

I looked at "casting" coordinates to solve https://github.com/locationtech/jts/issues/375 but am uncomfortable doing so in performance sensitive code.

dr-jts:
Looks good to me.

Issue #733, opened 2021-05-26
Writing and reading WKB loses M dimension


kaiwinter:
When using `WKBWriter` to write a Geometry which has XYM coordinates and afterwards using `WKBReader` for reading the result. The Geometry has XYZ coordinates instead of XYM coordinates.

Here is small example:
```
    String wkt = "MULTILINESTRING M((1 1 1, 2 2 2))";
    WKTReader wktReader = new WKTReader();
    Geometry geometryBefore = wktReader.read(wkt);

    WKBWriter wkbWriter = new WKBWriter(3);
    byte[] write = wkbWriter.write(geometryBefore);

    WKBReader wkbReader = new WKBReader();
    Geometry geometryAfter = wkbReader.read(write);

    System.out.println(Arrays.asList(geometryBefore.getCoordinates()));
    System.out.println(Arrays.asList(geometryAfter.getCoordinates()));
```

When running this you get the following output:
```
[(1.0, 1.0 m=1.0), (2.0, 2.0 m=2.0)]
[(1.0, 1.0, 1.0), (2.0, 2.0, 2.0)]

```
The `geometryBefore` has XYM coordinates. The `geometryAfter` has XYZ coordinates. 
In [`WKBReader.readGeometry`](https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/io/WKBReader.java#L203) are two variables evaluated [`hasZ`](https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/io/WKBReader.java#L241) and [`hasM`](https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/io/WKBReader.java#L244). In the above example `hasZ` is true and `hasM` is false. I think this is because [`WKBWriter.writeGeometryType`](https://github.com/locationtech/jts/blob/master/modules/core/src/main/java/org/locationtech/jts/io/WKBWriter.java#L405) always writes `0x80000000` regardless of the coordinate type. Debugging `WKBReader.readGeometry` makes me think `0x40000000` should be used for M coordinates in `WKBWriter.writeGeometryType`.

What do you think? Can this be improved or did I just misunderstood how this should work.

jnh5y:
I think you've identified a bug here.  Any interest in working on it?

The Contributing guide has some info about that here: https://github.com/locationtech/jts/blob/master/CONTRIBUTING.md

kaiwinter:
[...]

PR #734, opened 2021-05-29
Keep M coordinates when writing to WKB and reading from WKB

kaiwinter:
The changes I made (adding `ordinateFlags`) are mainly taken over from the `WKTReader`/`WKTWriter`.

PR #1092, opened 2024-10-17
Fix writing XYM geometries as WKB

Kontinuation:
Fixes https://github.com/locationtech/jts/issues/733

Please note that we must specify `outputDimension = 4` when writing geometries with M dimension, even when the geometry is XYM instead of XYZM. This behavior is the same as WKTWriter.

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/io/CheckOrdinatesFilter.java
+++ b/modules/core/src/main/java/org/locationtech/jts/io/CheckOrdinatesFilter.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright (c) 2016 Vivid Solutions.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * and Eclipse Distribution License v. 1.0 which accompanies this distribution.
+ * The Eclipse Public License is available at http://www.eclipse.org/legal/epl-v20.html
+ * and the Eclipse Distribution License is available at
+ *
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ */
+package org.locationtech.jts.io;
+
+import org.locationtech.jts.geom.CoordinateSequence;
+import org.locationtech.jts.geom.CoordinateSequenceFilter;
+
+import java.util.EnumSet;
+
+/**
+ * A filter implementation to test if a coordinate sequence actually has meaningful values for an
+ * ordinate bit-pattern
+ */
+class CheckOrdinatesFilter implements CoordinateSequenceFilter {
+
+  private final EnumSet<Ordinate> checkOrdinateFlags;
+  private final EnumSet<Ordinate> outputOrdinates;
+
+  /**
+   * Creates an instance of this class
+   *
+   * @param checkOrdinateFlags the index for the ordinates to test.
+   */
+  CheckOrdinatesFilter(EnumSet<Ordinate> checkOrdinateFlags) {
+
+    this.outputOrdinates = EnumSet.of(Ordinate.X, Ordinate.Y);
+    this.checkOrdinateFlags = checkOrdinateFlags;
+  }
+
+  /**
+   * @see CoordinateSequenceFilter#isGeometryChanged
+   */
+  public void filter(CoordinateSequence seq, int i) {
+
+    if (checkOrdinateFlags.contains(Ordinate.Z) && !outputOrdinates.contains(Ordinate.Z)) {
+        if (!Double.isNaN(seq.getZ(i))) {
+            outputOrdinates.add(Ordinate.Z);
+        }
+    }
+
+    if (checkOrdinateFlags.contains(Ordinate.M) && !outputOrdinates.contains(Ordinate.M)) {
+        if (!Double.isNaN(seq.getM(i))) {
+            outputOrdinates.add(Ordinate.M);
+        }
+    }
+  }
+
+  /**
+   * @see CoordinateSequenceFilter#isGeometryChanged
+   */
+  public boolean isGeometryChanged() {
+    return false;
+  }
+
+  /**
+   * @see CoordinateSequenceFilter#isDone
+   */
+  public boolean isDone() {
+    return outputOrdinates.equals(checkOrdinateFlags);
+  }
+
+  /**
+   * Gets the evaluated ordinate bit-pattern
+   *
+   * @return A bit-pattern of ordinates with valid values masked by {@link #checkOrdinateFlags}.
+   */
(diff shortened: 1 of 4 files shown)


modules/core/src/main/java/org/locationtech/jts/io/CheckOrdinatesFilter.java
```diff
@@ -0,0 +1,79 @@
+/*
+ * Copyright (c) 2016 Vivid Solutions.
```
mprins, line 2: this is incorrect for a new file

Kontinuation: All source code in this file is taken from `WKTWriter.java`, which is an old file. Should we update the copyright header or preserve the copyright header of the file where it originates from?

Kontinuation:
Issue https://github.com/locationtech/jts/issues/733 is closed but the problem still exists in the latest master. The PR https://github.com/locationtech/jts/pull/734 mentioned in the issue is for resolving a similar problem for WKB, not WKT.

jiayuasu:
@jodygarnett @dr-jts Would you please merge this PR? We are working on adding Geo support into Apache [Iceberg](https://github.com/apache/iceberg/pull/12346) and Parquet Java reader and writer. JTS is a fundamental lib in this process. WKB is the encoding mechanism used in all these PRs.

jiayuasu:
@jnh5y CC James too 🙏🏻

Kontinuation:
I'd like to provide more context for the problem we are solving. Given the following `MULTILINESTRING M` geometry:

```java
String wkt = "MULTILINESTRING M((1 1 1, 2 2 2))";
WKTReader wktReader = new WKTReader();
Geometry geometryBefore = wktReader.read(wkt);
```

Writing it using WKBWriter and read it back using WKBReader will result in a `MULTILINESTRING Z`:

```java
WKBWriter wkbWriter = new WKBWriter(3);
byte[] write = wkbWriter.write(geometryBefore);
WKBReader wkbReader = new WKBReader();
Geometry geometryAfter = wkbReader.read(write);
System.out.println(Arrays.asList(geometryAfter.getCoordinates())); // [(1.0, 1.0, 1.0), (2.0, 2.0, 2.0)]
```

If we use WKTWriter and WKTReader for serializing and deserializing the `MULTILINESTRING M` geometry, we don't see the "phantom" Z ordinate, but we still lose the M ordinate:

```java
WKTWriter wktWriter = new WKTWriter(3);
String write = wktWriter.write(geometryBefore);
Geometry geometryAfter = wktReader.read(write);
System.out.println(Arrays.asList(geometryAfter.getCoordinates())); // [(1.0, 1.0, NaN), (2.0, 2.0, NaN)]
```

We found that JTS will only output M ordinate when [`outputDimension` > 3](https://github.com/locationtech/jts/blob/1.20.0/modules/core/src/main/java/org/locationtech/jts/io/WKTWriter.java#L279-L280), so M is always treated as the 4th dimension, regardless of whether the coordinates are XYM or XYZM. Setting `outputDimension` as 4 when using WKTWriter generates the `MULTILINESTRING M(...)` WKT string we wanted:

```java
WKTWriter wktWriter = new WKTWriter(4);
String write = wktWriter.write(geometryBefore);
System.out.println(write);  // MULTILINESTRING M((1 1 1, 2 2 2))
Geometry geometryAfter = wktReader.read(write);
System.out.println(Arrays.asList(geometryAfter.getCoordinates())); // [(1.0, 1.0 m=1.0), (2.0, 2.0 m=2.0)]
```

I believe that WKBWriter should behave the same as WKTWriter. However, specifying `outputDimensions` as 4 results in the following exception:

```java
WKBWriter wkbWriter = new WKBWriter(4);
byte[] write = wkbWriter.write(geometryBefore);  // <- exception raised here
// java.lang.IllegalArgumentException: Invalid ordinate index: 3
//
//	at org.locationtech.jts.geom.CoordinateXYM.getOrdinate(CoordinateXYM.java:140)
//	at org.locationtech.jts.geom.impl.CoordinateArraySequence.getOrdinate(CoordinateArraySequence.java:260)
//	at org.locationtech.jts.io.WKBWriter.writeCoordinate(WKBWriter.java:524)
//	at org.locationtech.jts.io.WKBWriter.writeCoordinateSequence(WKBWriter.java:502)
//	at org.locationtech.jts.io.WKBWriter.writeLineString(WKBWriter.java:425)
//	at org.locationtech.jts.io.WKBWriter.write(WKBWriter.java:388)
//	at org.locationtech.jts.io.WKBWriter.writeGeometryCollection(WKBWriter.java:454)
//	at org.locationtech.jts.io.WKBWriter.write(WKBWriter.java:395)
//	at org.locationtech.jts.io.WKBWriter.write(WKBWriter.java:367)
```

This PR resolves the above issue by making the behavior of WKBWriter consistent with WKTWriter. No Z or M ordinates will be written with `outputDimension = 3`:

```java
WKBWriter wkbWriter = new WKBWriter(3);
byte[] write = wkbWriter.write(geometryBefore);
WKBReader wkbReader = new WKBReader();
Geometry geometryAfter = wkbReader.read(write);
System.out.println(Arrays.asList(geometryAfter.getCoordinates())); // [(1.0, 1.0), (2.0, 2.0)]
```

The M ordinate will be written as expected with `outputDimension = 4`:

```java
WKBWriter wkbWriter = new WKBWriter(4);
byte[] write = wkbWriter.write(geometryBefore);
WKBReader wkbReader = new WKBReader();
Geometry geometryAfter = wkbReader.read(write);
System.out.println(Arrays.asList(geometryAfter.getCoordinates())); // [(1.0, 1.0 m=1.0), (2.0, 2.0 m=2.0)]
```

jiayuasu:
@dr-jts @jodygarnett @jnh5y 

Hi Martin, Jody, and James,

This is Jia from Wherobots, Apache Sedona.

Iceberg and Parquet have adopted WKB as the default encoding for geometry, and we’re currently working on adding JTS as a dependency of Iceberg and Parquet geometry. This change will have a broad impact across the ecosystem.

Could you please review and consider merging this PR? Or let us know what’s needed to get it across the finish line?

Thanks!
Jia

modules/core/src/main/java/org/locationtech/jts/io/CheckOrdinatesFilter.java
jodygarnett: I do not think it needs to hold things up much, the options are:

```
/*
 * Copyright (c) 2016 Vivid Solutions, and others
 *
```

Or explicitly:
```
/*
 * Copyright (c) 2025 your name
 * Copyright (c) 2016 Vivid Solutions
 *
```

Issue #750, opened 2021-06-29
Orientation.Index problem with COLLINEARity


FObermaier:
The following unit test fails:
```java
  public void testCollinear() {
    Coordinate p0 = new Coordinate(0, 100);
    Coordinate p1 = new Coordinate(1, 102.1082);
    Coordinate p2 = new Coordinate(3, 106.3246);

    assertEquals(Orientation.COLLINEAR, Orientation.index(p0, p1, p2));
  }
```

I'm aware that this is due to floating point arithmetic, so would it be acceptable to add the following overload of `Orientation.index`:
```java
  /**
   * Returns the orientation index of the direction of the point <code>q</code> relative to
   * a directed infinite line specified by <code>p1-p2</code>.
   * The index indicates whether the point lies to the {@link #LEFT} or {@link #RIGHT}
   * of the line, or lies on it {@link #COLLINEAR}.
   * The index also indicates the orientation of the triangle formed by the three points
   * ( {@link #COUNTERCLOCKWISE}, {@link #CLOCKWISE}, or {@link #STRAIGHT} )
   * 
   * @param p1 the origin point of the line vector
   * @param p2 the final point of the line vector
   * @param q the point to compute the direction to
   * @param eps a threshold value for the determinat to be meaningful
   * 
   * @return -1 ( {@link #CLOCKWISE} or {@link #RIGHT} ) if q is clockwise (right) from p1-p2;
   *         1 ( {@link #COUNTERCLOCKWISE} or {@link #LEFT} ) if q is counter-clockwise (left) from p1-p2;
   *         0 ( {@link #COLLINEAR} or {@link #STRAIGHT} ) if q is collinear with p1-p2
   */
  public static int index(Coordinate p1, Coordinate p2, Coordinate q, double eps)
```

dr-jts:
It's worth noting that this could be due to **either**:
* robustness failure in the DD determinant computation
* imprecise conversion of the decimal values to DD floating point values

Either way the effect is the same.

Do you need to define what the epsilon value applies to?  Is it the value of the determinant, or the ordinate values?  That can affect the size of the provided value.  The method doc should discuss this, I think.

What is the algorithm you are suggesting be used to evaluate the fuzzy index value?   

How are you intending to use this method in computation?

I do think something like this will be needed.  I've been thinking about developing spatial predicates which will allow a tolerance value.  Although I was thinking more along the lines of adding some snapping, rather than a tolerance on the basic predicates.

[...]

PR #1093, opened 2024-10-28
Change Shewchuk Orientation filter to Ozaki et al. filter.

tinko92:
This commit changes the current Shewchuk-based (notably, the original code doesn't use Shewchuk's constant which is significantly narrower) to the filter published by Ozaki et al. in https://doi.org/10.1007/s10543-015-0574-9 . It is tighter, so it has strictly fewer misses and it has fewer and better predictable branches and fewer instructions so it is theoretically and practically faster (the referenced paper has benchmarks).

It does not come with new tests because it should not change any observable behavior by design except possibly speed, which is not tested, so all tests testing the current predicate are tests for the patched implementation.

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/algorithm/CGAlgorithmsDD.java
+++ b/modules/core/src/main/java/org/locationtech/jts/algorithm/CGAlgorithmsDD.java
@@ -109,12 +109,6 @@ public class CGAlgorithmsDD
   }
 
   /**
-   * A value which is safely greater than the
-   * relative round-off error in double-precision numbers
-   */
-  private static final double DP_SAFE_EPSILON = 1e-15;
-
-  /**
    * A filter for computing the orientation index of three coordinates.
    * <p>
    * If the orientation can be computed safely using standard DP
@@ -126,7 +120,8 @@ public class CGAlgorithmsDD
    * avoid the use of slower robust methods except when they are really needed,
    * thus providing better average performance.
    * <p>
-   * Uses an approach due to Jonathan Shewchuk, which is in the public domain.
+   * Uses an approach due to Ozaki et al., which is published at 
+   * <a href="https://doi.org/10.1007/s10543-015-0574-9">doi:10.1007/s10543-015-0574-9</a>.
    * 
    * @param pax A coordinate
    * @param pay A coordinate
@@ -140,34 +135,12 @@ public class CGAlgorithmsDD
   private static int orientationIndexFilter(double pax, double pay,
       double pbx, double pby, double pcx, double pcy) 
   {
-    double detsum;
-
     double detleft = (pax - pcx) * (pby - pcy);
     double detright = (pay - pcy) * (pbx - pcx);
     double det = detleft - detright;
 
-    if (detleft > 0.0) {
-      if (detright <= 0.0) {
-        return signum(det);
-      }
-      else {
-        detsum = detleft + detright;
-      }
-    }
-    else if (detleft < 0.0) {
-      if (detright >= 0.0) {
-        return signum(det);
-      }
-      else {
-        detsum = -detleft - detright;
-      }
-    }
-    else {
-      return signum(det);
-    }
-
-    double errbound = DP_SAFE_EPSILON * detsum;
-    if ((det >= errbound) || (-det >= errbound)) {
+    double errbound = Math.abs(detleft + detright) * 3.3306690621773724e-16;
+    if (Math.abs(det) >= errbound) {
       return signum(det);
     }
 


dr-jts:
Where does the value 3.3306690621773714e-16 come from?  I don't see it in the Ozaki paper (source from [here](https://www.tuhh.de/ti3/paper/rump/OzBueOgOiRu15.pdf)).

dr-jts:
See also https://github.com/libgeos/geos/pull/1184

dr-jts:
I wonder if this solves #750?

tinko92:
> I wonder if this solves #750?
> 
> Specifically, what is the orientation index computed for
> 
> ```
> LINESTRING (0 100, 3 106.3246)
> 
> POINT (1 102.1082)
> ```
> 
> Currently this is computed as -1, but by inspection it should be 0 (collinear).
> 
> If so, this case can be added as a unit test.

-1 is correct for this test case if the numbers are read as "nearest double to 106.3246" and "nearest double to 102.1082", which is how a compiler or text parser will read them, here is a visualization of the double values close to that: 
image

Blue represents -1, Red represents 1, Green represents 0. Notably, there is no green in this picture because the exact colinear point at x=1 is not representable in double precision. The pixels are not quadratic because at 1 and 100, the exponents of the x and y coordinates are different.

I can see how this is an unexpected result for library users, though. I see three possible solutions to solve this:

- Use this exact predicate for construction to guarantee algorithmic correctness.
- Use another rounding predicate just for simple point to linear tests to be more in line with user expectation where inconsistent individual predicate results cannot mess up the consistency of a more complicated datastructure like a computed intersection polynomial or a triangulation. In this case in the filter failure case 0 should be returned and the constant should be changed from ~3e-16 to roughly ~5e-16 (can try to compute the exact constant later) to guarantee that including the initial rounding error from a decimal input, "collinear up to rounding" points are recognised as such after the accumulated error of the determinant computation. This would also find some points collinear whose decimal representations are not collinear and will produce subtly inconsistent results between different algorithm.
- Use a decimal number type and extended precision arithmetic at the cost of a severe performance penalty (I'd expect at least 1 order of magnitude slower operations for everything where predicate computation cost is not negligible like triangulations).

dr-jts:
@tinko92 Great analysis, thanks.  It sounds like there is no easy/fast solution to this problem.  That's not a surprise, given the well-known issues of geometric robustness when using floating-point.  We'll just have document this situation, and ideally provide tolerance-based algorithms as a work-around.

grootstebozewolf:
I've machine-checked soundness proofs for both the Shewchuk-based 
and Ozaki γ₂ filter thresholds in Rocq/Flocq:

  b64_orient_sign_stage_d_sound (Stage D, expansion arithmetic)
  b64_ozaki_filter_sound (Ozaki γ₂ threshold, this PR's approach)

Both theorems confirm the same interpretation documented in this 
discussion: "correct sign of the exact determinant of the actual 
double inputs", not "collinear up to decimal rounding".

Corpus: https://github.com/grootstebozewolf/NetTopologySuite.Proofs
Relevant file: theories-flocq/Orient_b64_Ozaki.v

micycle1:
A relevant 2023 paper, if this topic is revisited: [s10543-023-00975-x.pdf](https://github.com/user-attachments/files/29918531/s10543-023-00975-x.pdf).

> More recently, Ozaki et al. developed an improved static filter as well as a new semi-static filter for the 2D orientation predicate, where the latter also handles floating-point exceptions such as overflow and underflow. Their filters, however, are not designed for arbitrary polynomial predicates.

image

grootstebozewolf:
Thanks for this — differential check from the NetTopologySuite.Proofs side (precision / orientation lane).

We mirrored **three** Stage A shapes in pure double and gated them against an independent extracted exact orient (`ORIENT_EXACT` / `b64_orient2d_exact`). We did **not** re-run the full Maven suite or claim bit-identity of the DD fallback with the oracle.

### Method
- **Oracle GT:** `ORIENT_EXACT` (full-plane exact sign).
- **JTS-master mirror:** current `orientationIndexFilter` — Shewchuk-style detsum + `DP_SAFE_EPSILON = 1e-15`.
- **Ozaki mirror (this PR):** `errbound = |detleft + detright| * 3.3306690621773724e-16`, CERTAIN iff `|det| >= errbound`.
- **Corpus Shewchuk Stage A:** detsum + `ccwerrboundA = (3+16ε)ε` (ε = 2⁻⁵³).

Write-up + table: [jts-1093-orient-lane](https://github.com/grootstebozewolf/NetTopologySuite.Proofs/blob/main/docs/jts-1093-orient-lane-2026-08.md) (`oracle/jts1093_orient_vectors.txt`, 45 vectors).

### Results
| Check | Result |
|---|---|
| Certified pin failures (EXPECTED set) | **0** |
| Ozaki **CERTAIN** vs exact sign conflicts | **0** |
| Corpus Shewchuk **CERTAIN** vs exact conflicts | **0** |
| JTS-master **CERTAIN** vs exact conflicts | **0** |
| Ozaki UNCERTAIN / Shewchuk UNCERTAIN / master UNCERTAIN | **10 / 10 / 14** |
| Ozaki CERTAIN while master UNCERTAIN | **4** |
| Master CERTAIN while Ozaki UNCERTAIN | **0** |
| Ozaki ↔ corpus Shewchuk certainty splits | **0** |

### “Fewer misses” band (same-sign products)

```text
a = (0, 0),  b = (1, 0.5),  q = (2, 1 + h)
```

| h | Exact | Master (1e-15) | Ozaki | Shewchuk ccwerrboundA |
|---|---|---|---|---|
| 1e-14 | ± | CERTAIN | CERTAIN | CERTAIN |
| **1e-15, 2e-15** | ± | **UNCERTAIN** | **CERTAIN** | **CERTAIN** |
| 5e-16, ~3.3e-16 | ± | UNCERTAIN | UNCERTAIN | UNCERTAIN |

So on this table the PR’s claim holds in miniature: the master constant escalates earlier; Ozaki commits correctly in the gap and never committed a wrong sign. Ozaki’s published double sits next to true Shewchuk `ccwerrboundA` (they never split certainty here). Deep collinear / zero-det cases still correctly force UNCERTAIN → DD, as designed.

### Bottom line
From an independent exact-orient gate, this filter swap looks **sound and slightly tighter** than master, and **aligned with** a Shewchuk-constant Stage A rather than the historical `1e-15` guard. Happy to re-run if you add more near-collinear fixtures to the suite.

*(Affiliation: [NetTopologySuite.Proofs](https://github.com/grootstebozewolf/NetTopologySuite.Proofs) — formal precision/predicate companion, not a JTS committer review.)*

PR #1103, opened 2024-12-19
Add GeometryCollection dimension cache

### LINKED ISSUE #1100: CoverageUnion - improve performance for huge amount of geometries
Input: 1M of rectangular polygons

org.locationtech.jts.coverage.CoverageUnion is slow because it internally calls GeometryCollection::getDimension multiple times. GeometryCollection::getDimension process all internal polygons O(n).

Simple test.
- Not suggested fix.
- Performance improvement using org.locationtech.jts.operation.overlayng.CoverageUnion. 
- !!! Package is "overlayng" !!!

current code:
CoverageUnion.union(factory.createGeometryCollection(coverage))
 - union from GeometryCollection  - O(n) for getDimension

improved performance:
CoverageUnion.union(factory.buildGeometry(coverage))
 - union from MultiPolygon - O(1) for getDimension

Suggested fix:
 - cache dimensions in org.locationtech.jts.operation.overlayng.InputGeometry
 - replace all access to Geometry::getDimension with method call

```
class InputGeometry {
 
  //private static final PointLocator ptLocator = new PointLocator();

  private Geometry[] geom = new Geometry[2];  // redundant array allocation
  private PointOnGeometryLocator ptLocatorA;
  private PointOnGeometryLocator ptLocatorB;
  private boolean[] isCollapsed = new boolean[2];

  private int[] dimension = new int[2]; // new code

  public InputGeometry(Geometry geomA, Geometry geomB) {
    geom = new Geometry[] { geomA, geomB };
    dimension[0] = geomA == null ? -1  : geomA.getDimension();  // new code
    dimension[1] = geomB == null ? -1  : geomB.getDimension(); 
  }
...
    public int getDimension(int index) {
        return this.dimension[index];
    }

    public boolean isArea(int geomIndex) {
        return this.getDimension(geomIndex) == 2;
    }

    public boolean isAllPoints() {
        return this.getDimension(0) == 0 && this.getDimension(1) == 0; // redundant: this.geom[1] != null 
    }

    public boolean hasEdges(int geomIndex) {
        return this.getDimension(geomIndex) > 0;
    }
...
}

```
Thank you

dr-jts:
Heterogeneous (mixed) `GeometryCollection`s have to use an O(n) algorithm to determine the dimensions of the elements.  This causes performance issues in algorithms which use the dimension of the collection (see #1100).

This PR solves this problem by caching the computed dimensions of a GeometryCollection.  It adds a single reference to each collection object, but the actual cache is lazily created when dimension information is accessed.

For example, for a `CoverageUnion` of a grid of 1M rectangles this improves the performance from 17 s to 370 ms (45x).

Fixes #1100

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/geom/GeometryCollectionDimension.java
+++ b/modules/core/src/main/java/org/locationtech/jts/geom/GeometryCollectionDimension.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright (c) 2024 Martin Davis.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * and Eclipse Distribution License v. 1.0 which accompanies this distribution.
+ * The Eclipse Public License is available at http://www.eclipse.org/legal/epl-v20.html
+ * and the Eclipse Distribution License is available at
+ *
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ */
+package org.locationtech.jts.geom;
+
+import java.util.Iterator;
+
+/**
+ * Computes and caches dimension information for {@link GeometryCollection}s.
+ * Optimizes performance of dimension reporting for heterogeneous collections.
+ * 
+ * @author mdavis
+ */
+class GeometryCollectionDimension {
+  private int dimension = Dimension.FALSE;
+  private boolean hasP = false;
+  private boolean hasL = false;
+  private boolean hasA = false;
+
+  public GeometryCollectionDimension(GeometryCollection coll) {
+    init(coll);
+  }
+
+  private void init(GeometryCollection coll) {
+    Iterator geomi = new GeometryCollectionIterator(coll);
+    while (geomi.hasNext()) {
+      Geometry elem = (Geometry) geomi.next();
+      if (elem.isEmpty()) 
+        continue;
+      if (elem instanceof Point) {
+        hasP = true;
+        if (dimension < Dimension.P) dimension = Dimension.P;
+      }
+      if (elem instanceof LineString) {
+        hasL = true;
+        if (dimension < Dimension.L) dimension = Dimension.L;
+      }
+      if (elem instanceof Polygon) {
+        hasA = true;
+        if (dimension < Dimension.A) dimension = Dimension.A;
+      }
+    }
+  }
+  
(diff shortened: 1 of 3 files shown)


modules/core/src/main/java/org/locationtech/jts/geom/GeometryCollectionDimension.java
```diff
@@ -0,0 +1,63 @@
...
+      }
+      if (elem instanceof Polygon) {
+        hasA = true;
+        if (dimension < Dimension.A) dimension = Dimension.A;
+      }
+    }
+  }
+  
+  public boolean hasDimension(int dim) {
+    if (dim == Dimension.A && hasA) return true;
```
dbaston, line 53: I don't understand the `dim == x` check here. If we have a mixed polygon/line collection, isn't `dim` equal to `Dimension.A` but `hasDimension(Dimension.L)` should be true?

dr-jts: `dim` is the argument to `hasDimension`.  So for `hasDimension(Dimension.L)` `dim` is 1, and that matches 
```
dim == Dimension.L && hasL
```

Actually this logic would be simpler as:
```
    if (dim == Dimension.A) return hasA;
    if (dim == Dimension.L) return hasL;
    if (dim == Dimension.P) return hasP;
    return false;
```

dr-jts: Or even better:
```
switch (dim) {
case Dimension.A: return hasA;
case Dimension.L: return hasL;
case Dimension.P: return hasP;
}
return false;
```

dbaston: Thank you, I was pretty sure I was having a brain failure of sorts.

dr-jts: See https://github.com/locationtech/jts/commit/9133b4717ce29cb20c2f21894a645ff1a44cd74f

dr-jts:
@dbaston are you able to port this to GEOS?

dbaston:
> @dbaston are you able to port this to GEOS?

Yes, https://github.com/libgeos/geos/pull/1220

Issue #1110, opened 2025-01-10
GeometrySnapper runtime dominated by slow `hypot()`.


micycle1:
PR https://github.com/locationtech/jts/pull/923 replaced replaced `Math.sqrt(x^2+y^2)` calls with `Math.hypot(x,y)`.

However this is a huge problem for things like the `GeometrySnapper` that call `Coordinate.distance()` repeatedly.

For instance, Geometry snapper is spending 83% of _its_ runtime in `hypot()` (on goem with 4k points), which is unacceptable.

![image](https://github.com/user-attachments/assets/a633ff13-4e7f-45e5-8191-9fd746ff9d20)

dr-jts:
Do you have a comparison with using `Math.sqrt(dx*dx + dy * dy)`?  That's required as a baseline to see the actual increase in time for `hypot`.

I'm fine with reverting the `hypot` change, either everywhere or in performance-critical places.  But it would be good to have metrics that confirm the issue.

micycle1:
Just pulled the code, and swapped out the hypot call.
Performance is roughly double.
Bit hard to capture since it's so call-heavy and instrumentation distorts it.

dr-jts:
Good enough for me.  I will look into reverting that PR.

dr-jts:
It's not surprising that `Math.hypot` is so much slower - it's doing a lot more work (see code [here](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/lang/FdLibm.java#L1968)).

I am thinking about creating a `MathUtil.hypot` function which is a wrapper over `Math.sqrt(x^2+y^2)`, to modularize the code and make it more self-documenting.  This could be used everywhere, or unwrapped in performance-critical areas (although I'm guessing that the JIT will make the performance pretty good.

PR #1111, opened 2025-01-11
`LineStringSnapper` Performance Improvement

micycle1:
Prompted by #1110.

commit 1 - we get a ~10% increase by removing assignments to the `seg` object and working with coordinates directly.
commit 2 - a further ~15% increase vs sqrt-based distance. Does add to the public API (because different package access required) which might not be desirable
(Also, I'm sure there are a lot of other code areas that could benefit from the same approach -- working with squared distances and avoiding `sqrt()`).


dr-jts:
Looks good, thanks.

I might do a bit of renaming/refactoring as a followup to this PR.

PR #1134, opened 2025-05-30
Fix BoundaryChainNoder to split chains at self-touch nodes

dr-jts:
Fixes the `BoundaryChainNoder` to split chains at self-touch nodes.  This is required by `CoverageUnion` to produce correct OGC topology for inverted polygons and exverted holes.

For example, the coverage:
```
GEOMETRYCOLLECTION (POLYGON ((1 9, 9 9, 9 6, 5 7, 2 6, 1 9)), POLYGON ((1 1, 1 9, 2 6, 4 3, 5 7, 7 3, 9 6, 9 1, 1 1)))
```
produces a polygon with two holes:
```
POLYGON ((9 9, 9 6, 9 1, 1 1, 1 9, 9 9), (5 7, 7 3, 9 6, 5 7), (2 6, 4 3, 5 7, 2 6))
```
image

Before this fix, the output would be an inverted polygon with a single self-touching ring.

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/noding/BasicSegmentString.java
+++ b/modules/core/src/main/java/org/locationtech/jts/noding/BasicSegmentString.java
@@ -29,6 +29,15 @@ import org.locationtech.jts.io.WKTWriter;
 public class BasicSegmentString
 	implements SegmentString 
 {
+  public static BasicSegmentString substring(SegmentString segString, int start, int end) {
+    Coordinate[] pts = new Coordinate[end - start + 1];
+    int ipts = 0;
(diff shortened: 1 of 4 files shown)


pramsey:
Answering my own question, it does the same thing as GEOS:
image

dr-jts:
> it does the same thing as GEOS:

The `TopologyException` is expected behaviour, since the input is an invalid coverage (the polygons overlap).  Are you seeing something that needs fixing in JTS?

pramsey:
It’s a change in behaviour. Previously the behavior was that invalid
coverages were just returned unchanged when fed through coversge union

PR #1136, opened 2025-06-02
Make segintersector.isDone() default to false

micycle1:
Give the SegmentIntersector interface's `isDone()` method a default value of false.

This way one can process all intersections cleanly with a lambda (rather than inner class that must implement both processIntersections() and isDone().

```
mci.process(segmentStringsB, (e0, segIndex0, e1, segIndex1) -> {
    sid.processIntersections(e0, segIndex0, e1, segIndex1);
    if (sid.hasIntersection()) {
    	// do
    }
});
```


dr-jts:
Looks good.  If this is done I'll merge.

PR #1137, opened 2025-07-08
Gh pages to docs

jodygarnett:
Looking at converting the existing PDFs using pandoc, or google docs (which offers to save as markdown).

image


dr-jts:
Good work on the PDF conversion.  The Tech Specs docx file can be removed I think?  

The two image directories have different names - is that intentional?  Can they both be just `img`?

dr-jts (changes requested):
The Tech Specs docx file can be removed I think?

The two image directories have different names - is that intentional? Can they both be just img?

dr-jts:
> * Q: How to host javadocs?

Can we include javadoc building in the doc build action?  So that the doc is generated, and then the javadoc is generated and pushed into a directory in the `gh-pages` branch.  (This is what GEOS does.)

@jodygarnett

jodygarnett:
Notes from chat:

* web/README.md has instructions for running locally
* add instructions for how to use mike to publish to gh-pages branch
* add workflow so main builds and publishes to gh-pages branch on doc change

PR #1140, opened 2025-08-08
Hausdorff distance percentile

ikgh9:
I've added the HausdorffDistancePercentile class. The percentile parameter allows for neglecting the impact of a specified percentage of the largest distances, which are treated as outliers.
I need this algorithm to provide a more flexible way of comparing line geometries and finding the best match.

The algorithm is based on the one from the HausdorffDistance class. However, there is a significant difference in the distribution of densification points. For HausdorffDistancePercentile, it is crucial to distribute these points evenly along the geometry, in other words, the distance between them should be same/similar. In the HausdorffDistance class, the distribution is proportional to the subsegment length, since each subsegment is divided by the same number.

I didn’t want to modify the existing HausdorffDistance algorithm, so I placed the version with the percentile parameter in a separate class.

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/algorithm/distance/DiscreteHausdorffDistancePercentile.java
+++ b/modules/core/src/main/java/org/locationtech/jts/algorithm/distance/DiscreteHausdorffDistancePercentile.java
@@ -2,7 +2,7 @@
  * Copyright (c) 2016 Vivid Solutions.
  *
  * All rights reserved. This program and the accompanying materials
- * are made available under the terms of the Eclipse Public License v2.0
+ * are made available under the terms of the Eclipse Public License 2.0
  * and Eclipse Distribution License v. 1.0 which accompanies this distribution.
  * The Eclipse Public License is available at http://www.eclipse.org/legal/epl-v20.html
  * and the Eclipse Distribution License is available at
--- a/modules/core/src/main/java/org/locationtech/jts/algorithm/distance/DiscreteHausdorffDistancePercentile.java
+++ b/modules/core/src/main/java/org/locationtech/jts/algorithm/distance/DiscreteHausdorffDistancePercentile.java
@@ -1,0 +1,437 @@
+/*
+ * Copyright (c) 2016 Vivid Solutions.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * and Eclipse Distribution License v. 1.0 which accompanies this distribution.
+ * The Eclipse Public License is available at http://www.eclipse.org/legal/epl-v20.html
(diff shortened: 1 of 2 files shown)


ikgh9:
Pinging @dr-jts for a review on this.

ikgh9:
@dr-jts 

Hi Martin,

This is Aleksander from HERE Technologies.

These changes are important for improving the identification of corresponding line geometries from different data sources. The Hausdorff distance is a key metric in this process, but its sensitivity to outliers in low-quality datasets can lead to inaccurate matches. Applying a percentile-based threshold to the Hausdorff distance reduces the impact of extreme deviations and improves matching accuracy, as confirmed by our tests.

We kindly ask you to merge this PR.

Thanks,
Aleksander

grootstebozewolf:
Corpus gate (NetTopologySuite.Proofs) — **GREEN**

We ran a faithful C# port of this tip against the PR unit cases plus relation pins to classic `DiscreteHausdorffDistance` and our banked discrete Hausdorff witnesses (Claim423a / `HausdorffDiscrete.v`).

**Harness** (NTS 2.6): `dotnet run --project tests/Discussion839Mre -c Release -- --jts1140` → **23 / 23 PASS**.

| Check | Result |
|---|---|
| PR unit witness lines (incl. densify offset 0.001 / 1 / 90) | PASS (`tol=1e-3`) |
| `percentile=1.0` ≡ `DiscreteHausdorffDistance.distance` (3 vertex pairs) | PASS (bit-equal) |
| Outlier case `p=0.9` spur line → distance `1.0` | PASS |
| Claim423a pair multipoints → `h=1` both ways | PASS |
| Claim423a asymmetric `{(0,0)}` vs `{(0,2),(3,0)}` → oriented `2` / `3` | PASS |
| Monotone in `p` on densified diagonal | PASS (`0 ≤ 70 ≤ 95 ≤ 100`) |

### Notes for reviewers

1. **`p = 1` recovers discrete Hausdorff** on the same sample set — good regression pin to keep next to the existing DHD tests.
2. **`p < 1` is an order statistic of min-distances**, not HKR Hausdorff and not a metric in general. Worth stating explicitly in the class javadoc so callers do not assume triangle inequality / zero-iff.
3. **Offset densify vs fraction densify** is a real semantic choice (even spacing along the whole length). Documented in the PR body; values need not match `DiscreteHausdorffDistance(fraction)`.
4. From the proofs corpus side: max-min (`p=1`) is already Qed; **no blocker for merge**. Optional follow-up: assert `distance(..., 1.0)` against classic DHD on a couple of multipoints in this PR’s test.

*(Affiliation: [NetTopologySuite.Proofs](https://github.com/grootstebozewolf/NetTopologySuite.Proofs) — formal companion, not a JTS committer review.)*

PR #1187, opened 2026-03-27
OverlayEdge: Don't skip first point when adding coordinates

### LINKED ISSUE #865: OverlayNG intersection of same Polygon rotates vertices
Using OverlayNG, the result of computing the intersection of two identical polygons has its vertices rotated by one from the inputs.  It is desirable that the vertices be identical. (This was the previous behaviour).

Reported in https://github.com/shapely/shapely/issues/1254

dbaston:
The `isFirstEdge` logic in `OverlayEdge::addCoordinates` appears to be backwards. It works out OK because it is only used to construct rings, and the dropped point ends up being added later on, with the ring being "rotated" by one vertex.

However, problems arise when extending this function to handle coordinates from `CircularString` because dropping the first point of the first arc causes a control point to be interpreted as an endpoint.

Fixes #865

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/operation/overlayng/OverlayEdge.java
+++ b/modules/core/src/main/java/org/locationtech/jts/operation/overlayng/OverlayEdge.java
@@ -147,7 +147,7 @@ class OverlayEdge extends HalfEdge {
    */
   public void addCoordinates(CoordinateList coords)
   {
-    boolean isFirstEdge = coords.size() > 0;
+    boolean isFirstEdge = coords.isEmpty();
     if (direction) {
       int startIndex = 1;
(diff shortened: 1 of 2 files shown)


dr-jts:
Nice.  This fixes #865.

It would be good to add a unit test in `OverlayNGTest`:
```
  /**
   * Tests that overlay produces stable (non-rotated) results.
   * See https://github.com/locationtech/jts/issues/865
   */
  public void testRingsNonRotated() {
    Geometry a = read("POLYGON ((1 9, 9 9, 9 1, 1 1, 1 9))");
    Geometry actual = intersection(a, a);
    checkEqualExact(a, actual);
  }
```

PR #1189, opened 2026-04-13
WIP: Add test for GEOS reported Orientation#isCCW robustness issue

strk:
See https://github.com/libgeos/geos/issues/1425


strk:
For the record, tested locally the test fails: 
> OrientationIsCCWTest.testGeosTicketGH1425:113->checkCCW:118 Coordinate array isCCW:  expected:false but was:true

strk:
@dr-jts by reading the code I was under the impression that similar case should be correctly handled but the Orientation#isCCW method, am I wrong ?

strk:
The case at hand is this (in HEXWKB form):

```0103000000010000000A000000890205230D6C31408177583619475140FD613B270D6C314031AA5436194751405D58985776943140195EE965DE47514085FFA6647B9F31408FC1C62B394851405D58985776943140195EE965DE475140FD613B270D6C314031AA543619475140A4613B270D6C31402FAA543619475140890205230D6C31408177583619475140320205230D6C31407E77583619475140890205230D6C31408177583619475140```

strk:
Simpler case:

`POLYGON ((0 0,0 2, 2 1, 2 0, 0 2, 0 0))`

that is, the next figure going clockwise starting from the bottom-left vertex is reported to be counterclockwise instead

image

dr-jts:
Both of these polygons compute as invalid.  The CCW algorithm isn't certified to work correctly on invalid geometry.

(The invalidity of the second polygon is obvious; the first one is hard to see).

PR #1210, opened 2026-06-22
Use try-with-resources in WKT and WKB readers

Oz-er:
## Summary

  This PR simplifies resource management in `WKTFileReader` and `WKBHexFileReader`.

  ## Changes

  - Replaced nested `try/finally` blocks with try-with-resources.
  - Kept reader cleanup behavior unchanged.
  - Kept the change localized to the two reader classes.

  ## Rationale

  Both classes used similar manual resource-closing logic. Try-with-resources makes the cleanup behavior clearer and
  less error-prone while preserving existing behavior.

  ## Validation

  Ran:

  ```bash
  mvn -pl modules/core verify

  Result:

  - Checkstyle: 0 violations
  - Tests run: 2294
  - Failures: 0
  - Errors: 0
  - PMD completed
  - Build successful

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/io/WKBHexFileReader.java
+++ b/modules/core/src/main/java/org/locationtech/jts/io/WKBHexFileReader.java
@@ -110,21 +110,18 @@ public class WKBHexFileReader
 	public List read() 
 	throws IOException, ParseException 
 	{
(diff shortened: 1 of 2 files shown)


dr-jts:
Good improvement, thanks.

PR #1211, opened 2026-06-28
Avoid Coordinate allocations in Area and Length sequence computations

nishantmehta:
### What

`Area.ofRingSigned(CoordinateSequence)` and `Length.ofLine(CoordinateSequence)` allocated `Coordinate` objects (three and one respectively) per call via `CoordinateSequence.createCoordinate()`, purely to read x/y values while iterating. These back `Geometry.getArea()` and `getLength()`, which are commonly called in hot paths.

This reads coordinates through the primitive `CoordinateSequence.getX(int)`/`getY(int)` accessors instead. The arithmetic is unchanged — `Area` still uses the Shoelace formula (caching the y values so each is read once), and `Length` still sums segment hypotenuses — and no `Coordinate` instances are created.

### Benchmark

Measured with a `ThreadMXBean` allocation driver (200k warmed ops):

```
Geometry.getArea     28.5 B/op -> 0 B/op
Geometry.getLength   40   B/op -> 0 B/op
```

### Testing

The full `jts-core` test suite (2294 tests) passes unchanged.

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/algorithm/Area.java
+++ b/modules/core/src/main/java/org/locationtech/jts/algorithm/Area.java
@@ -93,21 +93,15 @@ public class Area {
      * Based on the Shoelace formula.
      * http://en.wikipedia.org/wiki/Shoelace_formula
      */
-    Coordinate p0 = ring.createCoordinate();
-    Coordinate p1 = ring.createCoordinate();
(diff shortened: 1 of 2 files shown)


dr-jts:
Great, thanks.  This is definitely the best pattern to use.  The cutover to using ordinate accessors has been done only patchily, so it's good to apply it everywhere possible.

PR #1214, opened 2026-07-12
Add setRandom(Random) to RandomPointsBuilder and RandomPointsInGridBuilder

### LINKED ISSUE #1191: Allow passing a seed to RandomPointsBuilder, to enable reproducible behavior.
See, for example:
https://github.com/palisades-lakes/jts-issues/blob/main/src/main/java/jts/random/RandomPointsBuilder.java

woogi-kim:
## Summary

  Allow passing a `Random` instance to `RandomPointsBuilder` and `RandomPointsInGridBuilder` to enable reproducible
  point generation, as suggested by @dr-jts in #1191.

  ## Changes

  - Added `setRandom(Random)` setter to `RandomPointsBuilder`
  - Added `setRandom(Random)` setter to `RandomPointsInGridBuilder`
  - Replaced `Math.random()` with `random.nextDouble()` in both classes
  - Default behavior is unchanged (uses `new Random()`)
  - Added `RandomPointsBuilderTest` and `RandomPointsInGridBuilderTest`

  ## Validation

  ```bash
  mvn test -pl modules/core -Dtest="RandomPointsBuilderTest,RandomPointsInGridBuilderTest"

  - Tests run: 8
  - Failures: 0
  - Errors: 0
```

Closes #1191

### CODE REVIEW

File: modules/core/src/main/java/org/locationtech/jts/shape/random/RandomPointsBuilder.java [hunk 1]
```diff
@@ -55,6 +58,18 @@ public RandomPointsBuilder(GeometryFactory geomFact)
...
+   * This enables reproducible results by providing a {@link Random}
+   * with a fixed seed.
+   *
+   * @param random the random number generator to use
+   */
+  public void setRandom(Random random)
+  {
+  	this.random = random;
+  }
+
```
grootstebozewolf, line 72: This is the setter Martin asked for on #1191. `java.util.Random` is the right type: JTS is still Java 8, so `RandomGenerator` cannot land here. A second overload can wait for a language bump.

`setRandom(null)` NPEs on the next `getGeometry()`. Throw `IllegalArgumentException` (same style as `setExtent` rejecting a non-polygon), or document it.

Worth one javadoc sentence that a second `getGeometry()` continues the generator and does not rewind — the tests always wrap `new Random(seed)` and hide that.

modules/core/src/main/java/org/locationtech/jts/shape/random/RandomPointsInGridBuilder.java
```diff
@@ -126,18 +141,18 @@ private Coordinate randomPointInCell(double orgX, double orgY, double xLen, doub
...
+  private Coordinate randomPointInCircle(double orgX, double orgY, double width, double height)
   {
   	double centreX = orgX + width/2;
   	double centreY = orgY + height/2;
-  		
-  	double rndAng = 2 * Math.PI * Math.random();
-  	double rndRadius = Math.random();
+
+  	double rndAng = 2 * Math.PI * random.nextDouble();
+  	double rndRadius = random.nextDouble();
```
grootstebozewolf, line 155: Making this an instance method is the right cut so the circle path uses the same generator. Do not lift `Random` onto `GeometricShapeBuilder` — Hilbert / Koch / Sierpinski do not use an RNG. The duplicated setter on this sibling is correct (`RandomPointsInGridBuilder` does not extend `RandomPointsBuilder`).

Pre-existing, leave it: this still returns `new Coordinate` and skips `createCoord` / the precision model. The grid-cell path already goes through `createCoord`.

modules/core/src/test/java/org/locationtech/jts/shape/random/RandomPointsBuilderTest.java
```diff
@@ -0,0 +1,81 @@
...
+    Geometry result1 = createPoints(42, 100);
+    Geometry result2 = createPoints(42, 100);
+    assertTrue(result1.equalsExact(result2));
+  }
+
+  public void testDifferentSeedProducesDifferentResult() {
+    Geometry result1 = createPoints(42, 100);
+    Geometry result2 = createPoints(99, 100);
+    assertFalse(result1.equalsExact(result2));
+  }
```
grootstebozewolf, line 57: Same-seed / different-seed plus the polygonal-mask case below is enough to lock #1191. `testDifferentSeedProducesDifferentResult` is theoretically flaky; with 100 points it will not fire in practice.

Unseeded default is still non-reproducible. The stream is no longer the process-global `Math.random()` holder — each builder has its own `Random` — which is the better default and does not change the public contract.

modules/core/src/test/java/org/locationtech/jts/shape/random/RandomPointsInGridBuilderTest.java
```diff
@@ -0,0 +1,81 @@
...
+    Geometry result1 = createGridPoints(42, 25);
+    Geometry result2 = createGridPoints(99, 25);
+    assertFalse(result1.equalsExact(result2));
+  }
+
+  public void testReproducibleWithConstrainedToCircle() {
+    Geometry result1 = createGridPointsInCircle(42, 25);
+    Geometry result2 = createGridPointsInCircle(42, 25);
+    assertTrue(result1.equalsExact(result2));
+  }
```
grootstebozewolf, line 63: Good that constrained-to-circle is pinned. That is the path that used to call `Math.random()` from a `static` helper.

modules/core/src/main/java/org/locationtech/jts/shape/random/RandomPointsBuilder.java
grootstebozewolf, line 1: Both builders needed this (micycle1 had already forked the grid one on #1191). License headers and DCO are present. Ship it.

grootstebozewolf (commented):
LGTM — approve with nits.

This is the API Martin asked for on #1191: an optional `setRandom(Random)` so constructors do not explode. Both builders needed it (`RandomPointsInGridBuilder` is a sibling under `GeometricShapeBuilder`, not a subclass). JTS is still Java 8, so `java.util.Random` is the right type; `RandomGenerator` can wait for a language bump.

`randomPointInCircle` correctly stops being `static`. Tests lock same-seed equality, different-seed inequality, polygonal mask, and constrained-to-circle.

Nits (non-blocking):
- `setRandom(null)` NPEs on the next `getGeometry()` — throw `IllegalArgumentException` or document it.
- Javadoc: a second `getGeometry()` continues the generator; it does not rewind.
- Do not lift `Random` onto `GeometricShapeBuilder`.
- `randomPointInCircle` still skips `createCoord` / the precision model (pre-existing).

PR #1220, opened 2026-08-03
Add Geometry.getCoordinateDimension

### LINKED ISSUE #654: Question: Is there a way to determine geometry dimension?
The `WKBWriter` constructor expects output dimension (e.g. `new WKBWriter(dimension, ByteOrderValues.LITTLE_ENDIAN, true)`) and it has to be either 2 or 3. However, `Geometry#getDimension` is not appropriate to use here. Is there a generic way to determine the dimension for a given `Geometry` object? In other words, is there such method that satisfies the following: 

- `#getDimension(GEOMETRY_FACTORY.createPoint(new Coordinate(1,2)))` -> `2`
- `#getDimension(GEOMETRY_FACTORY.createPoint(new Coordinate(1,2,3)))` -> `3`
- `#getDimension(GEOMETRY_FACTORY.createPolygon(new Coordinate[]{new Coordinate(1,2,3), new Coordinate(2,2,3), new Coordinate(2,3,3), new Coordinate(1,2,3)}))` -> `3`
- `#getDimension(GEOMETRY_FACTORY.createMultiPolygon(new Polygon[]{GEOMETRY_FACTORY.createPolygon(new Coordinate[]{new Coordinate(1,2,3), new Coordinate(2,2,3), new Coordinate(2,3,3), new Coordinate(1,2,3)})}))` -> `3`

where `#getDimension` accepts `Geometry`.

mdedetrich:
Resolves: https://github.com/locationtech/jts/issues/654

Tests have also been added to make sure that it works as expected.

### DIFF
--- a/modules/core/src/main/java/org/locationtech/jts/geom/Geometry.java
+++ b/modules/core/src/main/java/org/locationtech/jts/geom/Geometry.java
@@ -1737,6 +1737,53 @@ public abstract class Geometry
   }
 
   /**
+   * Computes the coordinate dimension of a <code>Geometry</code>, based on the
+   * coordinate dimensions of its {@link CoordinateSequence}s
+   * (i.e. 2 for XY, 3 for XYZ or XYM, or 4 for XYZM).
+   * For a {@link GeometryCollection} this is the maximum coordinate dimension
+   * over all of its elements.
+   * Empty geometries (or empty elements of a collection) are assumed
+   * to have coordinate dimension 2, and do not affect the result
+   * unless all elements are empty.
+   *
+   * @param  g the <code>Geometry</code> to get the coordinate dimension of
+   * @return the coordinate dimension of the geometry (2, 3 or 4)
+   */
+  public static int getCoordinateDimension(Geometry g) {
+    if (g instanceof Point) {
+      return coordinateSequenceDimension(((Point) g).getCoordinateSequence());
+    }
+    if (g instanceof LineString) {
+      return coordinateSequenceDimension(((LineString) g).getCoordinateSequence());
+    }
+    if (g instanceof Polygon) {
+      Polygon poly = (Polygon) g;
+      int dimension = 2;
+      LinearRing shell = poly.getExteriorRing();
+      if (shell != null) {
+        dimension = coordinateSequenceDimension(shell.getCoordinateSequence());
+      }
+      for (int i = 0; i < poly.getNumInteriorRing(); i++) {
+        dimension = Math.max(dimension, coordinateSequenceDimension(poly.getInteriorRingN(i).getCoordinateSequence()));
+      }
+      return dimension;
+    }
+    if (g instanceof GeometryCollection) {
+      GeometryCollection gc = (GeometryCollection) g;
+      int dimension = 2;
+      for (int i = 0; i < gc.getNumGeometries(); i++) {
+        dimension = Math.max(dimension, getCoordinateDimension(gc.getGeometryN(i)));
+      }
+      return dimension;
+    }
+    return 2;
+  }
+
+  private static int coordinateSequenceDimension(CoordinateSequence seq) {
+    return seq.size() > 0 ? seq.getDimension() : 2;
+  }
+
+  /**
    * Tests whether this is an instance of a general {@link GeometryCollection},
    * rather than a homogeneous subclass.
    *
--- a/modules/core/src/test/java/org/locationtech/jts/geom/GeometryCoordinateDimensionTest.java
+++ b/modules/core/src/test/java/org/locationtech/jts/geom/GeometryCoordinateDimensionTest.java
@@ -1,0 +1,78 @@
+package org.locationtech.jts.geom;
+
+import org.locationtech.jts.geom.impl.PackedCoordinateSequenceFactory;
+import org.locationtech.jts.io.WKTReader;
+
+import junit.textui.TestRunner;
+import test.jts.GeometryTestCase;
+
+/**
+ * Tests for {@link Geometry#getCoordinateDimension(Geometry)}.
+ */
+public class GeometryCoordinateDimensionTest extends GeometryTestCase {
+  public static void main(String args[]) {
+    TestRunner.run(GeometryCoordinateDimensionTest.class);
+  }
+
+  static GeometryFactory geomFact = new GeometryFactory(PackedCoordinateSequenceFactory.DOUBLE_FACTORY);
+
+  /**
+   * A reader with the legacy "unlabelled 3rd ordinate is Z" syntax disabled,
+   * so that WKT lacking a Z or M tag is parsed as genuinely 2D.
+   */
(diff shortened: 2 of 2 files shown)


modules/core/src/main/java/org/locationtech/jts/geom/Geometry.java
```diff
@@ -1736,6 +1736,53 @@ static void checkNotGeometryCollection(Geometry g) {
...
+      GeometryCollection gc = (GeometryCollection) g;
+      int dimension = 2;
+      for (int i = 0; i < gc.getNumGeometries(); i++) {
+        dimension = Math.max(dimension, getCoordinateDimension(gc.getGeometryN(i)));
+      }
+      return dimension;
+    }
+    return 2;
+  }
+
```
grootstebozewolf, line 1781: This should be `g.getCoordinateDimension()`, not `Geometry.getCoordinateDimension(g)`.

`getDimension()` is already an instance method. GEOS — the port Dan pointed at on #654 — is `virtual uint8_t getCoordinateDimension() const` on `Point` / `SimpleCurve` / `Surface` / `GeometryCollection`. A static `instanceof` chain will not override for a later type and teaches the wrong call site. Put an instance method next to `getDimension()` and implement it on those four types.

A `CoordinateSequenceFilter` looks more “JTS,” but `Point.apply` / `LineString.apply` return without visiting a size-0 sequence, so a filter cannot see empty-sequence metadata. Direct `getCoordinateSequence()` (what you have, what GEOS has) is the right access — just not as a type switch on `Geometry`.

modules/core/src/main/java/org/locationtech/jts/geom/Geometry.java
```diff
@@ -1736,6 +1736,53 @@ static void checkNotGeometryCollection(Geometry g) {
...
+  private static int coordinateSequenceDimension(CoordinateSequence seq) {
+    return seq.size() > 0 ? seq.getDimension() : 2;
+  }
+
```
grootstebozewolf, line 1785: Do not force empty sequences to 2.

`WKTReader.createCoordinateSequenceEmpty` already builds `csFactory.create(0, toDimension(flags), measures)`. `POINT Z EMPTY` therefore already has a 0-size sequence of dimension 3. This helper throws that away. GEOS pins the same cases at 3 / 3 / 4 (`WKBReaderTest`).

The size check is papering over a different footgun: `GeometryFactory.createPoint()` uses `create(new Coordinate[]{})`, and `CoordinateArrays.dimension([])` is 3 (“unknown, assume default”). Packed’s default is also 3. Trust the sequence anyway. Document that factory-built empties may report 3. Add `POINT Z EMPTY` / `LINESTRING Z EMPTY` expecting 3.

modules/core/src/main/java/org/locationtech/jts/geom/Geometry.java
```diff
@@ -1736,6 +1736,53 @@ static void checkNotGeometryCollection(Geometry g) {
...
+      return coordinateSequenceDimension(((LineString) g).getCoordinateSequence());
+    }
+    if (g instanceof Polygon) {
+      Polygon poly = (Polygon) g;
+      int dimension = 2;
+      LinearRing shell = poly.getExteriorRing();
+      if (shell != null) {
+        dimension = coordinateSequenceDimension(shell.getCoordinateSequence());
+      }
+      for (int i = 0; i < poly.getNumInteriorRing(); i++) {
```
grootstebozewolf, line 1766: `shell != null` is dead: `Polygon` replaces a null shell with `createLinearRing()`.

modules/core/src/test/java/org/locationtech/jts/geom/GeometryCoordinateDimensionTest.java
```diff
@@ -0,0 +1,78 @@
...
+    checkDimension("LINESTRING Z (1 1 1, 2 2 2, 3 3 3)", 3);
+  }
+
+  public void testPolygonXYZ() {
+    checkDimension("POLYGON Z ((1 9 2, 9 9 2, 9 1 2, 1 1 2, 1 9 2))", 3);
+  }
+
+  public void testEmptyPoint() {
+    checkDimension("POINT EMPTY", 2);
+  }
```
grootstebozewolf, line 59: `POINT EMPTY` only locks the untagged case. Add `POINT Z EMPTY` / `LINESTRING Z EMPTY` expecting 3 (GEOS / the WKTReader empty sequence). Also run one case on the default `CoordinateArraySequence` factory — these tests are Packed-only, which hides the factory empty-default-3 footgun.

New test file also needs the license header `CONTRIBUTING` asks for.

modules/core/src/main/java/org/locationtech/jts/geom/Geometry.java
grootstebozewolf, line 1: XY / XYZ / XYM / XYZM, mixed collections taking the max, and “3 means Z or M” all match GEOS. `LinearRing` falls through `LineString`, `Multi*` through `GeometryCollection`. The strict WKT reader is the correct fixture. That is enough for the original #654 2-vs-3 examples.

This getter does not distinguish XYZ from XYM (both 3). That matches GEOS. Callers who need the ordinate set still have to look at `hasZ` / `hasM` on the sequence — fine for #654; do not claim it fully drives `WKBWriter.setOutputOrdinates`.

Empty `MultiLineString` / `MultiPolygon` from WKT already drop the Z flag in `WKTReader` (`createMultiLineString()` with no flags). Pre-existing; this method cannot invent a dimension those objects do not store.

grootstebozewolf (changes requested):
Request changes. The method is the right answer to #654 — `getDimension()` is topological, and WKB needs the coordinate count — but the API shape and the empty contract both freeze the wrong thing. Dan already pointed at the port: GEOS `Geometry::getCoordinateDimension()` is an **instance** method that trusts the sequence, including on empty.

**Blockers**
- This should be `g.getCoordinateDimension()`, not `Geometry.getCoordinateDimension(g)`. Implement it on the four types GEOS does. Do not add another 50-line `instanceof` switch to `Geometry.java`.
- Do not force empty sequences to 2. `WKTReader` already stores dimension on `POINT Z EMPTY`; GEOS reports 3 / 3 / 4. Trust `seq.getDimension()`. Document that factory-built empties may report 3 (`CoordinateArrays.dimension([])` defaults to 3). Add `POINT Z EMPTY` expecting 3.

**Nits:** dead `shell != null`; test file needs a license header; tests are Packed-only; XYZ vs XYM both being 3 matches GEOS and does not fully drive `WKBWriter.setOutputOrdinates`.

dr-jts:
I agree with the suggestions by @grootstebozewolf .  These need to be fixed before this PR can be merged.

Specifically:
* Make `getCoordinateDimension` a `Geometry` instance method
* Use `CoordinateSequence.getDimension` to determine the coordinate dimension
* Add license header to the test file

PR #1221, opened 2026-08-06
Harden GMLReader against XXE (disable DTDs and external entities)

Nexory:
## Problem
`GMLReader.read()` builds its `SAXParserFactory` with only `setNamespaceAware(false)` and `setValidating(false)`, so DOCTYPE processing and external entity resolution stay enabled. GML is routinely read from untrusted input (files, WFS responses, uploads), so a crafted document can disclose local files or trigger SSRF via an external entity (XXE):

```
<!DOCTYPE foo [ <!ENTITY xxe SYSTEM "file:///etc/hostname"> ]>
<gml:Point><gml:coordinates>&xxe;</gml:coordinates></gml:Point>
```

## Fix
Two features on the factory: `FEATURE_SECURE_PROCESSING` and `disallow-doctype-decl`. Without a DOCTYPE there is no external subset and no entity to resolve, so the three further entity features an earlier revision set were dropped as redundant; eight inputs were compared across both revisions with no difference in outcome.

This mirrors the hardening merged for the sibling `KMLReader` in #1204, which set `SUPPORT_DTD=false` and `IS_SUPPORTING_EXTERNAL_ENTITIES=false`. `GMLReader` uses SAX rather than StAX, so the equivalent is expressed as parser features. No behaviour change for valid GML. `read(String, ...)` delegates to `read(Reader, ...)`, so both public entry points are covered.

The two calls currently sit in a try/catch that logs a warning, at the request of @jodygarnett. That shape is under discussion in the review: see the open question about failing closed instead.

## Test
Two tests in `GMLReaderTest`: one asserts the referenced file content does not appear in the parsed output, on both the success and the exception path, and one asserts the DOCTYPE is rejected. Full `jts-core` suite: 2298 tests, 0 failures, unchanged against `master`.


modules/core/src/main/java/org/locationtech/jts/io/gml2/GMLReader.java
```diff
@@ -105,6 +106,14 @@ public Geometry read(Reader reader, GeometryFactory geometryFactory) throws SAXE
...
 		fact.setNamespaceAware(false);
 		fact.setValidating(false);
+		// Harden against XXE: disable DOCTYPE/DTDs and external entities (JAXP secure processing).
+		// GML input is frequently untrusted (files, WFS responses, uploads); the default SAX parser
+		// resolves external entities, enabling file disclosure and SSRF.
+		fact.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
+		fact.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
+		fact.setFeature("http://xml.org/sax/features/external-general-entities", false);
+		fact.setFeature("http://xml.org/sax/features/external-parameter-entities", false);
+		fact.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false);
```
ppkarwasz: The three additional features are over-kill: without a `DOCTYPE` declaration, there is no external subset nor external entities.

Also note that this will *fail* on Android and any JAXP implementation that doesn't support `FEATURE_SECURE_PROCESSING` (required by the JAXP specification) or `disallow-doctype-decl` (only supported by Xerces-derivatives). Even if Android's `expat` based parser does not support these feature, it is in practice safe to use, because it does not resolve entities by default.

**Shameless advertising**: we will release [Apache Commons XML](https://github.com/apache/commons-xml) in the next couple of weeks, which handles all these subtleties of JAXP implementations. The library can be both used as external dependency or shaded with `<minimizeJar>` and delegates the hassle of properly configuring an XML parser upstream.

Nexory:
Thanks, both points are right and the push has them.

On the portability one: `SAXParserFactory.setFeature` throws
`SAXNotRecognizedException` for a name the implementation does not know, and
that class extends `SAXException`, which `read()` already declares. The old code
therefore compiled and would have aborted the read at runtime, in a way the
caller cannot tell apart from malformed XML. Each feature is now applied through
a helper that ignores an unsupported one.

On the three extra features, I first wanted to keep them and argue that they
are the remaining layer where `disallow-doctype-decl` does not apply. That was
only ever measured on Xerces with the DOCTYPE feature switched off, which is not
the same thing, so I tried it on a parser that is not a Xerces derivative:
crimson 1.1.3, selected through `javax.xml.parsers.SAXParserFactory`.

| feature | Xerces (JDK 8) | crimson 1.1.3 |
|---|---|---|
| `FEATURE_SECURE_PROCESSING` | accepted | `SAXNotRecognizedException` |
| `disallow-doctype-decl` | accepted | `SAXNotRecognizedException` |
| `external-general-entities` | accepted | `SAXNotSupportedException` |
| `external-parameter-entities` | accepted | `SAXNotSupportedException` |
| `nonvalidating/load-external-dtd` | accepted | `SAXNotRecognizedException` |

All five throw there, including the two in the SAX namespace, so those three are
not a fallback on such an implementation. They are gone. What is left is
`FEATURE_SECURE_PROCESSING` and `disallow-doctype-decl`, each applied on its own.

One consequence of the guard is worth writing down, since it is the trade you
are pointing at. On crimson the same XXE payload now parses and resolves the
external entity, where the unguarded version would have thrown on the first
`setFeature`. The guard turns a loud failure into a silent absence of hardening.
That is the right call for a reader that must keep working, but it means the
hardening is only as good as the parser in use, and nothing in the API says
which one that is.

Two corrections to my own PR while I am here. The description claimed
`setFeature` only throws `SAXException` subclasses; it also declares
`ParserConfigurationException`, which is not one. And the new test asserted only
inside a `catch`, so it would have passed without running a single assertion had
`read()` returned normally. Both are fixed.

A new dependency is out of scope for this PR.

modules/core/src/main/java/org/locationtech/jts/io/gml2/GMLReader.java
```diff
@@ -105,6 +106,14 @@ public Geometry read(Reader reader, GeometryFactory geometryFactory) throws SAXE
 
 		fact.setNamespaceAware(false);
 		fact.setValidating(false);
+		// Harden against XXE. GML input is frequently untrusted (files, WFS responses,
+		// uploads) and the default SAX parser resolves external entities, which allows
+		// file disclosure and SSRF. Rejecting the DOCTYPE covers it: without one there is
+		// no internal or external subset, so no entity can be declared in the first place.
+		// Each feature is applied on its own, because a JAXP implementation which does not
+		// know one of them throws, and that must not abort the read.
+		setFeatureIfSupported(fact, XMLConstants.FEATURE_SECURE_PROCESSING, true);
```
ppkarwasz: I think:

```suggestion
		fact.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
```

was better. The rationale is that security should not be an option. If a user has Crimson on its classpath, it is fine to throw, since we don't know how to properly configure it. The user will have to say bye bye to Crimson (which BTW hasn't been maintained in 15 years).

The only parser choice that might occur on modern system and is not Xerces-based is Android's parser. However in this case you can skip the two settings since the parser does not resolve external references:

```java
if (!"org.apache.harmony.xml.parsers.SAXParserFactoryImpl".equals(fact.getClass().getName())) {
    fact.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
    fact.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
}
```

In [Commons XML](https://github.com/apache/commons-xml), we have an extensive test suite that tests an unconfigured `SAXParser` on Android API 33+ and we verified in the source code that leaving it unconfigured is fine.

This way you will have:

- Xerces-derived parsers (JDK's stock parser or external Xerces) covered,
- Android covered,
- For users of other parsers this code will throw and you'll need to adapt the code.

Nexory:
You are right, and the suggestion is in.

I had written that the guard turns a loud failure into a silent absence of
hardening, and then kept the guard anyway. Making security not optional is the
better answer to that, and skipping the one parser that is safe unconfigured is
cheaper than trying to configure everything.

Measured on the same XXE payload, guarded against unguarded:

| | Xerces (JDK 8) | crimson 1.1.3 |
|---|---|---|
| guarded | DOCTYPE rejected | parses, resolves the entity |
| unguarded, Android skipped | DOCTYPE rejected | throws while configuring |

So on a parser that cannot be configured the read now fails instead of quietly
running without the hardening.

I took the factory class name from AOSP rather than the suggestion alone:
`luni/.../org/apache/harmony/xml/parsers/SAXParserFactoryImpl.setFeature` rejects
every name outside `http://xml.org/sax/features/`, and `ExpatReader.setFeature`
returns immediately for the two SAX ones when the value is false, with a comment
that it is already the default. That matches what you describe. I have not run
this on a device, so the Android half rests on reading the source and on your
test suite, not on a measurement of mine.

`mvn clean install` on JDK 8 in a pinned container: 2467 tests, none failing.
The new test goes from 2 failures to 0 across the change.

ppkarwasz (approved):

modules/core/src/main/java/org/locationtech/jts/io/gml2/GMLReader.java
```diff
@@ -105,6 +106,19 @@ public Geometry read(Reader reader, GeometryFactory geometryFactory) throws SAXE
...
+		// Harden against XXE. GML input is frequently untrusted (files, WFS responses,
+		// uploads) and the default SAX parser resolves external entities, which allows
+		// file disclosure and SSRF. Rejecting the DOCTYPE covers it: without one there is
+		// no internal or external subset, so no entity can be declared in the first place.
+		//
+		// Android's parser rejects every feature name outside the SAX namespace and does
+		// not resolve external references in the first place, so it is skipped. Anywhere
+		// else the features are set without a guard: a parser that cannot be configured
+		// should fail here rather than parse untrusted input without the hardening.
+		if (!"org.apache.harmony.xml.parsers.SAXParserFactoryImpl".equals(fact.getClass().getName())) {
```
jodygarnett: Please do not include such verbose code comments, discussion should be left in the pull-request conversation rather than as a maintenance burden for the codebase!

I prefer to perform the above following settings in a try/catch block and log warnings if a parser such as the one you mention does not support the provided functionality.

modules/core/src/test/java/org/locationtech/jts/io/gml2/GMLReaderXXETest.java
```diff
@@ -0,0 +1,96 @@
...
+
+import junit.textui.TestRunner;
+import test.jts.GeometryTestCase;
+
+/**
+ * Tests that {@link GMLReader} does not resolve external XML entities (XXE).
+ * GML is frequently read from untrusted sources (files, WFS responses, uploads),
+ * so the underlying SAX parser must not fetch external entities or process DTDs.
+ */
+public class GMLReaderXXETest extends GeometryTestCase {
```
jodygarnett: Please add these tests to an existing test case, and reduce the level of detail / comments. We want just enough testing to show that your change does something.

We should publish an AGENTS.md to help those producing code/fixes.

jodygarnett (changes requested):
Thanks for the suggestion/improvement. Some feedback provided below.

Nexory:
Done, all three:

- the comment is one line and the reasoning moved here
- the two settings are in a try/catch that logs a warning
- the tests are in `GMLReaderTest` and the separate class is gone. I dropped the
  third one, a benign parse, since `GMLReaderTest` already covers that.

The try/catch also removed the factory class name check, so that is one less
thing to maintain.

Two things I cannot decide, both yours.

`jts-core` has no dependencies, and there is no logging anywhere under
`modules/*/src/main`, so I used `java.util.logging`. It would be the first in the
module. Happy to drop it, or to use whatever you prefer.

The other is the trade-off the try/catch makes. Measured on the same payload:

| | Xerces (JDK 8) | crimson 1.1.3 |
|---|---|---|
| try/catch and warn | DOCTYPE rejected | parses, resolves the entity |
| features set unguarded | DOCTYPE rejected | throws while configuring |

So on a parser that cannot be configured, the reader now parses untrusted GML
without the hardening and leaves a log line. Android is unaffected either way,
since its parser does not resolve external references.

While checking for duplicates I found that `KMLReader` was hardened in #1204 and
does it unguarded:

    // Disable DTDs completely (prevents DOCTYPE declarations)
    inputFactory.setProperty(XMLInputFactory.SUPPORT_DTD, false);
    // Prevent external entity expansion from DTDs
    inputFactory.setProperty(XMLInputFactory.IS_SUPPORTING_EXTERNAL_ENTITIES, false);

The APIs differ in whether they force the question: `setProperty` throws an
unchecked exception, while `setFeature` throws checked ones. `read` already
declares `SAXException` and `ParserConfigurationException`, so leaving the calls
unguarded compiles as it stands, and would match the sibling reader.

I have built it the way you asked. Say which of the two you want and I will make
it that.

`mvn clean install` on JDK 8 in a pinned container: 2467 tests across the five
modules, none failing. `GMLReaderTest` is 14 tests and goes from 2 failures to 0
across the change.

modules/core/src/main/java/org/locationtech/jts/io/gml2/GMLReader.java
```diff
@@ -105,6 +111,15 @@ public Geometry read(Reader reader, GeometryFactory geometryFactory) throws SAXE
...
 		fact.setValidating(false);
+		// Harden against XXE, as GML is often read from untrusted sources.
+		try {
+			fact.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
+			fact.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
+		}
+		catch (SAXNotRecognizedException | SAXNotSupportedException | ParserConfigurationException e) {
+			Logger.getLogger(GMLReader.class.getName())
+					.log(Level.WARNING, "SAX parser does not support XXE hardening", e);
+		}
```
grootstebozewolf, line 122: This is fail-open, and one try wraps both features.

If `FEATURE_SECURE_PROCESSING` throws, `disallow-doctype-decl` is never attempted. If errors are going to be ignored, each `setFeature` needs its own try/catch; otherwise a later supported hardening is skipped.

Worse, swallowing turns a config failure into a working XXE. That is not hypothetical — the crimson 1.1.3 table in https://github.com/locationtech/jts/pull/1221#issuecomment-5429639719 shows the try/catch path parses the payload and resolves the entity, while the unguarded path throws while configuring.

`KMLReader` (#1204) sets `SUPPORT_DTD` / `IS_SUPPORTING_EXTERNAL_ENTITIES` with no swallow. Match that: set the two features unguarded (`read` already declares the exception types). Skip only `org.apache.harmony.xml.parsers.SAXParserFactoryImpl` by class name (three lines, no essay). Android rejects names outside the SAX namespace and does not resolve external references, so it is safe unconfigured. Crimson is not. Everywhere else, fail the read rather than parse untrusted GML unhardened.

`java.util.logging` is also new to `jts-core` — there is no logger under `modules/*/src/main` today. Do not mint one for a warning that does not close the hole. If the catch stays, drop the log; if the catch goes, the imports go with it. Do not add Commons XML.

modules/core/src/test/java/org/locationtech/jts/io/gml2/GMLReaderTest.java
```diff
@@ -126,4 +128,37 @@ private void checkRead(String gml, String wktExpected, int srid) {
...
+      }
+      catch (Exception e) {
+        observed = String.valueOf(e.getMessage());
+      }
+      assertFalse(observed.contains(secret));
+    }
+    finally {
+      secretFile.delete();
+    }
+  }
```
grootstebozewolf, line 152: Canary is structured correctly: the secret assert runs on both the success and exception paths, so this cannot pass by never asserting (the first-revision bug). Enough to lock the JDK/Xerces case.

It does not lock the fail-open path. On a parser that rejects both features, `read` now returns a geometry and this test still passes as long as the canary string is absent for some other reason. That is why the factory change needs to fail closed rather than relying on this test alone.

modules/core/src/test/java/org/locationtech/jts/io/gml2/GMLReaderTest.java
```diff
@@ -126,4 +128,37 @@ private void checkRead(String gml, String wktExpected, int srid) {
...
+  public void testDoctypeIsRejected() throws Exception {
+    String gml = "<?xml version=\"1.0\"?>\n<!DOCTYPE foo>\n"
+        + "<gml:Point><gml:coordinates>5,10</gml:coordinates></gml:Point>";
+    try {
+      new GMLReader().read(gml, null);
+      fail("expected a DOCTYPE to be rejected");
+    }
+    catch (Exception e) {
+    }
+  }
```
grootstebozewolf, line 163: Fine for this JUnit 3 file. The empty `catch` only proves *some* exception, which is enough next to the canary test. No need to grow a separate `GMLReaderXXETest` again.

modules/core/src/main/java/org/locationtech/jts/io/gml2/GMLReader.java
grootstebozewolf, line 1: The two-feature set is the right pair on JDK Xerces: without a DOCTYPE there is no subset and no entity to resolve. The three extra Apache/SAX entity features were correctly dropped.

`read(String, …)` already delegates to `read(Reader, …)`, so one factory site covers both entry points. Direct `GMLHandler` use is caller-owned and out of scope; mention that in the class javadoc if you want, not as a blocker.

Jody’s mechanical notes (short comment, tests in `GMLReaderTest`, try/catch) are done. The remaining product question is fail-closed vs warn-and-continue — please fail closed, as above.

Also: PR body still talks about `GMLReaderXXETest` and the three extra features; update it. Commits 2 and 3 have no `Signed-off-by` (the first commit does); Eclipse DCO wants every commit.

grootstebozewolf (changes requested):
Request changes. Jody’s mechanical notes are done (short comment, tests in `GMLReaderTest`, try/catch). The remaining problem is that this version **fail-opens**: on a parser that does not accept the two features, untrusted GML is still read and XXE still works. That is the author’s crimson 1.1.3 table in https://github.com/locationtech/jts/pull/1221#issuecomment-5429639719.

`KMLReader` (#1204) sets `SUPPORT_DTD` / `IS_SUPPORTING_EXTERNAL_ENTITIES` with no swallow. Match that.

**Blockers**
- One try wraps both features. If `FEATURE_SECURE_PROCESSING` throws, `disallow-doctype-decl` is never attempted. Split them, or do not swallow.
- Swallowing turns a config failure into a working XXE. Set the two features unguarded (`read` already declares the exception types). Skip only `org.apache.harmony.xml.parsers.SAXParserFactoryImpl` by class name. Android is safe unconfigured; crimson is not. Everywhere else, fail the read.
- `java.util.logging` is new to `jts-core`. Do not mint the module’s first JUL sink for a warning that does not close the hole. Do not add Commons XML.

**Nits:** empty `catch` on the DOCTYPE test is fine for this JUnit 3 file; PR body still mentions `GMLReaderXXETest` and the extra features; commits 2 and 3 need `Signed-off-by`.

ppkarwasz:
We have just started the release process for [Apache Commons Secure XML](https://commons.apache.org/proper/commons-secure-xml/) (see the [release vote](https://lists.apache.org/thread/fxv1wmgx06nygjdngtc7c6t29wdrk39r)), so I drafted #1230 as an alternative to this PR. The main difference is the failure mode: where this PR logs a warning and continues with an unhardened parser if a feature is not recognized, Commons Secure XML fails **closed** by contract.

The trade-off is straightforward: about 10 lines of hand-written hardening, the way it has been done for the past 25 years, versus an external dependency of roughly 70 KiB (about 10 KiB if shaded). Both are valid choices, and #1230 exists so the maintainers can compare them side by side.

Not sure if it is relevant for JTS users, but Commons Secure XML also hardens Android as a best effort. Android's `expat`-based parser is rather secure by default (although the details depend on the `expat` version used), so it is whitelisted, while on the JVM **fail-close** is part of the contract.

Nexory:
Thanks, both of your code points hold when I check them.

The single try does wrap both calls, so a throw on `FEATURE_SECURE_PROCESSING`
skips `disallow-doctype-decl` entirely. And on the logger: there is no
`java.util.logging` and no other logging framework anywhere under `src/main` in
any module today, so this would be the first, for a warning that does not close
the hole.

For the record on how that shape got there: the unguarded version was what I had
after the crimson measurement, and the try/catch came in at @jodygarnett's
request. The measurement in that table argues the way you do, so the question is
his to settle rather than mine to relitigate. Whichever of the three shapes the
maintainers pick, guarded per feature, fail closed with the Android factory
skipped by class name, or #1230, I will implement it.

Two things you raised are fixed either way:

- All three commits now carry `Signed-off-by`. The rebase that added it also
  had to restore the SSH signatures, which the first attempt silently dropped.
- The description no longer mentions `GMLReaderXXETest` or the three removed
  features.

rubenada:
Any update on this one?

PR #1225, opened 2026-08-13
Fix WKTWriter invalid dimensional EMPTY tokens for nested empty geometries

### LINKED ISSUE #1223: WKTWriter emits invalid dimensional EMPTY tokens for nested empty geometries
### Description

`WKTWriter` can emit an invalid dimensional token when a non-empty geometry collection contains an empty child geometry. The dimension marker inherited from the collection is concatenated directly with `EMPTY`, producing `ZEMPTY`, `MEMPTY`, or `ZMEMPTY`.

The generated text cannot be read back by JTS's own `WKTReader`.

### Reproducer

Using JTS 1.20.0:

```java
import org.locationtech.jts.geom.Geometry;
import org.locationtech.jts.io.ParseException;
import org.locationtech.jts.io.WKTReader;
import org.locationtech.jts.io.WKTWriter;

WKTReader reader = new WKTReader();
Geometry geometry =
    reader.read(
        "GEOMETRYCOLLECTION Z ("
            + "LINESTRING Z (0 0 1, 1 1 2), "
            + "MULTILINESTRING Z EMPTY)");

String written = new WKTWriter(4).write(geometry);
System.out.println(written);
reader.read(written);
```

Actual output:

```text
GEOMETRYCOLLECTION Z(LINESTRING Z(0 0 1, 1 1 2), MULTILINESTRING ZEMPTY)
```

The final line then throws:

```text
org.locationtech.jts.io.ParseException: Expected EMPTY or ( but found 'ZEMPTY' (line 1)
```

The corresponding measured and ZM inputs fail in the same way:

```text
MULTILINESTRING MEMPTY
MULTILINESTRING ZMEMPTY
```

### Expected behavior

`WKTWriter` should keep the dimension marker and `EMPTY` as separate tokens, for example:

```text
MULTILINESTRING Z EMPTY
MULTILINESTRING M EMPTY
MULTILINESTRING ZM EMPTY
```

Every WKT produced by `WKTWriter` should be accepted by `WKTReader`, subject to the configured output dimensions.

### Notes

The issue appears to be that `appendOrdinateText` writes the dimension marker without a trailing separator, after which the empty geometry text writes `EMPTY` immediately. The same code remains present on current `master` at `03517d9b4249f7712569c4016c0d0c7d055f4b07`.

Regression coverage should include nested empty Z, M, and ZM geometries and writer-to-reader round trips.

### Environment

- JTS 1.20.0
- OpenJDK 17.0.13

arimu1:
## Summary

- Fix `WKTWriter` so dimensional empty geometries emit separate `Z`/`M`/`ZM` and `EMPTY` tokens (e.g. `MULTILINESTRING Z EMPTY`) instead of invalid concatenated tokens like `ZEMPTY`, `MEMPTY`, or `ZMEMPTY`.
- Add regression tests covering nested empty Z, M, and ZM geometries with writer-to-reader round trips.

Fixes #1223

## Test plan

- [x] `mvn -pl modules/core -Dtest=WKTWriterTest test` — 15 tests, 0 failures
- [x] `mvn -pl modules/core -Dtest=WKTReadWriteTest,WKTWriterTest,WKTWriterStaticFnTest test` — 32 tests, 0 failures

## Labels

- bug
- jts-core

## Contributor notes

- Eclipse ECA signed for `19286898+arimu1@users.noreply.github.com`
- Commit signed off with `-s`

Made with [Cursor](https://cursor.com)


mprins (approved):

modules/core/src/main/java/org/locationtech/jts/io/WKTWriter.java
```diff
@@ -731,6 +731,17 @@ private void appendOrdinateText(EnumSet<Ordinate> outputOrdinates, Writer writer
...
+  /**
+   * Appends the {@code EMPTY} token, inserting a separator after any dimensional
+   * marker written by {@link #appendOrdinateText}.
+   */
+  private void appendEmptyText(EnumSet<Ordinate> outputOrdinates, Writer writer) throws IOException {
+    if (outputOrdinates.contains(Ordinate.Z) || outputOrdinates.contains(Ordinate.M)) {
+      writer.write(" ");
+    }
+    writer.write(WKTConstants.EMPTY);
+  }
```
grootstebozewolf, line 743: Right cut. `appendOrdinateText` writes `Z`/`M`/`ZM` with no trailing space, and existing goldens pin non-empty WKT as `POINT Z(1 1 1)`, not `Z (`. Do not “fix” this by adding a trailing space on the ordinate marker — that would change every dimensional token.

Spacing only before `EMPTY` when Z or M is present also keeps XY as `POINT EMPTY`: the tagged writer already wrote the space after the type, so this helper must not add a second one.

modules/core/src/main/java/org/locationtech/jts/io/WKTWriter.java
```diff
@@ -920,7 +931,7 @@ private void appendGeometryCollectionText(
     throws IOException
   {
     if (geometryCollection.getNumGeometries() == 0) {
-      writer.write(WKTConstants.EMPTY);
+      appendEmptyText(outputOrdinates, writer);
```
grootstebozewolf, line 934: All six instance `writer.write(EMPTY)` sites now go through the helper (sequence / polygon / multi-point / multi-linestring / multi-polygon / geometry collection). Static `toLineString` already used `" " + EMPTY` and is correctly left alone.

modules/core/src/test/java/org/locationtech/jts/io/WKTWriterTest.java
```diff
@@ -199,4 +199,42 @@ public void testWktLineStringZM() throws ParseException {
...
+    WKTWriter writer4 = new WKTWriter(4);
+
+    Geometry geometryZ = reader.read(
+        "GEOMETRYCOLLECTION Z ("
+            + "LINESTRING Z (0 0 1, 1 1 2), "
+            + "MULTILINESTRING Z EMPTY)");
+    String writtenZ = writer4.write(geometryZ);
+    assertTrue(writtenZ.contains("MULTILINESTRING Z EMPTY"));
+    assertFalse(writtenZ.contains("ZEMPTY"));
+    reader.read(writtenZ);
```
grootstebozewolf, line 218: Exact #1223 repro for Z (and the M / ZM siblings below), plus a `WKTReader` round-trip. Enough to lock the reported bug.

`contains` + `assertFalse(...ZEMPTY)` is fine. An exact-string assert on the collection would also lock that the non-empty sibling still writes `LINESTRING Z(...)` with no space before `(`.

Optional, non-blocking: one top-level `POINT Z EMPTY` (and maybe `POLYGON Z EMPTY`) would pin `appendSequenceText` / `appendPolygonText` without changing the design. Nested empty LINESTRING / POLYGON take the same helper.

modules/core/src/test/java/org/locationtech/jts/io/WKTWriterTest.java
grootstebozewolf, line 1: Dimensional EMPTY belongs here, not in `WKTReadWriteTest` (that file already has XY empty round-trips). Leaving Z/M/ZM EMPTY in `WKTWriterTest` is the right place.

grootstebozewolf (commented):
LGTM — approve with nits.

The EMPTY-only space is the right shape for #1223. `appendOrdinateText` has no trailing space, so a nested empty used to emit `MULTILINESTRING ZEMPTY` and `WKTReader` failed with `Expected EMPTY or ( but found 'ZEMPTY'`. Do not “fix” this by adding a trailing space on the ordinate marker; existing goldens pin non-empty WKT as `POINT Z(1 1 1)`.

All six instance EMPTY sites go through `appendEmptyText`. XY empty stays `POINT EMPTY`. The new test is the exact #1223 repro for Z, M, and ZM.

Nits (non-blocking):
- Coverage is only nested `MULTILINESTRING`. One top-level `POINT Z EMPTY` would pin the other call sites.
- `contains` instead of exact WKT is fine next to the `ZEMPTY` reject.

PR #1229, opened 2026-08-31
Fix NaN handling in LinearLocation/EdgeIntersection compareTo

### LINKED ISSUE #1228: LinearLocation/EdgeIntersection equals() (added in #1201) is inconsistent with hashCode() and non-transitive when segmentFraction/dist is NaN
### Summary

PR #1201 (fixing #1184) added `equals()`/`hashCode()` to `LinearLocation` and `EdgeIntersection`, defined as `compareTo(o) == 0`. However, both classes' `compareTo()` implementations compare the floating-point field with `<`/`>` only:

```java
// LinearLocation.compareTo
if (segmentFraction < other.segmentFraction) return -1;
if (segmentFraction > other.segmentFraction) return 1;
return 0;
```
```java
// EdgeIntersection.compare
if (this.dist < dist) return -1;
if (this.dist > dist) return 1;
return 0;
```

Because IEEE-754 `<`/`>` are always `false` when either operand is `NaN`, `compareTo` falls through to `return 0` ("equal") whenever a `NaN` value is compared against *any* other value on the same segment — not just against another `NaN`. This makes the new `equals()`:

- **Non-transitive**: `a.equals(nan)` and `nan.equals(b)` can both be `true` while `a.equals(b)` is `false`.
- **Inconsistent with `hashCode()`**: `nan.equals(other)` can be `true` while `nan.hashCode() != other.hashCode()`, breaking hash-based collections (`HashSet`/`HashMap`).

### Reproduction

`LinearLocation` `segmentFraction` is `NaN` for zero-length line segments — a real, reachable state (see `LinearLocationTest` for zero-length lines):

```java
LinearLocation zero = new LinearLocation(0, 0, 0.0);
LinearLocation nan  = new LinearLocation(0, 0, Double.NaN);
LinearLocation half = new LinearLocation(0, 0, 0.5);

zero.equals(nan);   // true  (compareTo falls through to 0)
nan.equals(half);   // true  (compareTo falls through to 0)
zero.equals(half);  // false -- violates transitivity

nan.hashCode() == half.hashCode(); // false, even though nan.equals(half) above can be true
```

The same pattern applies to `EdgeIntersection` (`dist` field), which has no `NaN` handling at all in `compare()`.

### Suggested fix

Give `NaN` an explicit, well-defined position in the ordering, consistent with `Double.compare()`/`Double.equals()` semantics (which already treat `NaN` as equal to itself and ordered above all non-`NaN` values), e.g.:

```java
int cmp = Double.compare(segmentFraction, other.segmentFraction);
if (cmp != 0) return cmp;
```

This keeps `compareTo`, `equals`, and `hashCode` mutually consistent for all inputs, including `NaN`, without changing behavior for the (presumably overwhelmingly common) non-`NaN` case.

### Context

Found while porting #1201 to NetTopologySuite (NetTopologySuite/NetTopologySuite#872), where regression tests reproducing this were added to document the behavior pending a fix.

bjornharrtell:
Fixes #1228

`LinearLocation.compareTo()` and `EdgeIntersection.compare()` compared the floating-point fields (`segmentFraction`/`dist`) with raw `<`/`>`, which are always `false` for `NaN`. This made `compareTo` fall through to `return 0` ("equal") whenever a `NaN` was compared against *any* other value on the same segment/index, not just against another `NaN`.

Since `equals()` for both classes (added in #1201) is defined as `compareTo(o) == 0`, this made `equals()` non-transitive and inconsistent with `hashCode()` whenever `NaN` was involved (e.g. `segmentFraction` is `NaN` for zero-length line segments).

### Fix
Replace the manual `<`/`>` comparisons with `Double.compare()`, which gives `NaN` a well-defined, transitive ordering consistent with `Double.hashCode()`/`Objects.hash()` (already used by both classes' `hashCode()`). No behavior change for non-`NaN` values.

### Tests
Added regression tests reproducing the transitivity/hashCode issue from the report:
- `LinearLocationTest.testEqualsHashCodeConsistentWithNaN`
- `EdgeIntersectionTest.testEqualsHashCodeConsistentWithNaN`


modules/core/src/main/java/org/locationtech/jts/linearref/LinearLocation.java
grootstebozewolf, line 1: Same `<` / `>` fall-through still lives in `compareLocationValues` (instance and static). `LocationIndexOfPoint.indexOfFromStart` uses that for “strictly after `minIndex`”, so a `NaN` fraction still compares equal to every fraction on that segment. Same one-liner: `return Double.compare(...)`.

modules/core/src/main/java/org/locationtech/jts/linearref/LinearLocation.java
```diff
@@ -330,10 +330,8 @@ public int compareTo(Object o) {
     if (segmentIndex < other.segmentIndex) return -1;
     if (segmentIndex > other.segmentIndex) return 1;
     // same segment, so compare segment fraction
-    if (segmentFraction < other.segmentFraction) return -1;
-    if (segmentFraction > other.segmentFraction) return 1;
-    // same location
-    return 0;
+    // Double.compare is used to ensure consistency with equals/hashCode for NaN
+    return Double.compare(segmentFraction, other.segmentFraction);
```
grootstebozewolf, line 334: Finite-vs-finite order is unchanged except signed zero: `Double.compare(-0.0, +0.0) == -1`, while the old path returned `0`. `-0.0` survives `normalize()` (`-0.0 < 0.0` is false), and `Objects.hash` already hashed them differently, so this also closes an equals/hashCode hole. Worth a sentence in the PR; not a reason to revert.

modules/core/src/main/java/org/locationtech/jts/geomgraph/EdgeIntersection.java
```diff
@@ -97,9 +97,8 @@ public int compare(int segmentIndex, double dist)
   {
     if (this.segmentIndex < segmentIndex) return -1;
     if (this.segmentIndex > segmentIndex) return 1;
-    if (this.dist < dist) return -1;
-    if (this.dist > dist) return 1;
-    return 0;
+    // Double.compare is used to ensure consistency with equals/hashCode for NaN
+    return Double.compare(this.dist, dist);
```
grootstebozewolf, line 101: Same signed-zero note as on `LinearLocation`: `Double.compare` distinguishes `-0.0` and `+0.0`. `EdgeIntersectionList` is a `TreeMap` keyed on this order, so the `NaN` collapse was not only a `HashSet` issue.

modules/core/src/test/java/org/locationtech/jts/linearref/LinearLocationTest.java
```diff
@@ -61,6 +61,22 @@ public void testEqualsHashCodeContract() throws Exception
...
+    LinearLocation half = new LinearLocation(0, 0, 0.5);
+
+    // NaN must only be equal to itself, not to arbitrary other values (transitivity)
+    assertFalse(zero.equals(nan));
+    assertFalse(nan.equals(half));
+    assertFalse(zero.equals(half));
+    assertEquals(nan, new LinearLocation(0, 0, Double.NaN));
+
+    // equals/hashCode contract must hold even when NaN is involved
+    assertEquals(nan.hashCode(), new LinearLocation(0, 0, Double.NaN).hashCode());
```
grootstebozewolf, line 77: `assertFalse(zero.equals(half))` is not a `NaN` case. Stronger: put `zero` and `nan` in the same `HashSet` / `TreeSet` (that was the collection failure), and `assertTrue(nan.compareTo(zero) > 0)` for the `Double.compare` total order.

Also, `LineSegment.segmentFraction` already maps `NaN` to `1.0`, then `normalize()` walks that to the next vertex — that is why `testZeroLengthLineString` still compares against `(1, 0.0)`. The public constructors still accept `NaN`, so the contract fix stands; `indexOf` on a collapsed segment does not store `NaN`.

modules/core/src/test/java/org/locationtech/jts/geomgraph/EdgeIntersectionTest.java
```diff
@@ -47,4 +47,19 @@ public void testEqualsHashCodeContract() {
...
+
+    // NaN must only be equal to itself, not to arbitrary other values (transitivity)
+    assertFalse(zero.equals(nan));
+    assertFalse(nan.equals(half));
+    assertFalse(zero.equals(half));
+    assertEquals(nan, new EdgeIntersection(new Coordinate(1, 2), 0, Double.NaN));
+
+    // equals/hashCode contract must hold even when NaN is involved
+    assertEquals(nan.hashCode(), new EdgeIntersection(new Coordinate(1, 2), 0, Double.NaN).hashCode());
+  }
```
grootstebozewolf, line 64: Same as the `LinearLocation` twin: `assertFalse(zero.equals(half))` is not a `NaN` case. Worth putting `zero` and `nan` in one `HashSet` / `TreeSet`, and asserting `nan.compareTo(zero) > 0`.

grootstebozewolf (commented):
LGTM. `Double.compare` is the right fix for the #1201 `equals` / `compareTo` / `hashCode` contract on `NaN`. Nits below are optional follow-ups, not blockers.

